PR #1, opened 2022-06-20
[WIP] DKG spec

gl-rockx:


alonmuroch:
@gl-rockx should we close this PR?

PR #3, opened 2022-07-11
[WIP] New version of DKG

gl-rockx:
## Summary
- Go implementation of DKG based on herumi/bls
- Use of prepared random numbers as input to protocol instead of generating in flight

## TODO
- [ ] Adapt message types
- [ ] Add spectest
- [ ] Adapt runner
- [ ] Encrypt distributed share before sending
- [ ] Add blame


gl-rockx:
To be replaced

PR #5, opened 2022-07-19
Use RoundRobinProposer in qbft TestingConfig

nkryuchkov:


nkryuchkov:
Temporary on hold, do not merge: It breaks some tests because they expect a different proposal leader
cc @nivBlox

PR #6, opened 2022-07-19
Add remove share to key-manager interface

olegshmuelov:


alonmuroch (changes requested):
What is this used for?

olegshmuelov:
> What is this used for?

There are events like:
- ValidatorRemovalEvent
- OperatorRemovalEvent

that requires to delete the share secrets from the EKM

alonmuroch (approved):

PR #11, opened 2022-08-02
Shifu Testnet Domain

amir-blox:
- added DomainType for `shifu`
- added global domain to be used by other components

### DIFF
--- a/types/crypto_test.go
+++ b/types/crypto_test.go
@@ -6,6 +6,10 @@ import (
 	"testing"
 )
 
+func init() {
+	bls.Init(0)
+}
+
 type testSigningRoot struct {
 	root      []byte
 	Signature []byte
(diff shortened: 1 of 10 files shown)


types/crypto_test.go
```diff
@@ -6,6 +6,10 @@ import (
 	"testing"
 )
 
+func init() {
+	bls.Init(0)
```
nkryuchkov, line 10: Does `0` mean that we need to use `MCL_BN254` curve type? Why not `bls.BLS12_381`?

nkryuchkov (approved):

types/crypto_test.go
amir-blox: Was removed, should be handled outside of this PR scope

PR #12, opened 2022-08-02
Makefile and Workflow

amir-blox:
Adding a Makefile and CI (failed for now until we fix all the tests and lints)


.github/workflows/test.yaml
```diff
@@ -0,0 +1,34 @@
...
+        with:
+          go-version: 1.17
+
+      - name: Get Dependencies
+        run: go get -v -t -d ./...
+
+      - name: Lint
+        run: make lint-prepare && make lint
+
+      - name: Test Types
```
nkryuchkov, line 27: Do we need to test the `dkg` package as well?

nkryuchkov (approved):

alonmuroch:
@amir-blox I've made generating json tests simple by running 'go generate ./...'
Then you can simply run all tests and it should pass

amir-blox:
> @amir-blox I've made generating json tests simple by running 'go generate ./...' Then you can simply run all tests and it should pass

@alonmuroch `./types` and `/ssv` still fails

PR #26, opened 2022-08-08
[WIP] New version of DKG - No protobuf

gl-rockx:


dkg/node.go
```diff
@@ -69,6 +73,10 @@ func (n *Node) ProcessMessage(msg *types.SSVMessage) error {
 		return n.startNewDKGMsg(signedMsg)
 	case ProtocolMsgType:
 		return n.processDKGMsg(signedMsg)
+	case DepositDataMsgType:
```
alonmuroch, line 76: that should be in the runner struct. 
A node holds multiple runners and a runner should take care of the entire DKG cycle

dkg/protocol.go
```diff
@@ -10,6 +11,54 @@ type KeyGenOutput struct {
...
+}
+
+type keygenOutputJSON struct {
+	Share           []byte
+	OperatorPubKeys map[types.OperatorID][]byte
+	ValidatorPK     types.ValidatorPK
+	Threshold       uint64
+}
+
+func (k *KeyGenOutput) MarshalJSON() ([]byte, error) {
```
alonmuroch, line 24: We use the Encoder interface not marshalJSON
https://github.com/bloxapp/ssv-spec/blob/main/types/encoder.go

dkg/protocol.go
```diff
@@ -10,6 +11,54 @@ type KeyGenOutput struct {
...
+	rep := keygenOutputJSON{
```
alonmuroch, line 25: why can't a simple json.marshal work here?

dkg/runner.go
```diff
@@ -51,22 +54,28 @@ func (r *Runner) ProcessMsg(msg *SignedMessage) (bool, map[types.OperatorID]*Sig
 			if err != nil {
 				return false, nil, errors.Wrap(err, "could not generate deposit data")
 			}
+
 			r.DepositDataRoot = root
 
 			// sign
+			sig := r.KeyGenOutput.Share.SignByte(root)
```
alonmuroch, line 61: why was this replaced?

dkg/runner.go
```diff
@@ -132,15 +149,70 @@ func (r *Runner) ProcessMsg(msg *SignedMessage) (bool, map[types.OperatorID]*Sig
...
+		Message: &Message{
+			MsgType:    msgType,
+			Identifier: r.Identifier,
+			Data:       data,
+		},
+		Signer:    r.Operator.OperatorID,
+		Signature: nil,
+	}
+	// GLNOTE: Should we use SignDKGOutput?
+	sig, err := r.config.Signer.SignDKGOutput(signedMessage, r.Operator.ETHAddress)
```
alonmuroch, line 166: that not what this function is about, it takes an encoder interface and broadcast it to the network.. it's not a msg specific func

dkg/types.go
```diff
@@ -36,4 +36,5 @@ type Config struct {
 	Storage             Storage
 	SignatureDomainType types.DomainType
 	Signer              types.DKGSigner
+	Verifier            types.DKGVerifier
```
alonmuroch, line 39: why do we need it here? it should be in the validateSignedOutput func

types/testingutils/keygen_protocol.go
```diff
@@ -0,0 +1,15 @@
+package testingutils
```
alonmuroch, line 1: that should be in the dkg package.. testing utils is intended to configure existing structs. We better not use mocks

types/testingutils/fakes.go
```diff
@@ -0,0 +1,17 @@
+package testingutils
```
alonmuroch, line 1: let's create a dkg file in testingutils and put all fixtures there in one place

types/testingutils/keys.go
```diff
@@ -387,3 +388,17 @@ func (ks *TestKeySet) Committee() []*types.Operator {
 	}
 	return committee
 }
+
+func (ks *TestKeySet) KeyGenOutput(opId types.OperatorID) *dkg.KeyGenOutput {
```
alonmuroch, line 392: shouldn't be here, if it's a fixture lets put it in a dedicated dkg.go file

types/testingutils/network.go
```diff
@@ -6,12 +6,16 @@ import (
...
+	BroadcastedMsgs    []*types.SSVMessage
+	BroadcastedDKGMsgs []*dkg.SignedMessage // GLNOTE: Should we use two fields or should make a new TestingNetwork instance?
+	Outputs            map[types.OperatorID]*dkg.SignedOutput
 }
 
 func NewTestingNetwork() *TestingNetwork {
 	return &TestingNetwork{
-		BroadcastedMsgs: make([]*types.SSVMessage, 0),
+		BroadcastedMsgs:    make([]*types.SSVMessage, 0),
+		BroadcastedDKGMsgs: make([]*dkg.SignedMessage, 0),
```
alonmuroch, line 17: all network msgs are SSVMessage, all dkg messages need to be wrapped in it.

types/verifier.go
```diff
@@ -0,0 +1,7 @@
+package types
+
+import "github.com/ethereum/go-ethereum/common"
```
alonmuroch, line 3: that should be in the dkg package

types/testingutils/keymanager.go
```diff
@@ -191,12 +192,25 @@ func (km *testingKeyManager) Decrypt(pk *rsa.PublicKey, cipher []byte) ([]byte,
...
-	panic("implemet")
+	root, err := output.GetRoot()
+	if err != nil {
+		return nil, err
+	}
+	sig := FakeEcdsaSign(root, address[:]) // GLNOTE: Use fake so that we have deterministic message
+	return sig, nil
+}
+
+func (km *testingKeyManager) VerifyDKGOutput(signature types.Signature, output types.Root, address common.Address) bool {
```
alonmuroch, line 208: verify should not be part of the key manager struct

alonmuroch (changes requested):

dkg/protocol.go
gl-rockx: there's issue with using json.marshal for bls.SecretKey, bls.PublicKey

types/testingutils/fakes.go
gl-rockx: We can do that. I keep it here first so that we can minimize the changes to current code.

types/testingutils/keymanager.go
gl-rockx: You're right. I put it here so that we can see the (mock) sign and verify side by side.

dkg/runner.go
gl-rockx: The deposit data needs to be signed with the bls key, which in here means we need to use the share to sign. I don't understand what's the purpose of the original function `SignETHDepositRoot(root, r.Operator.ETHAddress)` but share is not related to `ETHAddress`.

Note: At the moment `SignETHDepositRoot` is not used at all.

dkg/node.go
gl-rockx: If we don't include this MsgType here. Runner won't be able to get the message. Node and runner are checking the same field.

gl-rockx: Or should the node check SSVMessage's MsgType instead of SignedMessage's MsgType?

dkg/runner.go
gl-rockx: I'm confused. The function name says we need to sign the message. But if we don't include the `SignedMessage` and `SignDKGOutput`, how do we do the signing?

dkg/types.go
gl-rockx: It is used in `validateSignedOutput`. It's included in the config so that we can use a different verification strategy (i.e. mock) for testing. Otherwise, we won't be able to use the ecdsa in spectest as ecdsa involves random number and hence non-deterministic.

types/testingutils/keygen_protocol.go
gl-rockx: If we want to include the ecdsa signature in spectest, we need mocks so that the messages are deterministic. Otherwise, we have to exclude signature verification in spectest.

dkg/types.go
alonmuroch: You verify that the signature is valid, that should be ok in spec tests as long as the operators have the same eth address.
I'm for removing it, all those types of validations should be constant

types/testingutils/fakes.go
alonmuroch: no need to open a new file, the dkg file should have all of it in it otherwise it's gonna be hard to manage

types/testingutils/keygen_protocol.go
alonmuroch: we can do signature verification instead of signature comparison.

alonmuroch (changes requested):

dkg/protocol.go
gl-rockx: Removed it.

dkg/node.go
gl-rockx: Here we only check InitMsgType.

types/testingutils/keymanager.go
```diff
@@ -193,12 +208,24 @@ func (km *testingKeyManager) Decrypt(pk *rsa.PublicKey, cipher []byte) ([]byte,
 
 // Encrypt given a rsa pubkey and data returns an PKCS1v15 e
 func (km *testingKeyManager) Encrypt(pk *rsa.PublicKey, data []byte) ([]byte, error) {
-	panic("implement")
+	return FakeEncryption(data), nil
```
alonmuroch, line 211: shouldn't be fake, the keys can be hardcoded but better use the real encryption for test correctness

types/testingutils/network.go
```diff
@@ -7,11 +7,13 @@ import (
 
 type TestingNetwork struct {
 	BroadcastedMsgs []*types.SSVMessage
+	DKGOutputs      map[types.OperatorID]*dkg.SignedOutput
 }
 
 func NewTestingNetwork() *TestingNetwork {
 	return &TestingNetwork{
 		BroadcastedMsgs: make([]*types.SSVMessage, 0),
+		DKGOutputs:      make(map[types.OperatorID]*dkg.SignedOutput, 0),
```
alonmuroch, line 16: SSVMessages include in them DKG messages, the test should decode it from the SSVMessage like other tests are doing

alonmuroch (changes requested):

dkg/node.go
```diff
@@ -56,6 +60,9 @@ func (n *Node) newRunner(id RequestID, initMsg *Init) (*Runner, error) {
 
 // ProcessMessage processes network Messages of all types
 func (n *Node) ProcessMessage(msg *types.SSVMessage) error {
+	if msg.MsgType != types.DKGMsgType {
```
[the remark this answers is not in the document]
gl-rockx: There's no validation of SSVMessage, and in here it's just one of the SSVMessage types.

dkg/protocol.go
```diff
@@ -10,6 +10,7 @@ type KeyGenOutput struct {
 	Share           *bls.SecretKey
 	OperatorPubKeys map[types.OperatorID]*bls.PublicKey
 	ValidatorPK     types.ValidatorPK
+	Threshold       uint64
```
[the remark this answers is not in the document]
gl-rockx: It's included here so that this struct contains sufficient information for the share to operate. It'll be more tedious to get the info if it's scattered across two types. However, if we enforce that threshold = 2f+1, we can derive it from the number of operators which is 3f+1. But including it here gives us more flexibility to allow other thresholds (e.g. 10 out of 13).

dkg/runner.go
```diff
@@ -132,15 +151,79 @@ func (r *Runner) ProcessMsg(msg *SignedMessage) (bool, map[types.OperatorID]*Sig
...
+		return errors.Wrap(err, "failed to broadcast message")
+	}
+	return nil
 }
 
 func (r *Runner) reconstructDepositDataSignature() (types.Signature, error) {
-	panic("implement")
+	sigBytes := map[types.OperatorID][]byte{}
+	for id, d := range r.DepositDataSignatures {
+		if err := r.validateDepositDataSig(d); err != nil {
```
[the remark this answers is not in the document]
gl-rockx: Actually what we need here is just to check if the signing roots match. I have made changes to split validation of signature and roots. The reason why we need to split them is when we receive the peer's message, we might not have our own root yet (if peer acts faster), so we have to defer the validation root.

types/testingutils/keymanager.go
gl-rockx: Changed to use types.Encrypt.

types/testingutils/network.go
gl-rockx: This is to align with `StreamDKGOutput` function we have. It's actually different from the network channel.
All the messages containing `SignedOutput` are actually received from the `network` then once all `SignedOutput` messages are received, these messages will be streamed to any local consumers. This is to cater to the second step.

dkg/node.go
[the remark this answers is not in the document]
alonmuroch: ok lets leave as is and we can refactor later

types/testingutils/network.go
alonmuroch: Will need to refactor later

alonmuroch (approved):

PR #42, opened 2022-08-17
Align error messages with ssv

nkryuchkov:
When validating messages, https://github.com/bloxapp/ssv usually uses common checks for all message types and the same error message on error. E. g., `message height is wrong` instead of `commit Height is wrong`. Aligning ssv errors to ssv-spec would require rewriting a lot of code in ssv. This PR can also fix the issue by changing error texts in the spec.


nkryuchkov:
Replaced by https://github.com/bloxapp/ssv/pull/692

PR #48, opened 2022-08-30
Qbft sync

alonmuroch:
A simple mockup of when qbft syncing is triggered


qbft/controller.go
```diff
@@ -40,13 +40,15 @@ type Controller struct {
...
 	StoredInstances InstanceContainer
-	Domain          types.DomainType
-	Share           *types.Share
-	signer          types.SSVSigner
-	valueCheck      ProposedValueCheckF
-	storage         Storage
-	network         Network
-	proposerF       ProposerF
+	// HigherReceivedMessages holds all msgs from a higher height
+	HigherReceivedMessages *MsgContainer
```
nivBlox, line 44: @alonmuroch missing initial for container

qbft/sync.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/pkg/errors"
+)
+
+func (c *Controller) processHigherHeightMsg(msg *SignedMessage) error {
+	added, err := c.HigherReceivedMessages.AddIfDoesntExist(msg)
+	if err != nil {
+		return errors.Wrap(err, "could not add higher height msg")
+	}
+	if added && c.f1SyncTrigger() {
+		// TODO should reset msg container? past msgs? all msgs?
```
nivBlox, line 14: i think we can reset each time sync is done

nivBlox: @alonmuroch also we need to add "clean" func to container

alonmuroch: suggestions? to prevent replay attacks?

nivBlox: clean once got f+1.
protect against replay attack will be handle using scoring as we discussed.

alonmuroch: Not sure this should be "part of the spec" though..

Issue #62, opened 2022-09-10
bump controller height when instance decides not when starts


alonmuroch:
It will enable starting height from 0 and make it uint

alonmuroch:
solved

PR #69, opened 2022-09-12
Message struct encoding SIP

olegshmuelov:
https://github.com/bloxapp/SIPs/blob/main/sips/msg_struct_encoding.md


qbft/instance.go
```diff
@@ -57,50 +57,50 @@ func (i *Instance) Start(value []byte, height Height) {
...
+			}
+
+			if err := i.Broadcast(broadcastMsg); err != nil {
 				// TODO - log
 			}
 		}
 	})
 }
 
-func (i *Instance) Broadcast(msg *SignedMessage) error {
```
alonmuroch, line 67: removing this func means we duplicate the broadcast func everywhere.. any reason we removed it?

qbft/messages.go
```diff
@@ -19,6 +19,17 @@ func HasQuorum(share *types.Share, msgs []*SignedMessage) bool {
 	return share.HasQuorum(len(uniqueSigners))
 }
 
+// HasQuorumHeaders returns true if a unique set of signers has quorum
```
alonmuroch, line 22: we need to find a better way to do this other than duplicating the code

qbft/messages.go
```diff
@@ -148,47 +159,56 @@ func (d *RoundChangeData) Validate() error {
...
+// Message includes the full consensus input to be decided on, used for decided, proposal and round-change messages
 type Message struct {
-	MsgType    MessageType
-	Height     Height // QBFT instance Height
-	Round      Round  // QBFT round for which the msg is for
-	Identifier []byte // instance Identifier this msg belongs to
-	Data       []byte
+	Height Height
+	Round  Round
+	Input  []byte `ssz-max:"2048"`
```
alonmuroch, line 166: is 2048 enough for all our msgs?

qbft/messages.go
```diff
@@ -214,23 +234,156 @@ func (msg *Message) GetRoot() ([]byte, error) {
...
-	if len(msg.Data) == 0 {
-		return errors.New("message data is invalid")
-	}
-	if msg.MsgType > 5 {
-		return errors.New("message type is invalid")
+func (msg *Message) Validate(msgType types.MsgType) error {
+	//if len(msg.Identifier) == 0 {
+	//	return errors.New("message identifier is invalid")
+	//}
+	if len(msg.Input) == 0 && msgType != types.ConsensusRoundChangeMsgType {
```
alonmuroch, line 241: ?

qbft/messages.go
```diff
@@ -214,23 +234,156 @@ func (msg *Message) GetRoot() ([]byte, error) {
...
+
+func (signedMsg *SignedMessageHeader) GetSignature() types.Signature {
+	return signedMsg.Signature
+}
+func (signedMsg *SignedMessageHeader) GetSigners() []types.OperatorID {
+	return signedMsg.Signers
+}
+
+// Encode returns a msg encoded bytes or error
+func (signedMsg *SignedMessageHeader) Encode() ([]byte, error) {
```
alonmuroch, line 323: we should add a followup cleanup for all the root and encoding interfaces we use as ssz comes with them already.

qbft/messages.go
```diff
@@ -319,27 +485,30 @@ func (signedMsg *SignedMessage) GetRoot() ([]byte, error) {
 // DeepCopy returns a new instance of SignedMessage, deep copied
 func (signedMsg *SignedMessage) DeepCopy() *SignedMessage {
```
alonmuroch, line 486: used anywhere? if not we can remove

qbft/messages.go
```diff
@@ -319,27 +485,30 @@ func (signedMsg *SignedMessage) GetRoot() ([]byte, error) {
...
+
+	copy(ret.Message.Input, signedMsg.Message.Input)
 	return ret
 }
 
 // Validate returns error if msg validation doesn't pass.
 // Msg validation checks the msg, it's variables for validity.
-func (signedMsg *SignedMessage) Validate() error {
+func (signedMsg *SignedMessage) Validate(msgType types.MsgType) error {
 	if len(signedMsg.Signature) != 96 {
```
alonmuroch, line 512: no need as ssz insures its 96 bytes

qbft/prepare.go
```diff
@@ -9,22 +9,24 @@ import (
...
+	//if err != nil {
+	//	return errors.Wrap(err, "could not get accepted proposal data")
+	//}
 	if err := validSignedPrepareForHeightRoundAndValue(
 		i.config,
 		signedPrepare,
 		i.State.Height,
 		i.State.Round,
-		acceptedProposalData.Data,
+		i.State.ProposalAcceptedForCurrentRound.Message.Input,
```
alonmuroch, line 28: not sure this works here.. input is the entire struct not just the value

qbft/prepare.go
```diff
@@ -9,22 +9,24 @@ import (
 func (i *Instance) uponPrepare(
```
alonmuroch, line 9: prepare msgs should use input root not the entire msg

qbft/prepare.go
```diff
@@ -56,23 +58,36 @@ func (i *Instance) uponPrepare(
 		return errors.Wrap(err, "could not create commit msg")
 	}
 
-	if err := i.Broadcast(commitMsg); err != nil {
+	commitEncoded, err := commitMsg.Encode()
```
alonmuroch, line 61: should be moved to a reusable func

qbft/commit.go
```diff
@@ -7,9 +7,12 @@ import (
 )
 
 // UponCommit returns true if a quorum of commit messages was received.
-func (i *Instance) UponCommit(signedCommit *SignedMessage, commitMsgContainer *MsgContainer) (bool, []byte, *SignedMessage, error) {
+func (i *Instance) UponCommit(
```
alonmuroch, line 10: commit msgs should use signed msg header with the input root not the full msg

qbft/proposal.go
```diff
@@ -1,9 +1,9 @@
...
 
 import (
-	"bytes"
-	"github.com/bloxapp/ssv-spec/types"
 	"github.com/pkg/errors"
+
+	"github.com/bloxapp/ssv-spec/types"
 )
 
 func (i *Instance) uponProposal(signedProposal *SignedMessage, proposeMsgContainer *MsgContainer) error {
```
alonmuroch, line 9: proposal should include the full input. Proposal justifications can be signed msg headers

qbft/proposal.go
```diff
@@ -29,17 +29,29 @@ func (i *Instance) uponProposal(signedProposal *SignedMessage, proposeMsgContain
 	}
 	i.State.Round = newRound
 
-	proposalData, err := signedProposal.Message.GetProposalData()
```
alonmuroch, line 32: no check for err

qbft/proposal.go
```diff
@@ -101,8 +116,8 @@ func isValidProposal(
 func isProposalJustification(
 	state *State,
 	config IConfig,
-	roundChangeMsgs []*SignedMessage,
-	prepareMsgs []*SignedMessage,
+	roundChangeJustifications []*SignedMessageHeader,
```
alonmuroch, line 119: justifications could be the headers

qbft/proposal.go
```diff
@@ -116,32 +131,49 @@ func isProposalJustification(
...
+		// no quorum, duplicate signers, invalid still has quorum, invalid no quorum
 		// prepared
-		for _, rc := range roundChangeMsgs {
-			if err := validRoundChange(state, config, rc, height, round); err != nil {
-				return errors.Wrap(err, "change round msg not valid")
+		for _, rcj := range roundChangeJustifications {
+			//if err := validRoundChange(state, config, rcj, height, round); err != nil {
+			//	return errors.Wrap(err, "change round msg not valid")
+			//}
+			if err := validSignedPrepareHeaderForHeightRoundAndValue(
```
alonmuroch, line 140: this is wrong, it should check round change

qbft/proposal.go
```diff
@@ -116,32 +131,49 @@ func isProposalJustification(
...
+				//if err != nil {
+				//	return false, errors.Wrap(err, "could not get round change data")
+				//}
+				//if rcData.Prepared() {
+				//	return true, nil
+				//}
+
+				// TODO<olegshmuelov> check for input root
+				//if rcj.Message.PreparedRound != NoRound || len(d.PreparedValue) != 0 {
+				if rcj.Message.PreparedRound != NoRound {
```
alonmuroch, line 171: we had an IsPrepared func for the previous msg struct we should use that

qbft/round_change.go
```diff
@@ -185,67 +223,77 @@ func validRoundChange(state *State, config IConfig, signedMsg *SignedMessage, he
...
-	if err != nil {
-		return errors.Wrap(err, "could not get roundChange data ")
-	}
-	if err := rcData.Validate(); err != nil {
-		return errors.Wrap(err, "roundChangeData invalid")
-	}
+	// TODO<olegshmuelov>: len(signedMsg.Message.Input) is always != 0 (this check already done in message validation),
+	// so is it always prepared?
+	// check if prepared
+	if signedMsg.Message.PreparedRound != NoRound || len(signedMsg.Message.Input) != 0 {
```
alonmuroch, line 229: should have a dedicate isPrepared func

ssv/consensus.go
```diff
@@ -30,8 +30,8 @@ func (dr *Runner) Decide(input *types.ConsensusData) error {
 	return nil
 }
 
-func (dr *Runner) ProcessConsensusMessage(msg *qbft.SignedMessage) (decided bool, decidedValue *types.ConsensusData, err error) {
-	decided, decidedValueByts, err := dr.QBFTController.ProcessMsg(msg)
+func (dr *Runner) ProcessConsensusMessage(msgID types.MessageID, msg *qbft.SignedMessage) (decided bool, decidedValue *types.ConsensusData, err error) {
```
alonmuroch, line 33: this is weird... it's not according to the new runner struct and interfaces...

alonmuroch (changes requested):

alonmuroch:
Discussion comments:

- value check should receive Data and not bytes
- in value check we need to verify the actual data root == Data.root
- move qbft message data source to signed message (along side justifications) and include in message only [32]byte root
- value check implementation should include verifying that the data's root == the source root 
- DKG changes for V2 of msg struct
- add source data root == Data.root in validateDecided func
- when constructing aggregated commit, we need to copy source data to it so it will pass decided validation
- should round change include source data? Since we construct a proposal post round change quorum and the node might not have the source data
-

qbft/controller.go
```diff
@@ -48,14 +50,21 @@ type Controller struct {
...
-		Height:              -1, // as we bump the height when starting the first instance
+		Identifier: identifier,
+		// TODO<olegshmuelov>: fastssz does not support int but only uint.
+		// this why the Height type was changed from int64 to uint64
+		// The initial value of the height was changed from -1 to math.MaxUint64
+		// when we bump the height at the first instance we get 0 (math.MaxUint64 + 1 = 0)
+		// if we want to keep the Height to be int64 there are 2 possible options:
+		// 1. create a messageSSZ with uint64 instead of Height and transform to message with int64
+		// 2. open a pr for fastssz with the implementation that supports int
+		Height:              math.MaxUint64, // as we bump the height when starting the first instance
```
olegshmuelov, line 67: fastssz does not support int but only uint.
this why the Height type was changed from int64 to uint64
the initial value of the height was changed from -1 to math.MaxUint64
when we bump the height at the first instance we get 0 (math.MaxUint64 + 1 = 0)
if we want to keep the Height to be int64 there are 2 possible options:
  1. create a messageSSZ with uint64 instead of Height and transform to message with int64
  2. [bump controller height when instance decides not when starts](https://github.com/bloxapp/ssv-spec/issues/62)

qbft/messages.go
```diff
@@ -214,39 +90,49 @@ func (msg *Message) GetRoot() ([]byte, error) {
...
+	case types.ConsensusProposeMsgType, types.DecidedMsgType:
+		if len(inputSource) == 0 || m.InputRoot == [32]byte{} {
+			return errors.New("message input data is invalid")
+		}
+	case types.ConsensusPrepareMsgType, types.ConsensusCommitMsgType:
+		if m.InputRoot == [32]byte{} {
+			return errors.New("message input data is invalid")
+		}
+	case types.ConsensusRoundChangeMsgType:
+		// TODO<olegshmuelov>: the input data for round change can be nil?
```
olegshmuelov, line 104: the input data [rooot,source] for round change can be nil?

qbft/messages.go
```diff
@@ -359,5 +249,128 @@ func (signedMsg *SignedMessage) Validate() error {
...
+// SizeSSZ returns the ssz encoded size in bytes for the SignedMessage object
+// TODO<olegshmuelov>: the calculation of the signed msg size should be improved performance wise
+func (s *SignedMessage) SizeSSZ() int {
+	smSSZ, err := s.toSignedMessageSSZ()
+	if err != nil {
+		return 0
+		//panic(err)
+	}
+	return smSSZ.SizeSSZ()
+}
```
olegshmuelov, line 294: the calculation of the signed msg size should be improved performance wise.

qbft/round_change.go
```diff
@@ -27,40 +22,44 @@ func (i *Instance) uponRoundChange(
...
-		if highestRCData.Prepared() {
-			valueToPropose = highestRCData.PreparedValue
+		valueToPropose := i.StartValue
+		if justifiedRoundChangeMsg.Message.Prepared() {
+			// TODO<olegshmuelov>: validate that justified round change msg holds the complete input data
+			valueToPropose = &Data{
+				Root:   justifiedRoundChangeMsg.Message.InputRoot,
+				Source: justifiedRoundChangeMsg.InputSource,
+			}
 		}
```
olegshmuelov, line 41: [Validate justified prepared round change msg holds the complete input data](https://github.com/bloxapp/ssv-spec/issues/103)

qbft/messages.go
```diff
@@ -30,181 +30,57 @@ func HasPartialQuorum(share *types.Share, msgs []*SignedMessage) bool {
...
-func (msg *Message) Encode() ([]byte, error) {
-	return json.Marshal(msg)
-}
-
-// Decode returns error if decoding failed
-func (msg *Message) Decode(data []byte) error {
-	return json.Unmarshal(data, &msg)
+// Prepared returns error if decoding failed
+func (m *Message) Prepared() bool {
+	return m.PreparedRound != NoRound
```
olegshmuelov, line 78: should we keep the same logic to check if the round check is prepared
```go
if d.PreparedRound != NoRound || len(d.PreparedValue) != 0 {
	return true
}
```

qbft/spectest/all_tests.go
```diff
@@ -101,14 +101,14 @@ var AllTests = []SpecTest{
 	proposer.ThirteenOperators(),
 
 	messages.RoundChangeDataInvalidJustifications(),
-	messages.RoundChangeDataInvalidPreparedRound(),
-	messages.RoundChangeDataInvalidPreparedValue(),
+	// TODO<olegshmuelov>: rc not prepared with data
+	//messages.RoundChangeDataInvalidPreparedRound(),
```
olegshmuelov, line 105: failing test should be resolved
https://github.com/bloxapp/ssv-spec/pull/69/files#r1023970249

olegshmuelov: https://github.com/bloxapp/ssv-spec/issues/103
could be related

qbft/spectest/all_tests.go
```diff
@@ -101,14 +101,14 @@ var AllTests = []SpecTest{
...
 	messages.RoundChangePrePreparedJustifications(),
 	messages.RoundChangeNotPreparedJustifications(),
 	messages.CommitDataEncoding(),
-	messages.MsgNilIdentifier(),
-	messages.MsgNonZeroIdentifier(),
-	messages.MsgTypeUnknown(),
+	//messages.MsgNilIdentifier(),
+	//messages.MsgNonZeroIdentifier(),
```
olegshmuelov, line 110: find a way to validate the identifier

ssv/spectest/alltests.go
```diff
@@ -25,10 +25,12 @@ var AllTests = []SpecTest{
 	newduty.Finished(),
 	newduty.Valid(),
 	newduty.PostWrongDecided(),
-	newduty.PostInvalidDecided(),
+	// TODO<olegshmuelov>: we cant marshal consensus data with unknown role
+	//newduty.PostInvalidDecided(),
```
olegshmuelov, line 29: marshaling/unmarshaling consensus data with unknown role will fail on decoding

ssv/spectest/alltests.go
```diff
@@ -25,10 +25,12 @@ var AllTests = []SpecTest{
...
 
 	consensus.FutureDecided(),
-	consensus.InvalidDecidedValue(),
+	// TODO<olegshmuelov>: we cant marshal consensus data with unknown role
+	//consensus.InvalidDecidedValue(),
```
olegshmuelov, line 33: marshaling/unmarshaling consensus data with unknown role will fail on decoding

types/consensus_data.go
```diff
@@ -49,13 +62,204 @@ type ConsensusData struct {
 	AggregateAndProof      *phase0.AggregateAndProof
 	SyncCommitteeBlockRoot phase0.Root
 	// SyncCommitteeContribution map holds as key the selection proof for the contribution
+	// TODO<olegshmuelov>: consider to use array instead of the map for ssz
 	SyncCommitteeContribution ContributionsMap
```
olegshmuelov, line 66: we should consider to use array instead of the map for ssz.
in this scenario we have to use temporary ssz struct to convert the map to sorted array in order to have the same struct after marshal/unmarshal ssz

PR #82, opened 2022-10-12
Reshare support

gl-rockx:


dkg/messages.go
```diff
@@ -113,6 +116,11 @@ func (signedMsg *SignedMessage) GetRoot() ([]byte, error) {
 	return signedMsg.Message.GetRoot()
 }
 
+type InitOrReshare struct {
```
alonmuroch, line 119: I'd separate that into 2 distinct messages ..

dkg/messages.go
```diff
@@ -150,6 +158,55 @@ func (msg *Init) Decode(data []byte) error {
 	return json.Unmarshal(data, msg)
 }
 
+// Reshare triggers the resharing protocol
+type Reshare struct {
+	// ValidatorPK is the the public key to be reshared
+	ValidatorPK types.ValidatorPK
+	// EncryptedShares contains the encrypted shares of the old set
+	EncryptedShares map[types.OperatorID][]byte
```
alonmuroch, line 166: The node should already have that, no need to pass it again

dkg/messages.go
```diff
@@ -150,6 +158,55 @@ func (msg *Init) Decode(data []byte) error {
...
+	// ThresholdOld is the threshold of the old set
+	ThresholdOld uint16
```
alonmuroch, line 168: should already hav that info

dkg/messages.go
```diff
@@ -150,6 +158,55 @@ func (msg *Init) Decode(data []byte) error {
...
+	// OperatorIDs are the operators in the new set
+	OperatorIDs []types.OperatorID
+	// Threshold is the threshold of the new set
+	Threshold uint16
+}
+
+func (msg *Reshare) Validate() error {
+
+	if len(msg.ValidatorPK) != blsPubkeySize {
```
alonmuroch, line 177: why not make validatorpk byte[48] and that's it?

dkg/node.go
```diff
@@ -51,13 +51,33 @@ func (n *Node) newRunner(id RequestID, initMsg *Init) (*Runner, error) {
...
 
-	if err := runner.protocol.Start(initMsg); err != nil {
+	if err := runner.protocol.Start(InitOrReshare{Init: initMsg}); err != nil {
 		return nil, errors.Wrap(err, "could not start dkg protocol")
 	}
 
 	return runner, nil
 }
 
+func (n *Node) newResharingRunner(id RequestID, reshareMsg *Reshare) (*Runner, error) {
```
alonmuroch, line 61: maybe use runner as interface?

dkg/spectest/tests/resharing.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"github.com/bloxapp/ssv-spec/dkg"
+	"github.com/bloxapp/ssv-spec/types"
+	"github.com/bloxapp/ssv-spec/types/testingutils"
+)
+
+// ResharingHappyFlow tests a simple (dummy) resharing flow, the difference between this and keygen happy flow is
+// resharing doesn't sign deposit data
+func ResharingHappyFlow() *MsgProcessingSpecTest {
```
alonmuroch, line 11: wasn't added to Alltests so not sure it's running

alonmuroch (changes requested):

dkg/messages.go
gl-rockx: The intention of adding this is so that this struct is self sufficient. The information can be fed by node. The protocol is agnostic about whether the node grabs the info from local storage or the smart contract. I think previously we included EncryptedShares so that it can be recovered from the smart contract. If we exclude this, the node then has to get the plaintext share and feed to the resharing protocol.

dkg/messages.go
gl-rockx: The node can fetch the info and prepare this struct and pass it to the protocol. Otherwise the protocol need to fetch directly (from storage)?

dkg/messages.go
gl-rockx: It's an SSV original type we didn't attempt to change.

dkg/messages.go
gl-rockx: Combining these two into one is because the `protocol` interface has one method `Start(initOrReshare InitOrReshare) error` which accepts one message. It's confusing also to pass two arguments where one has to be `nil`. I would think a more logical way is to have pass `Init` or `Reshare` in `New` and `Start` will not need to take the input. Think about `Init` / `Reshare` triggers the creation of the protocol instance while `Start` kicks off the protocol.

dkg/spectest/tests/resharing.go
gl-rockx: Enabled in https://github.com/bloxapp/ssv-spec/pull/82/commits/5a7d054b437990a8434fa20d5d3dc322c046ea1b

dkg/node.go
gl-rockx: Done in https://github.com/bloxapp/ssv-spec/pull/82/commits/a5908c6a0f04d8ad440b08579ee8b82e3490ac0c

dkg/messages.go
alonmuroch: Sure, suggest a change and lets do it

dkg/messages.go
alonmuroch: Yeah the issue with it is that you will need to add checks that it matches what the node knows anyhow so just another complication

dkg/messages.go
gl-rockx: Removed  these two fields and instead node will retrieve the `KeyGenOutput` from local storage.  https://github.com/bloxapp/ssv-spec/pull/82/commits/dbce44e474475b6485d877e6f7f2a9de70b11c18

dkg/messages.go
gl-rockx: Made the changes in https://github.com/bloxapp/ssv-spec/pull/82/commits/4d0e75bd6f3648d4ecd6ae92fea16064e572a818 and https://github.com/bloxapp/ssv-spec/pull/82/commits/30d3461b1202c1ea345831bccf3c148b6f1a8260

dkg/messages.go
```diff
@@ -150,6 +153,43 @@ func (msg *Init) Decode(data []byte) error {
...
+		return errors.New("invalid validator pubkey size")
+	}
+
+	if len(msg.OperatorIDs) < 4 || (len(msg.OperatorIDs)-1)%3 != 0 {
+		return errors.New("invalid number of operators which has to be 3f+1")
+	}
+
+	if int(msg.Threshold) != (len(msg.OperatorIDs)-1)*2/3+1 {
```
alonmuroch, line 176: Let's change operator count validation to 4,7,10,13 and threshold accordingly

alonmuroch (approved):

dkg/messages.go
gl-rockx: Sure. Will do.

PR #87, opened 2022-10-31
Network Interface Alignment

amir-blox:
**Changed**

- removed `BroadcastDecided()`
- generalized broadcaster interface
- syncer: changed to typed msg id

### DIFF
--- a/p2p/interface.go
+++ b/p2p/interface.go
@@ -4,7 +4,7 @@ import "github.com/bloxapp/ssv-spec/types"
 
 // Broadcaster is the interface used to abstract message broadcasting
 type Broadcaster interface {
-	Broadcast(message types.Encoder) error
+	Broadcast(message *types.SSVMessage) error
 }
 
 // Subscriber is used to abstract topic management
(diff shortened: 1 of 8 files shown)


p2p/interface.go
```diff
@@ -0,0 +1,13 @@
+package p2p
+
+import "github.com/bloxapp/ssv-spec/types"
+
+// Broadcaster is the interface used to abstract message broadcasting
+type Broadcaster interface {
+	Broadcast(message types.Encoder) error
```
alonmuroch, line 7: this means that the wrapping into ssv message needs to be done in the network implementation? 
I think it's ambiguous which type of msg are we expecting here? 
Maybe just enforce SSVMessage and that's it?

alonmuroch (changes requested):

p2p/interface.go
amir-blox: Agreed, changed in https://github.com/bloxapp/ssv-spec/pull/87/commits/7e61689470710bc09f2ec5a108363c6c7ec467c4

alonmuroch:
@amir-blox please rebase and squash for merge

lior-blox (approved):

alonmuroch (approved):

PR #89, opened 2022-11-03
instance container interface + inMem imp

nivBlox:
Add instance container interface and a inMemory implantation for the qbft controller to use.

### DIFF
--- a/qbft/controller.go
+++ b/qbft/controller.go
@@ -13,27 +13,6 @@ import (
 // guaranteed to arrive in a timely fashion, we physically limit how far back the processmsg will process messages for
 const HistoricalInstanceCapacity int = 5
 
-type InstanceContainer [HistoricalInstanceCapacity]*Instance
-
-func (i InstanceContainer) FindInstance(height Height) *Instance {
-	for _, inst := range i {
-		if inst != nil {
-			if inst.GetHeight() == height {
-				return inst
-			}
-		}
-	}
-	return nil
-}
-
-// addNewInstance will add the new instance at index 0, pushing all other stored InstanceContainer one index up (ejecting last one if existing)
-func (i *InstanceContainer) addNewInstance(instance *Instance) {
-	for idx := HistoricalInstanceCapacity - 1; idx > 0; idx-- {
-		i[idx] = i[idx-1]
-	}
-	i[0] = instance
-}
-
 // Controller is a QBFT coordinator responsible for starting and following the entire life cycle of multiple QBFT InstanceContainer
 type Controller struct {
 	Identifier []byte
@@ -58,7 +37,7 @@ func NewController(
 		Height:              -1, // as we bump the height when starting the first instance
 		Domain:              domain,
 		Share:               share,
-		StoredInstances:     InstanceContainer{},
+		StoredInstances:     NewInMemContainer(HistoricalInstanceCapacity),
 		FutureMsgsContainer: make(map[types.OperatorID]Height),
 		config:              config,
 	}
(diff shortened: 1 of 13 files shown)


qbft/instance_store.go
```diff
@@ -0,0 +1,84 @@
...
+func (i *inMemContainer) AddNewInstance(instance *Instance) {
+	for idx := i.capacity - 1; idx > 0; idx-- {
+		i.container[idx] = i.container[idx-1]
+	}
+	i.container[0] = instance
+}
+
+// AddNewInstanceAtPosition adding instance at specific position in store
+func (i *inMemContainer) AddNewInstanceAtPosition(position int, instance *Instance) {
+	if position >= i.capacity {
```
amir-blox, line 66: Check negative values

qbft/instance_store.go
```diff
@@ -0,0 +1,84 @@
+package qbft
+
+import "encoding/json"
+
+// InstanceContainer interface for managing multi instance lifecycle
+type InstanceContainer interface {
+	All() []*Instance
+	FindInstance(height Height) *Instance
+	FindInstanceByPosition(position int) *Instance
```
alonmuroch, line 9: what is position?

qbft/controller.go
```diff
@@ -58,7 +37,7 @@ func NewController(
 		Height:              -1, // as we bump the height when starting the first instance
 		Domain:              domain,
 		Share:               share,
-		StoredInstances:     InstanceContainer{},
+		StoredInstances:     NewInMemContainer(HistoricalInstanceCapacity),
```
alonmuroch, line 40: It should receive the instance via the func params not hardcoded

qbft/testing_utils.go
```diff
@@ -92,11 +92,18 @@ var testingInstanceStruct = &Instance{
...
 	},
 }
-var testingControllerStruct = &Controller{
-	Identifier: []byte{1, 2, 3, 4},
-	Height:     Height(1),
-	Share:      testingShare,
-	StoredInstances: [HistoricalInstanceCapacity]*Instance{
-		testingInstanceStruct,
-	},
+var testingControllerStruct = func() *Controller {
```
alonmuroch, line 95: who calls this?

qbft/instance_store.go
```diff
@@ -0,0 +1,84 @@
+package qbft
```
alonmuroch, line 1: this should be in testing utils not here

qbft/controller.go
```diff
@@ -227,14 +206,50 @@ func (c *Controller) Decode(data []byte) error {
...
 
+// UnmarshalJSON add a custom marshaling for Controller
+func (c *Controller) UnmarshalJSON(data []byte) error {
+	var objmap map[string]*json.RawMessage
+	if err := json.Unmarshal(data, &objmap); err != nil {
+		return err
+	}
+	fmt.Println(objmap)
+
+	var imc inMemContainer
```
alonmuroch, line 225: That shouldn't be here.. spec is not implementation specific

alonmuroch (changes requested):

alonmuroch:
I don't think this approach is good... it complicates things on multiple fronts. 
Might better have the store outside of controller marshaling/ root struct or something but that will cause issues with tests

qbft/controller.go
nivBlox: just aligned the old code.. no change has made

qbft/testing_utils.go
nivBlox: qbft/controller_test.go:72

qbft/instance_store.go
nivBlox: why testing? 
i can separate the interface and the implementation if that what you mean

qbft/controller.go
nivBlox: but it is implemented an in memory before. the other option is to not use interface

nivBlox:
> that

the previews container is rooted in the code & tests 
without the marshaling the instance store will not pass to the encoded bytes

nivBlox:
*DO NOT MERGE*

Use a different way to implement loading instance from storage

PR #96, opened 2022-11-08
Instance container to storage

nivBlox:
Removed instance container and use storage instead to save instance state. 

Controller root has been change so need to change the roots in most tests.


qbft/controller.go
```diff
@@ -139,7 +112,9 @@ func (c *Controller) baseMsgValidation(msg *SignedMessage) error {
 }
 
 func (c *Controller) InstanceForHeight(height Height) *Instance {
-	return c.StoredInstances.FindInstance(height)
+	state, _ := c.GetConfig().GetStorage().GetInstanceState(c.Identifier, height)
+	// TODO need to handle err?
```
alonmuroch, line 116: handle error

qbft/controller.go
```diff
@@ -154,14 +129,21 @@ func (c *Controller) GetIdentifier() []byte {
 // addAndStoreNewInstance returns creates a new QBFT instance, stores it in an array and returns it
 func (c *Controller) addAndStoreNewInstance() *Instance {
 	i := NewInstance(c.GetConfig(), c.Share, c.Identifier, c.Height)
-	c.StoredInstances.addNewInstance(i)
+	err := c.GetConfig().GetStorage().SaveInstanceState(i.State)
+	if err != nil {
```
alonmuroch, line 133: handle error

qbft/controller.go
```diff
@@ -221,7 +177,9 @@ func (c *Controller) Encode() ([]byte, error) {
 
 // Decode implementation
 func (c *Controller) Decode(data []byte) error {
-	err := json.Unmarshal(data, &c)
```
alonmuroch, line 224: cleanup

qbft/types.go
```diff
@@ -32,6 +32,12 @@ type Storage interface {
 	SaveHighestDecided(signedMsg *SignedMessage) error
 	// GetHighestDecided returns highest decided if found, nil if didn't
 	GetHighestDecided(identifier []byte) (*SignedMessage, error)
+	// SaveInstanceState saves instance state by identifier and height
+	SaveInstanceState(state *State) error
+	// GetInstanceState returns instance state
+	GetInstanceState(identifier []byte, height Height) (*State, error)
+	// GetAlInstancesState returns all instances state for identifier
+	GetAlInstancesState(identifier []byte) ([]*State, error)
```
nivBlox, line 40: typo

alonmuroch, line 40: bad practice, shouldn't return all

qbft/controller.go
```diff
@@ -232,7 +190,52 @@ func (c *Controller) Decode(data []byte) error {
 			i.config = config
 		}
 	}
-	return nil
+	return nil*/
+}
+
+// MarshalJSON implementation
+func (c *Controller) MarshalJSON() ([]byte, error) {
```
alonmuroch, line 197: no need for it, instances are outside of controller

alonmuroch (changes requested):

qbft/decided.go
```diff
@@ -28,18 +28,20 @@ func (c *Controller) UponDecided(msg *SignedMessage) (*SignedMessage, error) {
 	}
 
 	// did previously decide?
-	inst := c.InstanceForHeight(msg.Message.Height)
-	prevDecided := inst != nil && inst.State.Decided
+	inst, err := c.InstanceForHeight(msg.Message.Height)
+	prevDecided := err == nil && inst != nil && inst.State.Decided
```
alonmuroch, line 32: error needs to be checked independently

qbft/decided.go
```diff
@@ -28,18 +28,20 @@ func (c *Controller) UponDecided(msg *SignedMessage) (*SignedMessage, error) {
...
 
 	// Mark current instance decided
-	if inst := c.InstanceForHeight(c.Height); inst != nil && !inst.State.Decided {
+	if inst, err := c.InstanceForHeight(c.Height); inst != nil && !inst.State.Decided {
```
alonmuroch, line 35: error needs to be checked independently

qbft/decided.go
```diff
@@ -28,18 +28,20 @@ func (c *Controller) UponDecided(msg *SignedMessage) (*SignedMessage, error) {
...
 		inst.State.Decided = true
 		if msg.Message.Round > inst.State.Round {
 			inst.State.Round = msg.Message.Round
 		}
 		if c.Height == msg.Message.Height {
 			inst.State.DecidedValue = data.Data
 		}
+	} else if err != nil {
+		fmt.Printf("failed to get instance - %e", err)
```
alonmuroch, line 44: error handling

qbft/decided.go
```diff
@@ -49,8 +51,9 @@ func (c *Controller) UponDecided(msg *SignedMessage) (*SignedMessage, error) {
 		i.State.Round = msg.Message.Round
 		i.State.Decided = true
 		i.State.DecidedValue = data.Data
-		c.StoredInstances.addNewInstance(i)
-
+		if err := c.GetConfig().GetStorage().SaveInstanceState(i.State); err != nil {
```
alonmuroch, line 54: could cause a case where we decided current instance but saving failed

alonmuroch (changes requested):

alonmuroch:
@nivBlox please open an issue to remove the historical root func and use the actual root

qbft/decided.go
nivBlox: do we want to failed processing msg cause of this error?

qbft/decided.go
nivBlox: same as above, we want to failed all process cause of it?

qbft/decided.go
nivBlox: same as above

qbft/decided.go
nivBlox: thats why returning error

nivBlox:
@alonmuroch 
https://github.com/bloxapp/ssv-spec/issues/99

qbft/decided.go
alonmuroch: yes we do, if we can't fetch it

qbft/decided.go
alonmuroch: yeah but you decide the instance before

nivBlox: any way if there is an error it wont be saved.. 
we can in case of error revert the instance changes to be not decided for example 
WDYT

qbft/controller.go
```diff
@@ -112,6 +90,11 @@ func (c *Controller) UponExistingInstanceMsg(msg *SignedMessage) (*SignedMessage
 		return nil, errors.Wrap(err, "could not process msg")
 	}
 
+	// update instance state
+	if err := c.SaveInstance(inst); err != nil {
+		fmt.Printf("%s\n", err.Error())
```
alonmuroch, line 95: should return error

qbft/controller.go
```diff
@@ -152,16 +144,23 @@ func (c *Controller) GetIdentifier() []byte {
 }
 
 // addAndStoreNewInstance returns creates a new QBFT instance, stores it in an array and returns it
-func (c *Controller) addAndStoreNewInstance() *Instance {
+func (c *Controller) addAndStoreNewInstance() (*Instance, error) {
 	i := NewInstance(c.GetConfig(), c.Share, c.Identifier, c.Height)
-	c.StoredInstances.addNewInstance(i)
-	return i
+	err := c.GetConfig().GetStorage().SaveInstanceState(i.State)
```
alonmuroch, line 149: can simply return i,errors.wrap(c.GetConfig().GetStorage().SaveInstanceState(i.State), "error ")

qbft/decided.go
alonmuroch: should return error

alonmuroch (changes requested):
@nivBlox in terms of error handling since when a msg is processed we call storage to get instance state, any time there is an error staving the state back it should return error. This will cause a kind of revert from any changes the msg did as it's not saved

lior-blox:
closed, not part of spec scope

PR #100, opened 2022-11-13
Reset round timer upon instance start

amir-blox:
Added a call for round timeout upon instance start.

### DIFF
--- a/qbft/instance.go
+++ b/qbft/instance.go
@@ -52,6 +52,8 @@ func (i *Instance) Start(value []byte, height Height) {
 		i.State.Round = FirstRound
 		i.State.Height = height
 
+		i.config.GetTimer().TimeoutForRound(FirstRound)
+
 		// propose if this node is the proposer
 		if proposer(i.State, i.GetConfig(), FirstRound) == i.State.Share.OperatorID {
 			proposal, err := CreateProposal(i.State, i.config, i.StartValue, nil, nil)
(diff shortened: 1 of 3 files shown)


qbft/instance.go
```diff
@@ -65,6 +65,8 @@ func (i *Instance) Start(value []byte, height Height) {
 			}
 		}
 
+		i.config.GetTimer().TimeoutForRound(FirstRound)
```
alonmuroch, line 68: I wonder if this needs to come before the proposer if

qbft/spectest/tests/controller_spectest.go
```diff
@@ -42,6 +42,11 @@ func (test *ControllerSpecTest) Run(t *testing.T) {
 		err := contr.StartNewInstance(runData.InputValue)
 		if err != nil {
 			lastErr = err
+		} else {
```
alonmuroch, line 45: what this suppose to test?

alonmuroch (changes requested):

alonmuroch:
rebase and squash

qbft/spectest/tests/controller_spectest.go
amir-blox: That the timer was called upon start

nivBlox:
@amir-blox no missing the timeout call to function?

amir-blox:
@nivBlox the instance exposes `UponRoundTimeout()` which we should call from outside of spec.

alonmuroch:
@amir-blox rebase  and squash

PR #117, opened 2022-11-24
ssv,qbft/ refactor storage and height

### LINKED ISSUE #63: first round should be 0?

### LINKED ISSUE #115: remove storage from spec
We remove all storage from spec, will be handled by implementation in any way they see fit

- [ ] QBFT
- [ ] SSV

alonmuroch:
closes #115 
closes #63


qbft/spectest/tests/controller/decided/current_instance.go
```diff
@@ -74,16 +74,6 @@ func CurrentInstance() *tests.ControllerSpecTest {
 							Data:       testingutils.CommitDataBytes([]byte{1, 2, 3, 4}),
 						}),
 				},
-				SavedDecided: testingutils.MultiSignQBFTMsg(
```
amir-blox, line 77: don't we want to check that these messages were at least sent to broadcast?

alonmuroch: in this particular test we don't broadcast anything

PR #134, opened 2022-12-15
`MessageID.String`: human-friendly format

moshe-blox:
### Rationale

- More human-friendly `ATTESTER` rather than `00000000`.
- Logs can be easily filtered by public key, role or a combination of both (thanks to dash between them.)

### Change
Previously returned `{PublicKeyHex}{RoleHex}`
Now returns `{PublicKeyHex}-{RoleString}`

### Example
~`917494a7348b314221e4df2b22347bfaee841781fa8b11801418b1454045b8a30065be03983a7667275850e472afcff700000000`~
`917494a7348b314221e4df2b22347bfaee841781fa8b11801418b1454045b8a30065be03983a7667275850e472afcff7-ATTESTER`

### DIFF
--- a/types/messages.go
+++ b/types/messages.go
@@ -49,7 +49,7 @@ func NewMsgID(pk []byte, role BeaconRole) MessageID {
 }
 
 func (msgID MessageID) String() string {
-	return hex.EncodeToString(msgID[:])
+	return hex.EncodeToString(msgID.GetPubKey()) + "-" + msgID.GetRoleType().String()
 }
 
 func MessageIDFromBytes(mid []byte) MessageID {


alonmuroch:
Seems to me like an implementation specific thing, why can't you do a "prettyString" func in implementation for that?

moshe-blox:
> Seems to me like an implementation specific thing, why can't you do a "prettyString" func in implementation for that?

We can, but throwing human-readable output in the `Stringer` interface is more powerful, because nearly any printing method in Go calls the `String` method.

Specifically for us, this is something we often do:

`zap.Any("message", message)`

And since `zap` just calls `String()` on `message.Identifier`, we see the `00000000` output in logs, and I don't know of any way to make `zap` call our custom `prettyString` method instead

moshe-blox:
We've done a big logging refactor in SSV, so this PR isn't necessary anymore.

PR #138, opened 2022-12-20
qbft/ Round validation for consensus messages

amir-blox:
Fail base msg validation (instance) in case we notice a message with past round, before doing a more extensive validation.

### DIFF
--- a/qbft/spectest/tests/controller/latemsg/late_round_change_past_round.go
+++ b/qbft/spectest/tests/controller/latemsg/late_round_change_past_round.go
@@ -126,7 +126,7 @@ func LateRoundChangePastRound() *tests.ControllerSpecTest {
 
 	return &tests.ControllerSpecTest{
 		Name:          "late round change past round",
-		ExpectedError: "could not process msg: invalid signed message: wrong msg round",
+		ExpectedError: "could not process msg: invalid signed message: past round",
 		RunInstanceData: []*tests.RunInstanceData{
 			{
 				InputValue:    []byte{1, 2, 3, 4},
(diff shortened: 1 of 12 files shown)


alonmuroch:
@amir-blox not sure what this solves?

amir-blox:
@alonmuroch there are several issues that we solve here:
1. using a constant timeout of 2sec is flooding the network with change round messages
2. avoiding validation of the round for round change messages is causing the node to process all these messages and broadcast a message for each round with partial quorum
3. because we are starting to validate round, we need to align the validation to accept future rounds, otherwise they'll be discarded

qbft/spectest/tests/controller/latemsg/late_round_change_past_round.go
```diff
@@ -144,8 +145,7 @@ func LateRoundChangePastRound() *tests.ControllerSpecTest {
 							Data:       testingutils.CommitDataBytes([]byte{1, 2, 3, 4}),
 						}),
 				},
-
-				ControllerPostRoot: "bb2e61b20360347bf1c48c31e78069a26c59cfd0e06ada43ba799966b0f8b9fd",
+				ControllerPostRoot: "b0fd6098e01e0bee8cac21654e8a6c0f6a0b0634499146fbb722b9af2c459c2d",
```
alonmuroch, line 148: why the root change?

alonmuroch (changes requested):

qbft/spectest/tests/controller/latemsg/late_round_change_past_round.go
amir-blox: `Instance.State.RoundChangeContainer` used to hold past rounds, now it doesn't

alonmuroch (approved):

PR #139, opened 2022-12-25
ssv/ Blinded block support

alonmuroch:


alonmuroch:
Gaps
- determine blinded/ not blinded block after decided 
- We will support fee recipient signature by having a new duty type running every X epochs a simple sign ValidatorRegistration object. If account fee recipient changes it will change the next time the nodes sign ValidatorRegistration. A node coming online will know the account fee recipient but will get the fee signature the next time they sign it.

PR #147, opened 2023-01-02
ssv/ blinded block support

alonmuroch:


lior-blox (approved):

ssv/proposer.go
```diff
@@ -153,25 +175,44 @@ func (r *ProposerRunner) ProcessPostConsensus(signedMsg *SignedPartialSignatureM
...
 		}
 	}
 	r.GetState().Finished = true
 	return nil
 }
 
+// decidedBlindedBlock returns true if decided value has a blinded block, false if regular block
+// WARNING!! should be called after decided only
+func (r *ProposerRunner) decidedBlindedBlock() bool {
+	return r.BaseRunner.State.DecidedValue.BlindedBlockData != nil
```
nivBlox, line 203: late msg could be received and will result in panic. i think we should add check for decided value here

PR #150, opened 2023-01-12
Slashing protection adjustments

olegshmuelov:


types/signer.go
```diff
@@ -31,11 +31,11 @@ var (
 
 type BeaconSigner interface {
 	// SignBeaconObject returns signature and root.
-	SignBeaconObject(obj ssz.HashRoot, domain spec.Domain, pk []byte) (Signature, []byte, error)
+	SignBeaconObject(obj ssz.HashRoot, domain spec.Domain, pk []byte, role BeaconRole) (Signature, []byte, error)
```
alonmuroch, line 34: why do we need to pass role?

ssv/value_check.go
```diff
@@ -39,6 +39,7 @@ func AttesterValueCheckF(
 	network types.BeaconNetwork,
 	validatorPK types.ValidatorPK,
 	validatorIndex phase0.ValidatorIndex,
+	sharePublicKey []byte,
```
alonmuroch, line 42: that should probably be the validator pk no?

alonmuroch (changes requested):

types/signer.go
olegshmuelov: we want to be sure that there is no slashing event before signing. 

here example:
https://github.com/bloxapp/ssv/blob/6b4cde4cc1774d3b07e605d1782bd75b4e0848e4/ekm/eth_key_manager_signer.go#L89

the slashing check only for attester & proposer roles

ssv/value_check.go
olegshmuelov: We can use validator PK.
in this scenario we have to
 -  adjust AddShare signature to pass the validator PK
https://github.com/bloxapp/ssv-spec/blob/17ac6f1047cea3697b60e677d8c5485d9f9ad627/types/signer.go#L59
and store share private key by validator key and not share public key
 - adjust SignBeaconObject sig to pass validator PK
 https://github.com/bloxapp/ssv-spec/blob/17ac6f1047cea3697b60e677d8c5485d9f9ad627/types/signer.go#L34
 in order to check slashing protection before signing

types/signer.go
olegshmuelov: in theory we can check the slashing protection after we signed the object

ssv/value_check.go
alonmuroch: Probably best left to when we redo the remote signer

alonmuroch (approved):

PR #158, opened 2023-02-16
Versioned Beacon Block

nivBlox:


go.mod
```diff
@@ -18,55 +18,24 @@ require (
 
 require (
 	github.com/benbjohnson/clock v1.3.0 // indirect
-	github.com/btcsuite/btcd v0.22.1 // indirect
```
alonmuroch, line 21: these shouldn't change

ssv/proposer.go
```diff
@@ -124,12 +122,7 @@ func (r *ProposerRunner) ProcessConsensus(signedMsg *qbft.SignedMessage) error {
 	}
 
 	// specific duty sig
-	var blkToSign ssz.HashRoot
-	if r.decidedBlindedBlock() {
-		blkToSign, err = decidedValue.GetBellatrixBlindedBlockData()
-	} else {
-		blkToSign, err = decidedValue.GetBellatrixBlockData()
-	}
+	blkToSign, err := decidedValue.GetBlockRoot()
```
alonmuroch, line 125: should probably be in this file not consensus data

ssv/proposer.go
```diff
@@ -188,30 +181,20 @@ func (r *ProposerRunner) ProcessPostConsensus(signedMsg *types.SignedPartialSign
 		specSig := phase0.BLSSignature{}
 		copy(specSig[:], sig)
 
-		if r.decidedBlindedBlock() {
-			data, err := r.GetState().DecidedValue.GetBellatrixBlindedBlockData()
+		if r.GetState().DecidedValue.DecidedBlindedBlock() {
```
alonmuroch, line 184: should be in this file not consensus data as it's very specific to this runner

ssv/proposer.go
```diff
@@ -188,30 +181,20 @@ func (r *ProposerRunner) ProcessPostConsensus(signedMsg *types.SignedPartialSign
...
+			versionedBlindedBlock, err := r.GetState().DecidedValue.GetVersionedBlindedBlock(specSig)
```
alonmuroch, line 185: same here

alonmuroch (changes requested):

ssv/proposer.go
nivBlox: we use this func in more than 1 place

ssv/proposer.go
nivBlox: same line as above

go.mod
nivBlox: reverted

ssv/proposer.go
nivBlox: done

lior-blox (changes requested):

PR #159, opened 2023-02-17
Fix checks if H(data) != root

nkryuchkov:
In all cases of `errors.Wrap(err, "H(data) != root")` `err` is always `nil` (the result of `HashDataRoot`), so `errors.Wrap` always returns `nil` according to https://pkg.go.dev/github.com/pkg/errors#Wrap

### DIFF
--- a/qbft/proposal.go
+++ b/qbft/proposal.go
@@ -77,7 +77,7 @@ func isValidProposal(
 		return errors.Wrap(err, "could not hash input data")
 	}
 	if !bytes.Equal(signedProposal.Message.Root[:], r[:]) {
-		return errors.Wrap(err, "H(data) != root")
+		return errors.New("H(data) != root")
 	}
 
 	// get justifications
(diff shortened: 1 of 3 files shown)


aaron-blox (approved):

qbft/proposal.go
```diff
@@ -77,7 +77,7 @@ func isValidProposal(
 		return errors.Wrap(err, "could not hash input data")
 	}
 	if !bytes.Equal(signedProposal.Message.Root[:], r[:]) {
-		return errors.Wrap(err, "H(data) != root")
+		return errors.New("H(data) != root")
```
aaron-blox, line 80: you can also do a global error var like this, if you do it please choose a better variable name

`ErrHashDataIsNotRoot = errors.New("H(data) != root")` and then
```suggestion
		return ErrHashDataIsNotRoot
```

nkryuchkov: @aaron-blox I agree that it would reduce code repetitions but I didn't find any global errors in spec, so it would be the first one, which, I think, would make the error handling style a bit inconsistent throughout the project. I think, if we do that, we need to do that in a separate PR and search for other errors that are repeated.

@alonmuroch Your thoughts?

PR #189, opened 2023-04-19
message max size fixes + calculation

alonmuroch:
Fixes message ssz max size + explanation calculation


types/messages.go
```diff
@@ -103,7 +103,14 @@ type MessageSignature interface {
...
 	MsgID   MessageID `ssz-size:"56"`
-	Data    []byte    `ssz-max:"1074003968"` // 2^30+2^18
+	// Data max size as following
+	// Needs to support the largest ConsensusData possible which is
+	//		Duty(8 + 48 + 6*8 + 13*8)
+	//		Version (8)
+	// 		13*SignedPartialSignatureMessage(8 + 96 + PartialSignatureMessages(2*8 + 13*PartialSignatureMessage(96+32+8)))
+	//		DataSSZ max 2^22 = 4,194,304
+	// Max size is 4,219,064 = 2^22 + 2^15
+	Data []byte `ssz-max:"4219064"` // 2^22 + 2^15
```
GalRogozinski, line 113: maybe I am misunderstanding something.
I see that `Data` max-size is the same as `FullData` size in `SignedMessage`.

I thought that `Data` contains `SignedMessage` in its entirity. This means it contains `FullData` and more stuff. So how can it be equal?

alonmuroch: good catch, fixed

alonmuroch:
@GalRogozinski @y0sher @lior-blox merging if no critical issue found

GalRogozinski:
@alonmuroch 
I am giving this another look now

types/messages.go
```diff
@@ -103,7 +103,14 @@ type MessageSignature interface {
...
 	MsgID   MessageID `ssz-size:"56"`
-	Data    []byte    `ssz-max:"1074003968"` // 2^30+2^18
+	// Data max size as following
+	// Needs to support the largest ConsensusData possible which is
+	//		Duty(8 + 48 + 6*8 + 13*8)
+	//		Version (8)
+	// 		13*SignedPartialSignatureMessage(8 + 96 + PartialSignatureMessages(2*8 + 13*PartialSignatureMessage(96+32+8)))
+	//		DataSSZ max 2^22 = 4,194,304
+	// Max size is 4,219,064 = 2^22 + 2^15
+	Data []byte `ssz-max:"8421376"` // 2^23 + 2^15
```
GalRogozinski, line 113: it should be 2^22 or 2^23?

GalRogozinski, line 113: Can you elaborate why 2^23?
Why did you multiply by another 2

In case you want to add `SignedMessage` over head it should be an extra:
Sig(96) + Signers(13*8)+Message(8 + 8 + 8 + 56 + 32 + 8 + 2^16 + 2^16) = 2^17 + 320

So final max size should be 2^23 + 2^17 + 2^15 + 320.

Am I missing some case?

alonmuroch: pushed fixes + explanations

PR #190, opened 2023-04-19
State comparison - QBFT controller/decided

olegshmuelov:
Introduce state comparison for sanity QBFT controller/decided tests


qbft/spectest/comparable/types.go
```diff
@@ -2,33 +2,64 @@ package qbftcomparable
...
 	"github.com/bloxapp/ssv-spec/qbft"
 )
 
-var RootRegister = map[string]*qbft.State{}
+var RootRegister = map[string]*StateComparison{}
 
 type StateComparison struct {
-	PostState *qbft.State
+	PostState      *qbft.State
+	PostController *qbft.Controller
```
GalRogozinski, line 13: Maybe create some interface that can represent both objects?

qbft/spectest/comparable/types.go
```diff
@@ -2,33 +2,36 @@ package qbftcomparable
...
+
 	"github.com/bloxapp/ssv-spec/qbft"
 )
 
-var RootRegister = map[string]*qbft.State{}
+var RootRegister = map[string]*qbft.RootGetter{}
 
 type StateComparison struct {
-	PostState *qbft.State
+	qbft.RootGetter
```
GalRogozinski, line 12: Make it a pointer like Alon did.
Maybe also give the variable some meaningful name?

`PostTestedObjectState` or something (it is kind of long, other ideas are welcome)? 

Also can you explain to me why we can't have a general StateComparison? Why must it be under QBFT?

qbft/spectest/tests/controller/decided/late_decided.go
```diff
@@ -1,18 +1,19 @@
...
 	"github.com/bloxapp/ssv-spec/types/testingutils"
-	"github.com/herumi/bls-eth-go-binary/bls"
 )
 
 // LateDecided tests processing a decided msg for a just decided instance
 func LateDecided() tests.SpecTest {
 	ks := testingutils.Testing4SharesSet()
-	msgs := testingutils.DecidingMsgsForHeightWithRoot(testingutils.TestingQBFTRootData,
-		testingutils.TestingQBFTFullData, testingutils.TestingIdentifier, qbft.FirstHeight, ks)
+	msgs := testingutils.DecidingMsgsForHeightWithRoot(testingutils.TestingQBFTRootData, testingutils.TestingQBFTFullData, testingutils.TestingIdentifier, qbft.FirstHeight, ks)
```
GalRogozinski, line 16: first height?
Just wondering what is late here?

qbft/spectest/tests/controller/decided/late_decided_bigger_quorum.go
```diff
@@ -1,18 +1,19 @@
...
 	"github.com/bloxapp/ssv-spec/types/testingutils"
-	"github.com/herumi/bls-eth-go-binary/bls"
 )
 
 // LateDecidedBiggerQuorum tests processing a decided msg for a just decided instance (with a bigger quorum)
 func LateDecidedBiggerQuorum() tests.SpecTest {
 	ks := testingutils.Testing4SharesSet()
-	msgs := testingutils.DecidingMsgsForHeightWithRoot(testingutils.TestingQBFTRootData,
-		testingutils.TestingQBFTFullData, testingutils.TestingIdentifier, qbft.FirstHeight, ks)
+	msgs := testingutils.DecidingMsgsForHeightWithRoot(testingutils.TestingQBFTRootData, testingutils.TestingQBFTFullData, testingutils.TestingIdentifier, qbft.FirstHeight, ks)
```
GalRogozinski, line 16: ditto first height

qbft/spectest/tests/controller/decided/current_instance.go
```diff
@@ -32,8 +34,52 @@ func CurrentInstance() tests.SpecTest {
...
 					DecidedVal: testingutils.TestingQBFTFullData,
 				},
-				ControllerPostRoot: "9a5519c1584d326e7c20cb1a824af5dc7139bac2f4d1ced2ff13b5aacfe9ab33",
+				ControllerPostRoot: currentInstanceStateComparison().Register().Root(),
 			},
 		},
 	}
 }
+
+func currentInstanceStateComparison() *qbftcomparable.StateComparison {
```
alonmuroch, line 43: What do you think about separating to a file called current_instance_sc.go? For readability

qbft/spectest/tests/controller/decided/current_instance.go
```diff
@@ -32,8 +34,52 @@ func CurrentInstance() tests.SpecTest {
...
+	identifier := []byte{1, 2, 3, 4}
+	config := testingutils.TestingConfig(testingutils.Testing4SharesSet())
```
alonmuroch, line 45: for readability we might want to structure this code nicer, maybe inline instantiation so it reads better

qbft/spectest/tests/controller/decided/invalid_val_check.go
```diff
@@ -32,8 +35,42 @@ func InvalidValCheckData() tests.SpecTest {
 					DecidedCnt: 1,
 					DecidedVal: testingutils.TestingInvalidValueCheck,
```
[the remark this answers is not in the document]
alonmuroch: Specifically in this case (QBFT) when you have 2f+1 sigs it means there is a majority that decided. That majority might be corrupted but a majority nonetheless

qbft/spectest/tests/controller/decided/late_decided.go
```diff
@@ -25,8 +26,55 @@ func LateDecided() tests.SpecTest {
...
+	state := testingutils.BaseInstance().State
+	state.ProposalAcceptedForCurrentRound = testingutils.TestingProposalMessage(ks.Shares[1], types.OperatorID(1))
+	state.LastPreparedRound = 1
+	state.LastPreparedValue = testingutils.TestingQBFTFullData
+	state.Decided = true
+	state.DecidedValue = testingutils.TestingQBFTFullData
+
+	state.ProposeContainer = &qbft.MsgContainer{Msgs: map[qbft.Round][]*qbft.SignedMessage{
+		qbft.FirstRound: {
+			msgs[0],
```
alonmuroch, line 59: for readability maybe better to have the actual message written

qbft/types.go
```diff
@@ -27,6 +27,10 @@ type Syncer interface {
 	SyncDecidedByRange(identifier types.MessageID, from, to Height)
 }
 
+type RootGetter interface {
```
alonmuroch, line 30: no need for it, you have it in types

alonmuroch (changes requested):

qbft/spectest/tests/controller/decided/current_instance.go
olegshmuelov: Separating it into a separate file will cause a proliferation in the explorer's view, and will make it hard to know what tests we have, and how many of them.

Since we put the `Comparison()` function at the bottom of the file for each test, it doesn't hurt readability too much.
If you still insist we will move it.

qbft/spectest/tests/controller/decided/current_instance.go
olegshmuelov: can u please clarify?

qbft/spectest/tests/controller/decided/late_decided.go
olegshmuelov: It will add more unnecessary work
we are using already build-in msgs to build the state
https://github.com/olegshmuelov/ssv-spec/blob/3a7d5264669f26613f5cade076b50d8c84b7a954/types/testingutils/runner.go#L231

qbft/types.go
olegshmuelov: could not find it, the `RootGetter` interface moved to types package 
in this PR https://github.com/bloxapp/ssv-spec/pull/191

qbft/spectest/tests/controller/decided/late_decided.go
alonmuroch: you have a circular dependency here, if you change something in function creating the array of messages the test will change as well..

qbft/types.go
alonmuroch: under types/messages.go you have an interface called Root. 
No need to duplicate it

qbft/spectest/tests/controller/decided/late_decided.go
olegshmuelov: this specific test using [this function](https://github.com/olegshmuelov/ssv-spec/blob/3a7d5264669f26613f5cade076b50d8c84b7a954/qbft/spectest/tests/controller/decided/late_decided.go#L16) for input messages
if it will be changed for actual test, it should be changed for expected msgs as well.

agree that it can be changed for better readability.
but in some cases e.g. in [/late_prepare_past_instance.go](https://github.com/olegshmuelov/ssv-spec/blob/3a7d5264669f26613f5cade076b50d8c84b7a954/qbft/spectest/tests/controller/latemsg/late_prepare_past_instance.go#L24) we have about 42 messages.
@GalRogozinski

qbft/spectest/tests/controller/decided/invalid_val_check.go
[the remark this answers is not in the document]
GalRogozinski: #201 opened

qbft/spectest/tests/controller/decided/late_decided.go
GalRogozinski: maybe what we can do is add `expectedDecidedMessages` in testutils that have return the array of decided message.
This will keep the current design and break the circular dependency

qbft/spectest/tests/controller/decided/current_instance.go
GalRogozinski: I think he wants to init `contr` in oneline. To me it is less important and is up to @olegshmuelov to decide what is more readable.

qbft/spectest/tests/controller/decided/late_decided.go
GalRogozinski: In general it might be a good idea to do the same even with simple constants (even though one can argue it is an "overkill")
The idea should be that the dev should always change the test and the expected state separately

qbft/spectest/tests/controller/decided/late_decided.go
alonmuroch: // LateDecided tests processing a decided msg for a just decided instance

Maybe the name is wrong but the idea is to test out a received decided message for an already decided instance (via happy flow)

GalRogozinski (approved):

qbft/spectest/tests/controller/decided/late_decided.go
olegshmuelov: https://github.com/bloxapp/ssv-spec/pull/190/commits/2208a961c8eb0f7608e9ef1bfa0d653b322fc46f
https://github.com/bloxapp/ssv-spec/pull/190/commits/7658796119336c2b8e0c538c2ee2534e338fcc17

added expectedDecidedMessages
and made the code cleaner

types/testingutils/runner.go
```diff
@@ -276,3 +277,42 @@ var DecidingMsgsForHeightWithRoot = func(root [32]byte, fullData, msgIdentifier
 	}
 	return msgs
 }
+
+var ExpectedDecidingMsgsForHeightWithRoot = func(root [32]byte, fullData, msgIdentifier []byte, height qbft.Height, keySet *TestKeySet) []*qbft.SignedMessage {
```
y0sher, line 281: can you explain the difference between `DecidingMsgsForHeightWithRoot` and `ExpectedDecidingMsgsForHeightWithRoot`?

y0sher (approved):

qbft/spectest/tests/controller/decided/late_decided.go
alonmuroch: I still have an issue with using testingutils.DecidingMsgsForHeightWithRoot for the state comparison, we are creating circular dependency which cuts against the state comparison. If something happened inside DecidingMsgsForHeightWithRoot the test can be wrong but still pass

olegshmuelov: @alonmuroch take a look on my last comment with the commits links
we are using expectedDecidedMessages as requested.
You are looking on outdated file again

alonmuroch: I don't thin I do... You are getting messages from testingutils.DecidingMsgsForHeightWithRoot and putting them blindly (you don't know what message is in index 1/2/3/etc... 
If testingutils.DecidingMsgsForHeightWithRoot produced the wrong messages you won't know about it

alonmuroch (changes requested):

qbft/spectest/tests/controller/decided/late_decided.go
GalRogozinski: Ok, I just fell for that github bug as well...
Apparently @olegshmuelov is correct that it is outdated. But sometimes Github doesn't show it.

alonmuroch (approved):

PR #191, opened 2023-04-20
Dynamic state comparison

olegshmuelov:


qbft/types.go
```diff
@@ -27,6 +27,10 @@ type Syncer interface {
 	SyncDecidedByRange(identifier types.MessageID, from, to Height)
 }
 
+type RootGetter interface {
+	GetRoot() ([]byte, error)
+}
```
GalRogozinski, line 32: As we discussed.. this may later be moved outside of qbft

GalRogozinski (approved):
Great!
@lior-blox can you have a look at this?

lior-blox:
@GalRogozinski why did the test.json file changed?

GalRogozinski:
@lior-blox 

I checked with @olegshmuelov 
The file is automatically generated.
Apparently the `EncryptBLSSK()` is non deterministic and always generates a random key.

qbft/spectest/comparable/types.go
```diff
@@ -2,31 +2,32 @@ package qbftcomparable
 
 import (
 	"encoding/hex"
+
 	"github.com/bloxapp/ssv-spec/qbft"
 )
 
-var RootRegister = map[string]*qbft.State{}
+var RootRegister = map[string]*qbft.RootGetter{}
```
y0sher, line 9: 1. A little comment about what this is used for would be nice.
2. Why is this a global state variable? this will be an issue or limit us from running tests that use this in parallel, if possible let's not do it as global state.

qbft/types.go
```diff
@@ -27,6 +27,10 @@ type Syncer interface {
 	SyncDecidedByRange(identifier types.MessageID, from, to Height)
 }
 
+type RootGetter interface {
```
y0sher, line 30: this is only used in `qbft/spectest/comparable/types.go`, hence currently I don't see a point why it should be declared here. let's move it to `qbft/spectest/comparable/types.go` for now.

In general the best practice in go is to ACCEPT interfaces and declare them where they are accepted and not return Interfaces if you can and not declare them next to they're concrete types. (this is more of a java style like doing IRootGetter for any class)

types/spectest/generate/tests.json
```diff
@@ -1 +1 @@
...
...NIeyl/MUiPlDMQr0WiygKt6tTFzVN3fzz355A==","ExpectedRoot":[179,115,251,110,55,46,177,200,236,20,174,62,142,228,236,192,195,119,100,208,16,212,156,138,255,44,110,223,13,58,114,109]},"*ssvmsg.EncodingTest_SSVMessage encoding":{"Name":"SSVMessage encoding","Data":"AQAAAAAAAAAAAAMBjoAGZVGoGzGCWHCe2vfdH2PNaGoOTbiym7t6z+ZWCGd69aUn2USO5Hg1SF4CtQvABAAAAEQAAABsAAAAiuAHYeBnoJLPQ4DeKS6iGxaGwMQf3qYwHpyTLnbWI0Putu/2THkdv+/B0mllAGV9COw3lOeAOUA63gyjwQ7X4gBQqDkNccxVyQhtUHTCQOVZj5O7CDrcx0wMxnTp7LnRAQAAAAAAAAAAAAAAAAAAAAwAAAAAAAAAFAAAAIHOTUAW4aKab+DnYipc6IDHDjlZiSGQO5jekCbedMinIW27NQoWqCvP4M1Q0QdSMhaPMr2BH583R8fIBuFRHSRGXrbVQGRMJbhyxw051pAbDo9t6ItlkNEK6VMCHFYoa59BVr5kSYEeTm38fcbD8K3As9jUyxyxqhWSVjHvd+pZAQAAAAAAAACFI30D183QisXzZKefFPWC1wKdLVHdJ8fsYbfIzNTwWMPXVAxBDeShVYGPyv4yneALhKvGPWXFskxyBk6jaDFK99jD5tUW6OCYArESwcL82Cz2OpV3yOZTdKoz+RLRKpr0gCLeQ0LyvqIfzhBbri51Y/fx8Rtwn7Cg/78lKpY5DAEAAAAAAAAAgeoGNlGGzJ2/2j3YX1mLxd7YTzHU/2WIEPnvvwKTf4sYe8gG3KqJT8r+3Uph/mKVANj69KcWGetrcmwC3gZmDnC8JMcXhNKHsoao1LlNW1rzmU9b3FOWkVb0VJQXRithpzXy6V4FAJC0PPbGfBBkUQTq9CKF+UQ+1bmWHSCOBB8BAAAAAAAAAA==","ExpectedRoot":[35,127,95,31,85,239,156,36,192,176,9,106,147,66,178,138,51,163,36,229,98,241,185,201,55,2,156,124,103,186,68,21]}}
```
y0sher, line 1: @olegshmuelov, @GalRogozinski can someone explain me why is this changed?

qbft/types.go
alonmuroch: it already exists in types

qbft/spectest/comparable/types.go
olegshmuelov: 1. Done
2. Unexported variable RootRegister will no longer be accessible from variable diff to print source/target states
example: https://github.com/olegshmuelov/ssv-spec/blob/8e6311c63cdabe08f1b6be1b8f76f21153116a7e/qbft/spectest/tests/msg_processing_spectest.go#L82

qbft/types.go
olegshmuelov: moved to types package because `ssv` package has types that implements the interface.
this also was mentioned before in this comment
https://github.com/bloxapp/ssv-spec/pull/191#discussion_r1172328133

types/spectest/generate/tests.json
olegshmuelov: removed

alonmuroch:
@olegshmuelov under types/messages.go you have an interface called Root. 
No need to duplicate it

types/types.go
```diff
@@ -0,0 +1,5 @@
+package types
```
alonmuroch, line 1: under types/messages.go you have an interface called Root. 
No need to duplicate it

alonmuroch (changes requested):

types/types.go
olegshmuelov: got u.
we will use existing interface but moving to independent file, since it used in qbft/ssv packages as well and not for types/messages only

qbft/spectest/comparable/types.go
y0sher: I don't understand, `RootRegister` is still global state, as it is used in testing it might be accessed and changed in parallel from different tests.

olegshmuelov: got your point.
added PostState to MsgProcessingSpecTest as an example

y0sher: great then! Looks good to me.

GalRogozinski (approved):

qbft/spectest/tests/msg_processing_spectest.go
```diff
@@ -21,6 +20,7 @@ type MsgProcessingSpecTest struct {
 	Name               string
 	Pre                *qbft.Instance
 	PostRoot           string
+	PostState          types.Root `json:"-"` // Field is ignored by encoding/json
```
GalRogozinski, line 23: @alonmuroch, a little question to you (not for this PR):
Is it really bad to include `PostState` in the json?

Even though `PostRoot` is shorter I see an advantage that the json will describe the full post state in a human readable format

qbft/spectest/tests/msg_processing_spectest.go
```diff
@@ -76,8 +78,8 @@ func (test *MsgProcessingSpecTest) Run(t *testing.T) {
 	}
 
 	// test root
-	if test.PostRoot != hex.EncodeToString(postRoot) {
-		diff := comparable3.PrintDiff(test.Pre.State, comparable2.RootRegister[test.PostRoot])
+	if test.PostRoot != hex.EncodeToString(postRoot[:]) {
+		diff := typescomparable.PrintDiff(test.Pre.State, test.PostState)
```
GalRogozinski, line 82: There is a case that PostState may be nil (for tests we didn't do SC, or for tests at `ssv` that run via json

olegshmuelov: in case postState is null, the printDiff will output null for Target

GalRogozinski: I think it may be a bit ugly when running the spec tests against `ssv` but it is not the worst thing.
You have plenty of work so you can choose to either change or not.

qbft/spectest/tests/happy_flow.go
```diff
@@ -32,3 +35,36 @@ func HappyFlow() SpecTest {
 		},
 	}
 }
+
+func happyFlowStateComparison() *qbftcomparable.StateComparison {
```
y0sher, line 39: just for future safety I would accept `ks`, and `state` to this function, just because if someone decides to change them in `HappyFlow` then they won't match.
running this function on the already initialized key shares and state will fool-proof it.

qbft/spectest/tests/timeout/test.go
```diff
@@ -55,5 +57,5 @@ func (test *SpecTest) Run(t *testing.T) {
 
 	postRoot, err := test.Pre.State.GetRoot()
 	require.NoError(t, err)
-	require.EqualValuesf(t, test.PostRoot, hex.EncodeToString(postRoot), "post root not valid")
+	require.EqualValuesf(t, test.PostRoot, hex.EncodeToString(postRoot[:]), "post root not valid")
```
y0sher, line 60: would it be helpful to print the diff here as well?

ssv/spectest/tests/runner/duties/synccommitteeaggregator/proof_spectest.go
```diff
@@ -45,7 +47,7 @@ func (test *SyncCommitteeAggregatorProofSpecTest) Run(t *testing.T) {
 	// post root
 	postRoot, err := r.GetBaseRunner().State.GetRoot()
 	require.NoError(t, err)
-	require.EqualValues(t, test.PostDutyRunnerStateRoot, hex.EncodeToString(postRoot))
+	require.EqualValues(t, test.PostDutyRunnerStateRoot, hex.EncodeToString(postRoot[:]))
```
y0sher, line 50: print diff it fails?

y0sher (approved):
approved but added minor comments

qbft/spectest/tests/timeout/test.go
olegshmuelov: not related to this scope. Will be done only after implementing state comparison to those tests

ssv/spectest/tests/runner/duties/synccommitteeaggregator/proof_spectest.go
olegshmuelov: ditto

qbft/spectest/tests/happy_flow.go
olegshmuelov: noted. this design suggested by Alon.
@GalRogozinski pls approve if we want to adjust this logic

GalRogozinski: I think it is fine the way it is now.
The idea here is to have a seperation between the expected state and the test logic.

I think it is better for the dev to actively change the state as needed.

qbft/spectest/tests/msg_processing_spectest.go
GalRogozinski: opened this issue for running against `ssv` with no diffs :
https://github.com/bloxapp/ssv-spec/issues/213

From looking at the code I think the issue also existed before. Because I didn't that we serialize the `sc` files

qbft/spectest/tests/msg_processing_spectest.go
alonmuroch: @olegshmuelov @GalRogozinski this means that json tests will not be able to do state comparison

alonmuroch (approved):

PR #194, opened 2023-04-24
Fix Tests - 1

GalRogozinski:
Use signed message root instead of message root.
This will enforce signature checks.

We temporarily don't use FullData in the calculations. 
We should, but we don't want to break too many tests since this change is large enough.


qbft/spectest/all_tests.go
```diff
@@ -122,8 +122,8 @@ var AllTests = []tests.TestF{
 	proposal.InvalidFullData,
 	proposal.PastRoundProposalPrevPrepared,
 	proposal.NotPreparedPreviouslyJustification,
-	proposal.PreparedPreviouslyJustification,
-	proposal.DifferentJustifications,
+	//proposal.PreparedPreviouslyJustification,
```
y0sher, line 125: why comment tests?

GalRogozinski: The PR is split...
At each split we comment out or in some tests in order to get a ✅ .
When we combine all the splits in the end no tests should be commented out (this should be verified at the final review).

I just didn't know how to do it better 🤷

GalRogozinski:
closed because of priorities

PR #171, opened 2023-03-16
Versioned block support

olegshmuelov:
- Versioned block support for regular/blinded beacon block
- Deprecate `Bellatrix` ===> Introduce `Capella`
- Update tests with new roots
closes #140 

THIS PR split to 2 different PRs:
 - basic support for versioned block -  https://github.com/bloxapp/ssv-spec/pull/208
 - TBD. deprecates older forks & introduce Capella with state comparison.

PR #208, opened 2023-05-02
Version Block Support base

olegshmuelov:
split from https://github.com/bloxapp/ssv-spec/pull/171 as a base (basic support) for version block.


types/spectest/tests/consensusdata/proposer/versioned_block_spectest.go
```diff
@@ -0,0 +1,122 @@
...
+	ExpectedBlkRoot [32]byte
+	ExpectedCdRoot  [32]byte
+	ExpectedError   string
+}
+
+func (test *ProposerSpecTest) TestName() string {
+	return test.Name
+}
+
+// TODO: make multi proposer spec test for all (upcoming) versions
```
GalRogozinski, line 26: Before merging can you open a github issue for this and replace TODO with the issue number?

I don't believe in using TODOs for managing tasks.
I usually use them while I create a PR, and then at the end of the PR I clean them up.

types/spectest/tests/consensusdata/proposer/versioned_block_unknown_version.go
```diff
@@ -0,0 +1,34 @@
...
+	if err != nil {
+		panic(err.Error())
+	}
+
+	return &ProposerSpecTest{
+		Name:            "consensus data versioned block unknown version",
+		DataCd:          cdSSZ,
+		DataBlk:         nil,
+		ExpectedCdRoot:  [32]byte{},
+		ExpectedBlkRoot: [32]byte{},
```
GalRogozinski, line 31: ditto

types/testingutils/beacon_node.go
```diff
@@ -234,8 +236,8 @@ const (
 	UnknownDutyType = 100
 )
 
-var TestingSyncCommitteeBlockRoot = spec.Root{2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2}
-var TestingSyncCommitteeWrongBlockRoot = spec.Root{1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1}
+var TestingSyncCommitteeBlockRoot = phase0.Root{2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2}
```
GalRogozinski, line 239: Question:
I am guessing you don't have a preimage for this root ;-)
How does this work in tests?

I think for current tests you simply don't need the preimage.
For the future, maybe it is better to aways use you created from some value, to make the constant more useful.

No need to change anything now.

types/ssz.go
```diff
@@ -3,6 +3,7 @@ package types
 import (
 	"encoding/binary"
 	"fmt"
+
```
GalRogozinski, line 6: For the future, formatting only files should have their own PRs

ssv/value_check.go
```diff
@@ -96,11 +98,19 @@ func ProposerValueCheckF(
 			return errors.Wrap(err, "duty invalid")
 		}
 
-		if blockData, err := cd.GetBellatrixBlindedBlockData(); err == nil {
-			return signer.IsBeaconBlockSlashable(sharePublicKey, blockData.Slot)
+		if blockData, _, err := cd.GetBlindedBlockData(); err == nil {
```
y0sher, line 101: I would do something like this 

```go
type blockSubset interface { // if some other interface exist this isn't needed
   Slot() (slot, error)
   IsBeaconBlockSlashable() error
}
var blockData blockSubset
var err error
blockData, _, err = cd.GetBlindedBlockData()
if err != nil {
   blockData, err = cd.GetBlockData()
   if err != nil {
       return err // neither type of block works?
   }
}
slot, serr := blockData.Slot()
if serr != nil {
  return errors.Wrap(serr, "failed to get slot from block data")
}
return signer.IsBeaconBlockSlashable(sharePublicKey, slot)
}
```

this is just code-style and trying to make code more generic and less repeating yourself. its up to you if you prefer existing version just keep it.

types/spectest/tests/consensusdata/proposer/versioned_blinded_block_consensus_data_nil.go
```diff
@@ -0,0 +1,31 @@
...
+func VersionedBlindedBlockConsensusDataNil() *ProposerSpecTest {
+	cd := &types.ConsensusData{
+		Duty:    testingutils.TestingProposerDuty,
+		Version: spec.DataVersionBellatrix,
+		DataSSZ: nil,
+	}
+
+	cdSSZ, err := cd.MarshalSSZ()
+	if err != nil {
+		panic(err.Error())
```
y0sher, line 20: so basically this will always fail here?

y0sher (approved):
Approved but please look at my comments.

ssv/value_check.go
olegshmuelov: noted

types/spectest/tests/consensusdata/proposer/versioned_blinded_block_consensus_data_nil.go
olegshmuelov: no, marshaling ssz with nil bytes is possible

types/consensus_data.go
```diff
@@ -190,36 +183,54 @@ func (ci *ConsensusData) GetAttestationData() (*phase0.AttestationData, error) {
...
-	}
-	return ret, nil
-}
-
-func (ci *ConsensusData) GetCapellaBlockData() (*capella.BeaconBlock, error) {
-	ret := &capella.BeaconBlock{}
-	if err := ret.UnmarshalSSZ(ci.DataSSZ); err != nil {
-		return nil, errors.Wrap(err, "could not unmarshal ssz")
+func (ci *ConsensusData) GetBlockData() (*spec.VersionedBeaconBlock, ssz.HashRoot, error) {
+	switch ci.Version {
```
alonmuroch, line 187: why support these old versions? adds redundant tests and complexity

alonmuroch (changes requested):
A few others points

- all consensus data tests need to cover all different versions not just one
- SSV should include cappella block tests to test versioning actually. works

types/consensus_data.go
```diff
@@ -190,36 +183,54 @@ func (ci *ConsensusData) GetAttestationData() (*phase0.AttestationData, error) {
...
-		return nil, errors.Wrap(err, "could not unmarshal ssz")
-	}
-	return ret, nil
-}
-
-func (ci *ConsensusData) GetCapellaBlockData() (*capella.BeaconBlock, error) {
-	ret := &capella.BeaconBlock{}
-	if err := ret.UnmarshalSSZ(ci.DataSSZ); err != nil {
-		return nil, errors.Wrap(err, "could not unmarshal ssz")
+func (ci *ConsensusData) GetBlockData() (*spec.VersionedBeaconBlock, ssz.HashRoot, error) {
```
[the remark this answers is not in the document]
olegshmuelov: there are 2 cases we already discussed about:
1. [ProcessConsensus](https://github.com/olegshmuelov/ssv-spec/blob/cfb3893d0533f66cc1d4911b86ff9caa9cf61ba3/ssv/proposer.go#L136) - expect to receive the ssz.HashRoot for `signBeaconObject`
2. [ProcessPostConsensus](https://github.com/olegshmuelov/ssv-spec/blob/cfb3893d0533f66cc1d4911b86ff9caa9cf61ba3/ssv/proposer.go#L207) - expect to receive versioned block for `SubmitBeaconBlock`

 - if we want to get rid of the `HashRoot`, we have to switch/case the versioned block to get the relevant block and send it to the `signBeaconObject`
 - if we want to get rid of the `versionedBlock`, we have to pass the `HashRoot` and the `spec.DataVerion` to `SubmitBeaconBlock`

ssv/value_check.go
```diff
@@ -96,11 +98,19 @@ func ProposerValueCheckF(
...
+			slot, err := blockData.Slot()
+			if err != nil {
+				return errors.Wrap(err, "failed to get slot from blinded block data")
+			}
+			return signer.IsBeaconBlockSlashable(sharePublicKey, slot)
 		}
-		if blockData, err := cd.GetBellatrixBlockData(); err == nil {
-			return signer.IsBeaconBlockSlashable(sharePublicKey, blockData.Slot)
+		if blockData, _, err := cd.GetBlockData(); err == nil {
+			slot, err := blockData.Slot()
```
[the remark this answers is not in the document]
olegshmuelov: this is the [usage](https://github.com/attestantio/go-eth2-client/blob/6b2aee8911556e19ed0d349ace1caa08719eb27d/spec/versionedbeaconblock.go#L40) of the VersionedBeaconBlock to get the slot

olegshmuelov:
> A few others points
> 
> * all consensus data tests need to cover all different versions not just one
> * SSV should include cappella block tests to test versioning actually. works

the original PR https://github.com/bloxapp/ssv-spec/pull/171 decided to be split.
this is the base PR that not including Capella tests, but the upcoming split PRs (TBD) will include all the new Capella tests & Bellatrix with state comparison. 
@GalRogozinski

types/consensus_data.go
GalRogozinski: you wrote:
> all consensus data tests need to cover all different versions not just one

Anyhow, @lior-blox asked to split the old #171 PR up to chunks and do it correctly.

When we anticipate a new fork there must be a time when our code should support at least 2 versions so it will automatically adjust to the new version. Currently this is hard to do...

The final goal when all those PRs are merged is to be able to easily add a new fork support while supporting at least one old fork. In practice it will be able to support more than one old fork.

GalRogozinski: If you look at the way the `attestantio` lib implemented `VersionedBeaconBlock` you will see that Oleg's implementation is very much in line with the lib's intentions.

I personally think that the lib's implementation is bad... why do we need all the switch cases? I can think of a smarter less wasteful way utilizing interfaces that won't make us add a new field for every fork...
But given that this is the situation the current code is ok imo

However, since you asked and I agree that this is wasteful, I'll ask @olegshmuelov to have a default case for the old versions that will give out an error...
Otherwise, as new forks will come out this function will become even longer.

olegshmuelov: Removed all versions besides Bellatrix.
Upcoming split PR will include Capella support

types/spectest/tests/consensusdata/proposer/versioned_block_spectest.go
GalRogozinski:
```suggestion
ISSUE 219: make multi proposer spec test for all (upcoming) versions
```

GalRogozinski (approved):

types/consensus_data.go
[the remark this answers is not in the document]
alonmuroch: versioned block has a function called Root() which does the switching for you, it's kind of redundant to do it twice. If it's slowing us down let's open up an issue and clean this up later on.

types/consensus_data.go
alonmuroch: capella is still there.. if we add a new version (cappella) we need tests for it. If you want you can add it in another PR

GalRogozinski: The plan is definitely to add it in another PR.
We are currently in the process of breaking up the original PR up in order to meet the standard of "reviewable PRs"

types/consensus_data.go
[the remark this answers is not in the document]
olegshmuelov: sure. opened an issue https://github.com/bloxapp/ssv-spec/issues/221

types/consensus_data.go
GalRogozinski: @alonmuroch

you sure there is still Capella support here?
@olegshmuelov and I can't find it...

I think you may have an older version of the PR in your head before Oleg cleared it up

alonmuroch: I see it's gone, great.

alonmuroch (approved):

PR #210, opened 2023-05-02
State comparison - QBFT proposal

olegshmuelov:
Introduce state comparison for sanity QBFT proposal tests


GalRogozinski (approved):

y0sher (approved):

types/testingutils/comparable/instance.go
```diff
@@ -20,28 +17,46 @@ func SetMessages(instance *qbft.Instance, messages []*types.SSVMessage) {
 		if err := msg.Decode(ssvMsg.Data); err != nil {
 			panic(err.Error())
 		}
+		setMessage(instance, msg)
+	}
+}
+
+func SetSignedMessages(instance *qbft.Instance, messages []*qbft.SignedMessage) {
+	initContainers(instance)
```
y0sher, line 25: this is new, is it used anywhere?

types/testingutils/comparable/instance.go
```diff
@@ -20,28 +17,46 @@ func SetMessages(instance *qbft.Instance, messages []*types.SSVMessage) {
 		if err := msg.Decode(ssvMsg.Data); err != nil {
 			panic(err.Error())
 		}
+		setMessage(instance, msg)
+	}
+}
+
+func SetSignedMessages(instance *qbft.Instance, messages []*qbft.SignedMessage) {
```
alonmuroch, line 24: I rather not use this type of functions because it will become very hard to visually see the actual state comparison being made. This function could somehow fail and the test will pass

alonmuroch (changes requested):

types/testingutils/comparable/instance.go
GalRogozinski: The hard to see part I understand... I figured that what is important is the JSON (currently we don't print state comparisons but we will). I also thought it is in general a good practice to have shorter code so IO was fien with it.

> This function could somehow fail and the test will pass
Can you give an example? If the function doesn't do what it is supposed to do I believe the test will fail, no?

alonmuroch: It is but with state comparison we should be very visual to what the end result is. If we are not how do we know the state comparison is correct? Remember we do not have post root anymore

olegshmuelov: > I rather not use this type of functions because it will become very hard to visually see the actual state comparison being made. This function could somehow fail and the test will pass

I understand the concern about using this type of function, as it can make it difficult to visually observe the actual state comparison being performed.
Just to be clear that we were inspired by the example you implemented in [full_happy_flow_sc](https://github.com/bloxapp/ssv-spec/blob/83995dcf7aba81df6971b6473995c0d9f0e36ef0/ssv/spectest/tests/runner/full_happy_flow_sc.go#L24).

types/testingutils/comparable/instance.go
GalRogozinski: See https://github.com/bloxapp/ssv-spec/pull/210#discussion_r1196025399

alonmuroch (approved):

PR #212, opened 2023-05-02
State comparison - QBFT startinstance timeout

olegshmuelov:
Introduce state comparison for sanity QBFT startinstance timeout tests


qbft/spectest/tests/controller_spectest.go
```diff
@@ -25,6 +26,7 @@ type RunInstanceData struct {
 	InputValue           []byte
 	InputMessages        []*qbft.SignedMessage
 	ControllerPostRoot   string
+	ControllerPostState  types.Root `json:"-"` // Field is ignored by encoding/json
```
y0sher, line 29: so do we need both `ControllerPostRoot` and `ControllerPostState`?

y0sher: since it looks like we're using `ControllerPostState` as Root

olegshmuelov: its a good question
ideally we want to have the ControllerPostState only and generate the tests.json with this state (since we can get the root from this state), the issue is that we need to adjust the way ssv-node is working with those json files.

GalRogozinski: Currently we don't serialize the `PostState` json
When we change this, we can discuss the question again

As Oleg said, this is not for this PR

qbft/spectest/tests/startinstance/post_future_decided.go
```diff
@@ -21,12 +23,57 @@ func PreviousDecided() tests.SpecTest {
...
-				ControllerPostRoot: "5c23b58681df24d3131dab8fcff377035d03bb5a8db99c16c68bd644baa2d3b3",
+				InputValue:          []byte{1, 2, 3, 4},
+				ControllerPostRoot:  previousDecidedStateComparison(1, false).Root(),
+				ControllerPostState: previousDecidedStateComparison(1, false).ExpectedState,
 			},
 		},
 	}
 }
+
+func previousDecidedStateComparison(height qbft.Height, decidedState bool) *comparable.StateComparison {
```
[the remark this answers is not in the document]
alonmuroch: Just a simple test of starting a new instance after the previous one decided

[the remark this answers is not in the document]
GalRogozinski: opened #263

alonmuroch (approved):

GalRogozinski (approved):

Issue #214, opened 2023-05-03
A second triggering of a prepare quorum can cause a duplicate commit message to be omitted


GalRogozinski:
Since self-generated messages are only processed when gossiped back we can have a case where:
1. Prepare quorum is triggered and a commit is omitted.
2. Prepare quorum is retriggered and a duplicate commit is emitted because `didSendCommitForHeightAndRound()` wrongly returns false

Since network cache should mask duplicates from propagating in the network the implications are not serious. But it is unexpected behavior.

GalRogozinski:
@MatheusFranco99 
did you take this issue into account in MsgValidation?
I am now trying to fix this

PR #216, opened 2023-05-03
Issue C: Missing Tests Could Lead to Incorrect Implementations

GalRogozinski:
### Synopsis
The QBFT specification is tested by simple tests that run the consensus protocol with invalid data. We found some tests are missing, in particular there is no test for:
- a Prepare message that triggers quorum even though the commit stage is reached already; and
- a Commit message that is added to the message container but does not result in quorum.

### Impact
The tests aim to verify the correctness of the flows in the QBFT folder and in addition, the generated JSON file from the tests can be run in any implementation and hence should help developers testing their own implementation against the specification code. Missing tests can lead to wrong implementations by other developers.

### DIFF
--- a/qbft/spectest/tests/commit/no_commit_quorum.go
+++ b/qbft/spectest/tests/commit/no_commit_quorum.go
@@ -1,0 +1,34 @@
+package commit
+
+import (
+	"github.com/bloxapp/ssv-spec/qbft"
+	"github.com/bloxapp/ssv-spec/qbft/spectest/tests"
+	"github.com/bloxapp/ssv-spec/types/testingutils"
+)
+
+// NoCommitQuorum tests the state of the QBFT instance when received commit messages don't create a quorum
+func NoCommitQuorum() tests.SpecTest {
+	pre := testingutils.BaseInstance()
+	ks := testingutils.Testing4SharesSet()
+
+	msgs := []*qbft.SignedMessage{
+		testingutils.TestingProposalMessage(ks.Shares[1], 1),
+
+		testingutils.TestingPrepareMessage(ks.Shares[1], 1),
+		testingutils.TestingPrepareMessage(ks.Shares[2], 2),
+		testingutils.TestingPrepareMessage(ks.Shares[3], 3),
+
+		testingutils.TestingCommitMessage(ks.Shares[1], 1),
+		testingutils.TestingCommitMessage(ks.Shares[2], 2),
+	}
+	return &tests.MsgProcessingSpecTest{
+		Name:          "no commit quorum",
+		Pre:           pre,
+		PostRoot:      "0a940d3ffeb4d28f3ef9de1f119aeade7ad6ec9b998548b24bd00b79739e3e0f",
+		InputMessages: msgs,
+		OutputMessages: []*qbft.SignedMessage{
+			testingutils.TestingPrepareMessage(ks.Shares[1], 1),
+			testingutils.TestingCommitMessage(ks.Shares[1], 1),
+		},
+	}
+}
--- a/qbft/spectest/tests/prepare/all_prepares_sent.go
+++ b/qbft/spectest/tests/prepare/all_prepares_sent.go
@@ -1,0 +1,34 @@
+package prepare
+
+import (
+	"github.com/bloxapp/ssv-spec/qbft"
+	"github.com/bloxapp/ssv-spec/qbft/spectest/tests"
+	"github.com/bloxapp/ssv-spec/types/testingutils"
+)
+
+// AllPreparesSent is a spec test that checks the case where all prepares are sent and quorum event is triggered more than once.
+func AllPreparesSent() tests.SpecTest {
+	ks := testingutils.Testing4SharesSet()
+	pre := testingutils.BaseInstance()
+	msgs := []*qbft.SignedMessage{
+		testingutils.TestingProposalMessage(ks.Shares[1], 1),
+
+		testingutils.TestingPrepareMessage(ks.Shares[1], 1),
+		testingutils.TestingPrepareMessage(ks.Shares[2], 2),
+		testingutils.TestingPrepareMessage(ks.Shares[3], 3),
+
+		testingutils.TestingCommitMessage(ks.Shares[1], 1),
+
+		testingutils.TestingPrepareMessage(ks.Shares[4], 4),
+	}
+	return &tests.MsgProcessingSpecTest{
+		Name:          "all prepares sent",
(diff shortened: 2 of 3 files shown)


GalRogozinski:
SC will be added upon merger of #191

qbft/spectest/tests/commit/no_commit_quorum.go
```diff
@@ -0,0 +1,71 @@
...
+
+func NoCommitQuorumStateComparison() *qbftcomparable.StateComparison {
+	ks := testingutils.Testing4SharesSet()
+
+	state := testingutils.BaseInstance().State
+	state.ProposalAcceptedForCurrentRound = testingutils.TestingProposalMessage(ks.Shares[1], types.OperatorID(1))
+
+	state.LastPreparedRound = 1
+	state.LastPreparedValue = testingutils.TestingQBFTFullData
+
```
alonmuroch, line 48: specify here, even though by default, that decided=false

qbft/spectest/tests/prepare/prepare_quorum_triggered_twice.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"github.com/bloxapp/ssv-spec/qbft"
+	qbftcomparable "github.com/bloxapp/ssv-spec/qbft/spectest/comparable"
+	"github.com/bloxapp/ssv-spec/qbft/spectest/tests"
+	"github.com/bloxapp/ssv-spec/types"
+	"github.com/bloxapp/ssv-spec/types/testingutils"
+)
+
+// PrepareQuorumTriggeredTwice is a spec test that checks the case where all prepares are sent and quorum event is triggered more than once.
```
alonmuroch, line 11: PrepareQuorumTriggeredTwice tests triggering prepare quorum twice by sending > 2f+1 prepare messages

qbft/spectest/tests/prepare/prepare_quorum_triggered_twice.go
```diff
@@ -0,0 +1,71 @@
...
+func PrepareQuorumTriggeredTwice() tests.SpecTest {
```
alonmuroch, line 12: what is the effect tested here? did we find a bug? commit message sent twice?

qbft/spectest/tests/prepare/prepare_quorum_triggered_twice_late_commit.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"github.com/bloxapp/ssv-spec/qbft"
+	"github.com/bloxapp/ssv-spec/qbft/spectest/tests"
+	"github.com/bloxapp/ssv-spec/types/testingutils"
+)
+
+// PrepareQuorumTriggeredTwiceLateCommit is a spec test that checks the case where all prepares are sent and quorum event is triggered more than once.
+// A commit message was seen only after the last prepare
+func PrepareQuorumTriggeredTwiceLateCommit() tests.SpecTest {
```
alonmuroch, line 11: why is this a special case?

alonmuroch (changes requested):

qbft/spectest/tests/commit/no_commit_quorum.go
GalRogozinski: done

qbft/spectest/tests/prepare/prepare_quorum_triggered_twice.go
GalRogozinski: The issue was that the audit team thought that core tests were missing.
They didn't specify a bug and on this case we didn't find a bug.

However, look at the net comment for a bug that I found :-)

qbft/spectest/tests/prepare/prepare_quorum_triggered_twice_late_commit.go
GalRogozinski: See #214 for a bug on this case.
We send two commit messages instead of one.
After consulting @lior-blox and @y0sher it was  decided it is low priority for now.
This is because networking cache takes care of this issue, and even if it didn't it is unclear how it can be a problem (if we disregard message validation).

qbft/spectest/tests/prepare/prepare_quorum_triggered_twice.go
GalRogozinski: done

qbft/spectest/tests/commit/no_commit_quorum.go
alonmuroch: where?

qbft/spectest/tests/prepare/prepare_quorum_triggered_twice.go
alonmuroch: mm u didn't change the comment

alonmuroch: you didn't change the comment

qbft/spectest/tests/prepare/prepare_quorum_triggered_twice.go
alonmuroch: so this is purely audit feedback and fixes?

alonmuroch (changes requested):

qbft/spectest/tests/commit/no_commit_quorum.go
GalRogozinski: https://github.com/bloxapp/ssv-spec/pull/216/files#diff-0e78fc58e6c29d591e645957f5f351fb7b63c695b917bdc4026d2fe8f8b34d6fR48

qbft/spectest/tests/prepare/prepare_quorum_triggered_twice_late_commit.go
y0sher: @GalRogozinski it is to be discussed what do we want to do when we get a message of type that we already get a quorum of. 
This acts the same on prepare and also on commits, I was sure this is a design decision. but maybe we can stop processing after quorum? or at least not broadcast another message?

y0sher (approved):

qbft/spectest/tests/prepare/prepare_quorum_triggered_twice_late_commit.go
GalRogozinski: @y0sher Maybe,
The problem is that you don't know if you broadcasted the first commit message after you receive it from gossip

Issue #221, opened 2023-05-08
GetBlockData/GetBlindedBlockData return versioned block only


olegshmuelov:
ConsensusData GetBlockData should return versioned block only
```
// TODO func (ci *ConsensusData) GetBlockData() (*spec.VersionedBeaconBlock, error) {
func (ci *ConsensusData) GetBlockData() (*spec.VersionedBeaconBlock, ssz.HashRoot, error) {
	switch ci.Version {
	case spec.DataVersionBellatrix:
		ret := &bellatrix.BeaconBlock{}
		if err := ret.UnmarshalSSZ(ci.DataSSZ); err != nil {
			return nil, nil, errors.Wrap(err, "could not unmarshal ssz")
		}
		return &spec.VersionedBeaconBlock{Bellatrix: ret, Version: spec.DataVersionBellatrix}, ret, nil
	default:
		return nil, nil, errors.Errorf("unknown block version %s", ci.Version.String())
	}
}
```

PR #222, opened 2023-05-08
Return versioned block only

olegshmuelov:
closes https://github.com/bloxapp/ssv-spec/issues/221


ssv/proposer.go
```diff
@@ -232,19 +240,29 @@ func (r *ProposerRunner) expectedPreConsensusRootsAndDomain() ([]ssz.HashRoot, p
...
+			return nil, phase0.DomainType{}, errors.Wrap(err, "could not get block root")
+		}
 	}
 
-	_, data, err := r.GetState().DecidedValue.GetBlockData()
-	if err != nil {
-		return nil, phase0.DomainType{}, errors.Wrap(err, "could not get block data")
-	}
-	return []ssz.HashRoot{data}, types.DomainProposer, nil
+	hashRoot := types.SSZ32Bytes(root)
```
alonmuroch, line 264: @GalRogozinski notice you returned the root instead of object here right?

GalRogozinski: you mean return `vBlk` instead of `[]ssz.HashRoot{hashRoot}`. That's even more refactoring considering that the current function name is `expectedPostConsensusRootsAndDomain()`.

I wouldn't spend farther time refactoring this code. Given that the current way isn't bad in my view and there are more pressing issues.

y0sher (approved):
lgtm

GalRogozinski (approved):

GalRogozinski:
Closing this since it is not in priority.
May be reopened later

PR #177, opened 2023-03-29
MEV support 

nkryuchkov:
Supersedes https://github.com/bloxapp/ssv-spec/pull/163
Includes https://github.com/bloxapp/ssv-spec/pull/171 as ssv depends on it

PR #163, opened 2023-03-01
Mev support

nivBlox:

PR #226, opened 2023-05-09
MEV support

nkryuchkov:
Supersedes https://github.com/bloxapp/ssv-spec/pull/177, https://github.com/bloxapp/ssv-spec/pull/163

Spec part of https://github.com/bloxapp/ssv/pull/918


nkryuchkov:
Hey @GalRogozinski

1. I'll check #210 
2. I think I could extract around 5-10 line changes that are not directly related to MEV, but the rest looks to me like something whole. Why do you think the PR is not small enough? It's 109 changed lines now

ssv/spectest/run_test.go
```diff
@@ -229,6 +229,10 @@ func fixRunnerForRun(t *testing.T, runnerMap map[string]interface{}, ks *testing
 		ret.(*ssv.ProposerRunner).ProducesBlindedBlocks = blindedBlocks.(bool)
 	}
 
+	if gasLimit, ok := runnerMap["GasLimit"]; ok {
+		ret.(*ssv.ValidatorRegistrationRunner).GasLimit = uint64(gasLimit.(float64))
```
y0sher, line 233: ret will always be `ssv.ValidatorRegistrationRunner` when `runnerMap["GasLimit"]` is set?

ssv/value_check.go
```diff
@@ -99,6 +101,9 @@ func ProposerValueCheckF(
 		}
 
 		if blockData, _, err := cd.GetBlindedBlockData(); err == nil {
```
y0sher, line 103: isn't it better to run this code only if `supportsBlinded` is true?

types/beacon_types.go
```diff
@@ -167,7 +169,7 @@ func (n BeaconNetwork) EstimatedSlotAtTime(time int64) spec.Slot {
 }
 
 func (n BeaconNetwork) EstimatedTimeAtSlot(slot spec.Slot) int64 {
-	d := int64(slot) * int64(n.SlotsPerEpoch())
+	d := int64(slot) * int64(n.SlotDurationSec().Seconds())
```
y0sher, line 172: logic seems different, was this function used anywhere else? does is still work for the other places if true?

y0sher (approved):
Approved. some minor comments.

ssv/spectest/run_test.go
nkryuchkov: Yes, it's done the same way as the code above. If it's not there, it means that's JSON is from a previous version

ssv/value_check.go
nkryuchkov: I think no, because if blinded is off and a blinded block is passed, it'd be good to return a special error (the one below) instead of something like "failed to decode ..."

types/beacon_types.go
nkryuchkov: it's used only in `calculateValidatorRegistration` to calculate a timestamp

ssv/types.go
```diff
@@ -35,10 +36,12 @@ type AttesterCalls interface {
 
 // ProposerCalls interface has all block proposer duty specific calls
 type ProposerCalls interface {
+	// SubmitValidatorRegistration submits a validator registration
+	SubmitValidatorRegistration(pubkey []byte, feeRecipient bellatrix.ExecutionAddress, sig phase0.BLSSignature) error
 	// GetBeaconBlock returns beacon block by the given slot and committee index
-	GetBeaconBlock(slot phase0.Slot, committeeIndex phase0.CommitteeIndex, graffiti, randao []byte) (ssz.Marshaler, spec.DataVersion, error)
```
GalRogozinski, line 39: what is the reason for deletion?

ssv/validator_registration.go
```diff
@@ -60,6 +61,20 @@ func (r *ValidatorRegistrationRunner) ProcessPreConsensus(signedMsg *types.Signe
 		return nil
 	}
 
+	// only 1 root, verified in basePreConsensusMsgProcessing
+	root := roots[0]
+	// randao is relevant only for block proposals, no need to check type
```
GalRogozinski, line 66: you are correct about Randao, but what is the point of putting this comment here?
I am missing some context?

ssv/validator_registration.go
```diff
@@ -60,6 +61,20 @@ func (r *ValidatorRegistrationRunner) ProcessPreConsensus(signedMsg *types.Signe
 		return nil
 	}
 
+	// only 1 root, verified in basePreConsensusMsgProcessing
```
GalRogozinski, line 64: why?
Was there an explanation in the SIP?

types/beacon_types.go
GalRogozinski: so this is a bug fix, great!

types/beacon_types.go
```diff
@@ -58,6 +58,8 @@ func (r BeaconRole) String() string {
 		return "SYNC_COMMITTEE"
 	case BNRoleSyncCommitteeContribution:
 		return "SYNC_COMMITTEE_CONTRIBUTION"
+	case BNRoleValidatorRegistration:
+		return "VALIDATOR_REGISTRATION"
```
GalRogozinski, line 62: Great!

types/testingutils/beacon_node.go
```diff
@@ -497,8 +497,12 @@ func (bn *TestingBeaconNode) SubmitAttestation(attestation *phase0.Attestation)
 	return nil
 }
 
+func (bn *TestingBeaconNode) SubmitValidatorRegistration(pubkey []byte, feeRecipient bellatrix.ExecutionAddress, sig phase0.BLSSignature) error {
+	return nil
+}
+
```
GalRogozinski, line 503: Shouldn't we have a similar treatment that we do for attestations for tests? 
See `BroadcastedRoots` above

GalRogozinski (changes requested):
Initial review

GalRogozinski (changes requested):
Initial review

ssv/types.go
nkryuchkov: According to beacon node spec https://ethereum.github.io/beacon-APIs/#/Validator/produceBlockV2, we don't need committee index, we need only slot, graffiti, and randao

ssv/validator_registration.go
nkryuchkov: Because `PartialSignatureMessages` contains only one message for this runner according to https://github.com/nkryuchkov/ssv-spec/blob/mev-support-2/ssv/validator_registration.go#L116. This particular check is copied from the aggregator's and proposer's `ProcessPreConsensus`

types/beacon_types.go
nkryuchkov: thanks!

ssv/validator_registration.go
nkryuchkov: AFAIK there is no explanation in the SIP but I'm not sure if we need it because I think it's rather implementation-specific. 

@moshe-blox As you wrote the SIP, what's your opinion on this?

ssv/types.go
GalRogozinski: ok so there was a difference between our spec and impl?
because it is hard for me to beleieve that the impl had the same mistake

Can you also fix the comment?

types/testingutils/beacon_node.go
nkryuchkov: done

ssv/validator_registration.go
moshe-blox: @GalRogozinski @nkryuchkov Gal spoke with me, I think we only take 1 root because we've signed only 1 message, is that right?

ssv/validator_registration.go
```diff
@@ -60,6 +61,19 @@ func (r *ValidatorRegistrationRunner) ProcessPreConsensus(signedMsg *types.Signe
...
+	root := roots[0]
+	fullSig, err := r.GetState().ReconstructBeaconSig(r.GetState().PreConsensusContainer, root, r.GetShare().ValidatorPubKey)
+	if err != nil {
+		return errors.Wrap(err, "could not reconstruct randao sig")
```
GalRogozinski, line 68:
```suggestion
		return errors.Wrap(err, "could not reconstruct validator registration sig")
```

ssv/validator_registration.go
```diff
@@ -13,6 +13,7 @@ import (
 
 type ValidatorRegistrationRunner struct {
 	BaseRunner *BaseRunner
+	GasLimit   uint64
```
GalRogozinski, line 16: A design question..
Why isn't `FeeRecipientAddress` here?
Why is it part of `Share`?

Also imo `NewValidatorRegistrationRunner` should initialize both `GasLimit` and `FeeRecipientAddress`. If `FeeRecipientAddress` isn't fed a value an error should occur. If `GasLimit` limit isn't fed a value we should use the default 30 mil.

GalRogozinski: Also please adjust `testingutils.TestingValidatorRegistration` to use the correct GasLimit and make sure tests pass

nkryuchkov: > Why is it part of `Share`?

Because it's set in `FeeRecipientAddressUpdatedEvent`, but we don't allow configuring gas limit in contracts, so `GasLimit` is not a part of `Share`.

> or maybe never feed `GasLimit` a value.

Do you mean we should probably just use a 30M constant an remove `GasLimit` from `ValidatorRegistrationRunner`?

GalRogozinski: Regarding `FeeRecipientAddress`:
I am guessing (correct me if I am wrong) that this is a per validator setting.
So why it should be in Share?

Also the comment says:
> // Share holds all info about the QBFT/ SSV Committee for msg signing and verification

So it looks not in place

GalRogozinski: Regarding `GasLimit`

see https://github.com/bloxapp/SIPs/pull/25/files#r1194886554

I think constant is good enough!
Also it will remove all the changed roots hopefully making this PR safer to merge :-)

GalRogozinski:
@nkryuchkov 
I think that for all tests in the `preconsensus` folder that are for "validator registration" we should add 
`	BeaconBroadcastedRoots: []string{
					testingutils.GetSSZRootNoError(testingutils.TestingValidatorRegistration),
				}`

ssv/validator_registration.go
nkryuchkov: > Also it will remove all the changed roots

Unfortunately not 🙁 They were calculated with `GasPrice == 1`. And the implementation runs spec tests on itself to verify itself. So,
- if we leave `GasPrice == 1` in spec, the implementation won't pass spec tests because we cannot use `1` there
- if we change `GasPrice` to `30_000_000` in spec, the roots will be affected, and also if we need to use another value in the future, we'll need to change roots again
- if we make `GasPrice` configurable and use `GasPrice == 1` in spec and `30_000_000` in the implementation (the way it's done in this PR), the roots will be also affected

nkryuchkov: And anyway, when implementing State Comparison, we'll need to touch roots

GalRogozinski: It is correct to have it as a constant regardless imo.. but if you disagree you can comment on the SIP

I think it is best to have the last commit change the value of the constant and the roots of the tests.
This way we will have a proof to what caused the roots to change.

nkryuchkov: > I am guessing (correct me if I am wrong) that this is a per validator setting.

Yes.

> So why it should be in Share?

Because `Share` describes a validator being a part of a committee. And `FeeRecipientAddress` is one of its properties. Is your concern about the same values in some fields for the same validator for different operators?

> So it looks not in place

Do you mean that the `for msg signing and verification` part does not provide a full description of what `Share` does or anything else?

nkryuchkov: > It is correct to have it as a constant regardless imo.. but if you disagree you can comment on the SIP

I don't mind having it as a constant, I'll do that

GalRogozinski: > Because Share describes a validator being a part of a committee.
AFAIU:
Share describes an operator ~~validator~~ being a part of a committee, and what the operator needs for signing and validation.

`FeeRecipientAddress` is a validator property not an operator.
`Grafitti` for example is an operator property because it can be different for each operator.
`ValidatorPubKey` - is a validator property but at least the operator has to know about it for signing and verification.
`Commitee`, `Quorum`, and `DomainType` - ssv data that makes sense it is in share.

But I can't understand why `FeeRecipientAddress` is there.
But before we do anything let's see what @y0sher thinks.

nkryuchkov: > Share describes a validator being a part of a committee.

I meant that `Share` describes a validator being a part of a committee of operators (in other words, validators' share in the committee of operators). Operators have shares for each of their validators. So it's kind of a descriptor of validator that is used by operator. And I think `FeeRecipientAddress` should be there because it's a validator's property used by operator.

> But before we do anything let's see what @y0sher thinks.

Alright, also if @moshe-blox has any comments to this topic, that would be helpful

types/testingutils/runner.go
```diff
@@ -18,13 +18,13 @@ var AttesterRunner7Operators = func(keySet *TestKeySet) ssv.Runner {
...
 var ProposerRunner = func(keySet *TestKeySet) ssv.Runner {
-	return baseRunner(types.BNRoleProposer, ssv.ProposerValueCheckF(NewTestingKeyManager(), types.BeaconTestNetwork, TestingValidatorPubKey[:], TestingValidatorIndex, nil), keySet)
+	return baseRunner(types.BNRoleProposer, ssv.ProposerValueCheckF(NewTestingKeyManager(), types.BeaconTestNetwork, TestingValidatorPubKey[:], TestingValidatorIndex, nil, true), keySet)
 }
 
 var ProposerBlindedBlockRunner = func(keySet *TestKeySet) ssv.Runner {
 	ret := baseRunner(
 		types.BNRoleProposer,
-		ssv.ProposerValueCheckF(NewTestingKeyManager(), types.BeaconTestNetwork, TestingValidatorPubKey[:], TestingValidatorIndex, nil),
+		ssv.ProposerValueCheckF(NewTestingKeyManager(), types.BeaconTestNetwork, TestingValidatorPubKey[:], TestingValidatorIndex, nil, true),
```
GalRogozinski, line 27: It is confusing that we have a blinded block value checker that works on the consensus and a blinded block runner that works on preconsesnsus...

Since those are 2 different things I guess it is fine...

ssv/types.go
```diff
@@ -35,10 +36,12 @@ type AttesterCalls interface {
 
 // ProposerCalls interface has all block proposer duty specific calls
 type ProposerCalls interface {
-	// GetBeaconBlock returns beacon block by the given slot and committee index
-	GetBeaconBlock(slot phase0.Slot, committeeIndex phase0.CommitteeIndex, graffiti, randao []byte) (ssz.Marshaler, spec.DataVersion, error)
-	// GetBlindedBeaconBlock returns blinded beacon block by the given slot and committee index
-	GetBlindedBeaconBlock(slot phase0.Slot, committeeIndex phase0.CommitteeIndex, graffiti, randao []byte) (ssz.Marshaler, spec.DataVersion, error)
+	// SubmitValidatorRegistration submits a validator registration
+	SubmitValidatorRegistration(pubkey []byte, feeRecipient bellatrix.ExecutionAddress, sig phase0.BLSSignature) error
```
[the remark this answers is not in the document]
nkryuchkov: Because registering validators is needed for proposals using MEV. But since we have a separate duty for validator registrations, I agree that we need to extract this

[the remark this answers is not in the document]
nkryuchkov: Fixed

types/testingutils/runner.go
nkryuchkov: IIRC in pre-consensus phase they sign randao and once pre-consensus is reached, they get a block from node and run a consensus with the block data as consensus input, and that's where we need to handle a case if blinded support is off but a blinded block was received

ssv/validator_registration.go
moshe-blox: Fee recipient in shares is inherited from validators which inherit from their owner.
```
- Owner <preferred fee recipient>
  - Validator
    - Share
    - Share
    - Share
    - Share
  - Validator
    - Share
    - Share
    - Share
    - Share
```
I agree that since it cannot change on a per-validator or per-share basis, it'd be ideal if it wouldn't be a property in shares to avoid confusion.

However, changing this would be out of scope for this PR, I think.

@alonmuroch or @lior-blox wdyt?

GalRogozinski:
@alonmuroch 
If we change `DefaultGasLimit` to 1, the roots stay the same.

ssv/validator_registration.go
GalRogozinski: Agree it is out of scope of this PR.
Opened #242

GalRogozinski (approved):

ssv/spectest/tests/runner/full_happy_flow.go
```diff
@@ -173,11 +164,14 @@ func FullHappyFlow() tests.SpecTest {
 					testingutils.SSVMsgValidatorRegistration(nil, testingutils.PreConsensusValidatorRegistrationMsg(ks.Shares[2], 2)),
 					testingutils.SSVMsgValidatorRegistration(nil, testingutils.PreConsensusValidatorRegistrationMsg(ks.Shares[3], 3)),
 				},
-				PostDutyRunnerStateRoot: fullHappyFlowValidatorRegistrationSC().Root(), // "f36c8b537afaba0894dbc8c87cb94466d8ac2623e9283f1c584e3d544b5f2b88",
+				PostDutyRunnerStateRoot: fullHappyFlowValidatorRegistrationSC().Root(), // "ec573732e70b70808972c43acb5ead6443cff06ba30d8abb51e37ac82ffe0727",
```
alonmuroch, line 167: tested to see the post state is what we expect?

alonmuroch (changes requested):

ssv/spectest/tests/runner/consensus/valid_message.go
```diff
@@ -132,10 +132,13 @@ func ValidMessage() tests.SpecTest {
 							testingutils.TestAttesterConsensusDataByts,
 						), nil),
 				},
-				PostDutyRunnerStateRoot: "f36c8b537afaba0894dbc8c87cb94466d8ac2623e9283f1c584e3d544b5f2b88",
+				PostDutyRunnerStateRoot: "ec573732e70b70808972c43acb5ead6443cff06ba30d8abb51e37ac82ffe0727",
```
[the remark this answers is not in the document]
GalRogozinski: https://github.com/bloxapp/ssv-spec/pull/226#issuecomment-1551094855

ssv/value_check.go
```diff
@@ -84,6 +85,7 @@ func ProposerValueCheckF(
 	validatorPK types.ValidatorPK,
 	validatorIndex phase0.ValidatorIndex,
 	sharePublicKey []byte,
+	supportsBlinded bool,
```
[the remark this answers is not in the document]
nkryuchkov: @alonmuroch We need this for a case if a non-round-leader operator has disabled MEV support but the round leader has enabled MEV support and broadcasts a blinded block payload. With this, the one with MEV off would be able to check such case and return a human-readable error (`blinded blocks are not supported`)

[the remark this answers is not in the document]
GalRogozinski: According to the current discussion on the SIP we want this configuration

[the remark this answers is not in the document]
moshe-blox: Reasoning provided in this comment: https://github.com/bloxapp/SIPs/pull/25#discussion_r1196242791

ssv/spectest/tests/runner/postconsensus/invalid_msg_slot.go
```diff
@@ -138,10 +138,13 @@ func InvalidMessageSlot() tests.SpecTest {
 					testingutils.SSVMsgValidatorRegistration(nil, testingutils.PreConsensusValidatorRegistrationMsg(ks.Shares[3], 3)),
 					testingutils.SSVMsgValidatorRegistration(nil, invalidateSlot(testingutils.PostConsensusAttestationMsg(ks.Shares[1], 1, qbft.FirstHeight))),
 				},
-				PostDutyRunnerStateRoot: "f36c8b537afaba0894dbc8c87cb94466d8ac2623e9283f1c584e3d544b5f2b88",
```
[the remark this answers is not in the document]
GalRogozinski: confirmed

alonmuroch:
> @alonmuroch
> Could you shed light on the purpose of `preconsensus/unordered_expected_roots` tests? It looks the same as `valid_msg`?

just checking roots not ordered, should pass as a valid msg

alonmuroch (approved):

moshe-blox (approved):

PR #234, opened 2023-05-11
Version Block Support - Capella

olegshmuelov:
### Description
This pull request adds support for Capella beacon block


types/spectest/tests/ssz/ssz_spectest.go
```diff
@@ -0,0 +1,41 @@
...
+
+func (test *SSZSpecTest) Run(t *testing.T) {
+	cd := &types.ConsensusData{}
+	require.NoError(t, cd.Decode(test.Data))
+
+	vBlk, _, err := cd.GetBlockData()
+	require.NoError(t, err)
+	require.NotNil(t, vBlk)
+
+	var withdrawals types.SSZWithdrawals
```
GalRogozinski, line 31: As discussed with Oleg, currently SSZSpecTest is only good for withdrawals but later we can come up with better infra

types/testingutils/beacon_node_versioned.go
```diff
@@ -174,8 +242,68 @@ var TestingInvalidDutySlotV = func(version spec.DataVersion) phase0.Slot {
...
 
 	default:
 		panic("unsupported version")
 	}
 }
+
+// taken from https://github.com/attestantio/go-eth2-client/blob/1a94a0a534b9d75087403954c33f2a5819a896a9/spec/capella/beaconblock_test.go#L133
+// the block adjusted with the following changes:
+// - body.deposits proofs: 0x000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20 -> 0x000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f (they have wrong size) https://github.com/attestantio/go-eth2-client/issues/54
+// - slot: 1 -> TestingDutySlotCapella ----> to keep the consistency with TestingProposerDutyV Capella slot
```
GalRogozinski, line 256: `TestingDutyCapella` is not 1...

types/testingutils/beacon_node_versioned.go
```diff
@@ -174,8 +242,68 @@ var TestingInvalidDutySlotV = func(version spec.DataVersion) phase0.Slot {
...
...406307396dc7ece7938225fc5a55c65ed3608c27f668ba9b08875979e249655d15a4c3d77b1433ab26f56d80e2ca9178d88501ee682a14d07b17635e9968da2d98aa9eb4a5ae639e42eed0c7735313d3308fd079589d5cb2dd381a9298f67325976a6dcd3e38f836e4bc3106885f6348c57c90beadd7cbca17570b90130aaadc7eb0ed2bfb97c5c8a791d0b6b2736d43b0f444909449e6ea32ba706ddfe28c407a85e82e3f22064696d2fd7b37b33e01e0016ad91047f95f0cf5c5d6abc8fd7698c4155c290a1a1db842e77e656c69ad8bc06cea9e00dadbec886ca64af5fe3045ebc8c549bf23e71c634f02da8b250618c169a54e4af45d799f2de6747bde01395a13f1a605e9be5","0x9ced04a51e77ed0730f3420571164ea5247a670b962ebf6b453660748ca5"],"withdrawals":[{"index":"2","validator_index":"3","address":"0x000102030405060708090a0b0c0d0e0f10111213","amount":"1000000000000000000"}]},"bls_to_execution_changes":[{"message":{"validator_index":"2","from_bls_pubkey":"0xb89bebc699769726a318c8e9971bd3171297c61aea4a6578a7a4f94b547dcba5bac16a89108b6b6a1fe3695d1a874a0b","to_execution_address":"0x000102030405060708090a0b0c0d0e0f10111213"},"signature":"0x000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f202122232425262728292a2b2c2d2e2f303132333435363738393a3b3c3d3e3f404142434445464748494a4b4c4d4e4f505152535455565758595a5b5c5d5e5f"}]}}`)
```
GalRogozinski, line 257: This is currently a huge one liner.
I would create a new file to keep block constants and place it there in a formatted way.
Maybe also give it a better name than `tmp`?

types/ssz.go
```diff
@@ -82,6 +83,34 @@ func (b SSZTransactions) HashTreeRoot() ([32]byte, error) {
...
 
+// SSZWithdrawals --
+type SSZWithdrawals []*capella.Withdrawal
+
+func (b SSZWithdrawals) GetTree() (*ssz.Node, error) {
+	return ssz.ProofTree(b)
+}
+
+func (b SSZWithdrawals) HashTreeRootWith(hh ssz.HashWalker) error {
+	// taken from https://github.com/attestantio/go-eth2-client/blob/bc14358487b6d32cb45feef14b170458abc5d14a/spec/capella/executionpayload_ssz.go#L332-L346
```
GalRogozinski, line 94: Why can't this be used directly from the lib?

types/spectest/tests/ssz/withdrawals_marshaling_validation.go
```diff
@@ -0,0 +1,24 @@
...
+
+	root, err := withdrawals.HashTreeRoot()
+	if err != nil {
+		panic(err.Error())
+	}
+
+	return &SSZSpecTest{
+		Name:         "ssz withdrawals marshalling",
+		Data:         testingutils.TestProposerConsensusDataBytsV(spec.DataVersionCapella),
+		ExpectedRoot: root,
```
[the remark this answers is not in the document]
olegshmuelov: do we want to add it in the current PR or in Capella tests?

types/testingutils/beacon_node_versioned.go
```diff
@@ -93,6 +127,16 @@ var TestingWrongBeaconBlockV = func(version spec.DataVersion) *spec.VersionedBea
 			Version:   version,
 			Bellatrix: ret,
 		}
+	case spec.DataVersionCapella:
+		ret := &capella.BeaconBlock{}
+		if err := ret.UnmarshalSSZ(blkByts); err != nil {
+			panic(err.Error())
+		}
+		ret.Slot = 100
```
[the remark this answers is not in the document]
olegshmuelov: lets use `TestingDutySlotCapella + 100` to keep consistency with un-versioned funcs

types/testingutils/beacon_node_versioned.go
olegshmuelov: what we mean here is instead slot 1 use TestingDutySlotCapella

types/testingutils/beacon_node_versioned.go
```diff
@@ -1,22 +1,33 @@
...
 	"github.com/attestantio/go-eth2-client/spec/phase0"
 	ssz "github.com/ferranbt/fastssz"
 
 	"github.com/bloxapp/ssv-spec/types"
 )
 
 const (
-	// ForkSlotCapella taken from https://github.com/ethereum/consensus-specs/blob/1c424d76eddbacae3cbffed8276264b46951456b/specs/capella/fork.md?plain=1#L30
-	ForkSlotCapella = 6209536 // Epoch(194048)
+	// ForkEpochCapella Goerli taken from https://github.com/ethereum/execution-specs/blob/37a8f892341eb000e56e962a051a87e05a2e4443/network-upgrades/mainnet-upgrades/shanghai.md?plain=1#L18
```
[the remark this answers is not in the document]
olegshmuelov: Goerli is an execution layer testnet

types/ssz.go
olegshmuelov: There is no independent implementation for the `Withdrawals HashTreeRootWith` but for the wrapper struct `ExecutionPayload`
see the link attached as a comment

ssv/spectest/tests/runner/full_happy_flow_sc.go
```diff
@@ -23,7 +23,7 @@ func fullHappyFlowSyncCommitteeContributionSC() *comparable.StateComparison {
 				StartingDuty: &testingutils.TestSyncCommitteeContributionConsensusData.Duty,
 				PreConsensusContainer: ssvcomparable.SetMessagesInContainer(
 					ssv.NewPartialSigContainer(3),
-					testingutils.SSVDecidingMsgsV(testingutils.TestSyncCommitteeContributionConsensusData, ks, types.BNRoleSyncCommitteeContribution)[:3]),
+					testingutils.ExpectedSSVDecidingMsgsV(testingutils.TestSyncCommitteeContributionConsensusData, ks, types.BNRoleSyncCommitteeContribution)[:3]),
```
y0sher, line 26: V function shouldn't recv version?

y0sher: same q for all below changes in the same file.

types/spectest/tests/ssz/ssz_spectest.go
y0sher: would love a brief explanation about this whole folder `ssz`, so we're testing that we're able to serialize and de-serialize our objects using SSZ?

types/spectest/tests/ssz/withdrawals_marshaling_validation.go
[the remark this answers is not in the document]
y0sher: I know we're trying to keep small PRs but adding Capella but not adding Capella tests on the same PR kinda misses the point for me.

types/testingutils/beacon_node.go
```diff
@@ -515,30 +467,21 @@ func (bn *TestingBeaconNode) SubmitBeaconBlock(block *spec.VersionedBeaconBlock,
...
-		r, _ = sb.HashTreeRoot()
-	case spec.DataVersionAltair:
-		if block.Altair == nil {
+	case spec.DataVersionBellatrix:
+		if block.Bellatrix == nil {
 			return errors.Errorf("%s block is nil", block.Version.String())
 		}
-		sb := &altair.SignedBeaconBlock{
-			Message:   block.Altair,
+		sb := &bellatrix.SignedBeaconBlock{
```
y0sher, line 474: if we're deprecating support for `phase0` and `altair` here, why leave bellatrix?

types/testingutils/ssv_msgs.go
```diff
@@ -52,13 +51,6 @@ var TestAggregatorConsensusData = &types.ConsensusData{
 }
 var TestAggregatorConsensusDataByts, _ = TestAggregatorConsensusData.Encode()
 
-var TestProposerConsensusData = &types.ConsensusData{
```
y0sher, line 55: how come we don't use all the code we deleted here? I thought we are not deprecating Bellatrix.

types/spectest/tests/consensusdata/proposer/versioned_block_spectest.go
```diff
@@ -60,6 +60,10 @@ func (test *ProposerSpecTest) Run(t *testing.T) {
 			require.NotNil(t, vBlk.Bellatrix)
 			blkSSZ, err = vBlk.Bellatrix.MarshalSSZ()
 			require.NoError(t, err)
+		case spec.DataVersionCapella:
```
alonmuroch, line 63: did we add any tests?

types/testingutils/beacon_node_versioned.go
```diff
@@ -174,8 +242,64 @@ var TestingInvalidDutySlotV = func(version spec.DataVersion) phase0.Slot {
...
+	err := json.Unmarshal(capellaBlock, &res)
+	if err != nil {
+		panic(err)
+	}
+	// using TestingDutySlotCapella to keep the consistency with TestingProposerDutyV Capella slot
+	res.Slot = TestingDutySlotCapella
+	return &res
+}()
+
+var TestingBlindedBeaconBlockCapella = func() *apiv1capella.BlindedBeaconBlock {
```
alonmuroch, line 264: why is the full block taken from a json and the versioned one generated here?

alonmuroch (changes requested):

types/spectest/tests/consensusdata/proposer/versioned_block_spectest.go
y0sher: @GalRogozinski I added pretty much the same question on top of one of your old comments, in case you missed - 
I know we're trying to keep small PRs but adding Capella but not adding Capella tests on the same PR kinda misses the point imho.

types/spectest/tests/ssz/ssz_spectest.go
GalRogozinski: Yes,

Note we don't fully rely on a third party lib, as we copy-pasted from there :-P
See https://github.com/bloxapp/ssv-spec/pull/234#discussion_r1193130165

So we must test

types/spectest/tests/ssz/withdrawals_marshaling_validation.go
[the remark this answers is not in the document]
GalRogozinski: Note that the base branch here is not `main`. 
When we merge to `main` the branch will have all tests

types/testingutils/beacon_node.go
GalRogozinski: I don't think we really have any tests that use Altair.
We do have tests that use Bellatrix though

types/spectest/tests/consensusdata/proposer/versioned_block_spectest.go
GalRogozinski: I will write the same answer again.
This is not a merge to `main` branch.
When we merge the long-lived branch to `main` it will be completed.
We do have a happy-flow example though, mainly for comments so that the subsequent PRs will be created with initial feedback.

types/testingutils/beacon_node_versioned.go
olegshmuelov: I'm unsure about the issue with the JSON format. Firstly, it improves readability, and secondly, it's utilized by a reliable package (go-eth2-client) that the entire SSV spec relies on.

Apart from that, I believe the question might contain a typo. Alon's question states:
> "why is the full block taken from a JSON and the versioned one generated here?"

I assume he meant "blinded" instead of "versioned." @alonmuroch, please correct me if I'm mistaken.

Addressing the question:
- Why not? We employ the same logic to construct the blinded block from the full block. The only difference is that the Bellatrix full block was manually built, whereas Capella's block is derived from the JSON.

By the way, the [SSZTransactions](https://github.com/olegshmuelov/ssv-spec/blob/8a9e52945bf721763e7fe7f537f9b2a2d8cdf401/types/ssz.go#L53) were created for a specific purpose:
1. to build the blinded block
2. verify that the hash root of the full data block matches the blinded one. You can refer to the [related test](https://github.com/olegshmuelov/ssv-spec/blob/8a9e52945bf721763e7fe7f537f9b2a2d8cdf401/types/testingutils/beacon_node_test.go#L10) for more information.

That's why we've maintained the same design and introduced [SSZWithdrawals](https://github.com/olegshmuelov/ssv-spec/blob/8a9e52945bf721763e7fe7f537f9b2a2d8cdf401/types/ssz.go#L87) for Capella, to ensure that the hash root of the full data Capella block matches the blinded one. With that being said, we can deprecate SSZWithdrawals and use the blinded Capella block in JSON format from go-eth2-client, while still retaining the test to verify that their hash roots are identical.

GalRogozinski: In a private conversation @olegshmuelov convinced me.
It is better to have json only copy-paste from `go-eth-client` and get rid of some custom `SSZ` functions we have

alonmuroch (approved):

ssv/spectest/tests/runner/full_happy_flow_sc.go
GalRogozinski: The deprecated `SSVDecidingMsgs` method used hardcoded version.
Now the version used in the consensus data is passed to this method.

Since `SyncCommitteeContributionConsensusData` wasn't affected by the Capella fork we didn't create a factory for it that creates different variations.

y0sher: I still think its confusing that most V funcs recv a version and this one not. it felt like V is a convention and it's now broken.

y0sher (approved):

ssv/spectest/tests/runner/full_happy_flow_sc.go
GalRogozinski: @olegshmuelov 
I kinda agree with @y0sher. Maybe we should quickly remove the V before merge?

olegshmuelov: If we intend to remove it, we should follow these steps:
- Create a separate PR dedicated to the removal of the V functions, as it is not directly related to this PR.
- Keep in mind that this change will impact approximately ~120 files.
- It would be better to remove the V functions once all the tests using SC (State Comparison) functions have been updated. This approach ensures better code readability and clarity in determining if a test uses SC functions or not.
- In addition, there are certain versioned functions that have corresponding non-versioned counterparts. It is necessary to deprecate the non-versioned functions before proceeding with their removal.

PR #238, opened 2023-05-14
Remove non-versioned functions

GalRogozinski:
### Description

Remove non versioned code


y0sher (approved):
approved.
1 question: why no rename all V funcs back to the regular names now?

GalRogozinski:
@y0sher good question!

The main reason is that we are now trying to create PRs with as little changes as possible. They leave less room for error and take less time to review.
Renaming all V functions will create a massive change that spans many files.

alonmuroch (approved):

PR #240, opened 2023-05-16
Pre consensus fix

alonmuroch:
- fixes round change (not prev prepared) pre-consensus justification
- fixing adding pre-consensus justification when calling baseRunner.decide

### DIFF
--- a/ssv/runner.go
+++ b/ssv/runner.go
@@ -184,6 +184,9 @@ func (b *BaseRunner) basePartialSigMsgProcessing(
 		}
 	}
 
+	// added to pre-consensus justification container
+	b.State.preConsensusJustificationContainer = append(b.State.preConsensusJustificationContainer, signedMsg)
+
 	return anyQuorum, roots, nil
 }
 
--- a/ssv/runner_state.go
+++ b/ssv/runner_state.go
@@ -20,6 +20,9 @@ type State struct {
 	StartingDuty *types.Duty
 	// flags
 	Finished bool // Finished marked true when there is a full successful cycle (pre, consensus and post) with quorum
+
+	// preConsensusJustificationContainer holds signed partial signature messages for pre-consensus
+	preConsensusJustificationContainer []*types.SignedPartialSignatureMessage
 }
 
 func NewRunnerState(quorum uint64, duty *types.Duty) *State {
@@ -29,7 +32,13 @@ func NewRunnerState(quorum uint64, duty *types.Duty) *State {
 
 		StartingDuty: duty,
 		Finished:     false,
+
+		preConsensusJustificationContainer: []*types.SignedPartialSignatureMessage{},
 	}
+}
+
+func (pcs *State) GetPreConsensusJustification() []*types.SignedPartialSignatureMessage {
+	return pcs.preConsensusJustificationContainer
 }
 
 // ReconstructBeaconSig aggregates collected partial beacon sigs
--- a/qbft/round_change.go
+++ b/qbft/round_change.go
@@ -311,7 +311,9 @@ func getRoundChangeData(state *State, config IConfig, instanceStartValue []byte)
 
 		return state.LastPreparedRound, r, state.LastPreparedValue, justifications, nil
 	}
-	return NoRound, [32]byte{}, nil, nil, nil
+
+	// not prepared, return start value as full value for pre-consensus justification
+	return NoRound, [32]byte{}, instanceStartValue, nil, nil
 }
 
 // CreateRoundChange
--- a/qbft/spectest/tests/create_message_spectest.go
+++ b/qbft/spectest/tests/create_message_spectest.go
@@ -18,11 +18,11 @@ const (
 
 type CreateMsgSpecTest struct {
 	Name                                             string
-	Value                                            [32]byte
+	Value                                            []byte
 	Round                                            qbft.Round
 	RoundChangeJustifications, PrepareJustifications []*qbft.SignedMessage
 	CreateType                                       string
-	ExpectedRoot                                     string
+	ExpectedSSZRoot                                  string
 	ExpectedError                                    string
 }
 
@@ -31,18 +31,18 @@ func (test *CreateMsgSpecTest) Run(t *testing.T) {
 	var lastErr error
 	switch test.CreateType {
 	case CreateProposal:
-		msg, lastErr = test.createProposal()
+		msg, lastErr = test.createProposal(t)
 	case CreatePrepare:
-		msg, lastErr = test.createPrepare()
+		msg, lastErr = test.createPrepare(t)
 	case CreateCommit:
-		msg, lastErr = test.createCommit()
+		msg, lastErr = test.createCommit(t)
 	case CreateRoundChange:
-		msg, lastErr = test.createRoundChange()
+		msg, lastErr = test.createRoundChange(t)
 	default:
 		t.Fail()
 	}
 
-	r, err := msg.GetRoot()
+	r, err := msg.HashTreeRoot()
 	if err != nil {
 		lastErr = err
 	}
@@ -53,10 +53,10 @@ func (test *CreateMsgSpecTest) Run(t *testing.T) {
 		require.NoError(t, lastErr)
 	}
 
-	require.EqualValues(t, test.ExpectedRoot, hex.EncodeToString(r[:]))
+	require.EqualValues(t, test.ExpectedSSZRoot, hex.EncodeToString(r[:]))
 }
 
-func (test *CreateMsgSpecTest) createCommit() (*qbft.SignedMessage, error) {
+func (test *CreateMsgSpecTest) createCommit(t *testing.T) (*qbft.SignedMessage, error) {
 	ks := testingutils.Testing4SharesSet()
 	state := &qbft.State{
 		Share: testingutils.TestingShare(ks),
@@ -64,10 +64,13 @@ func (test *CreateMsgSpecTest) createCommit() (*qbft.SignedMessage, error) {
 	}
 	config := testingutils.TestingConfig(ks)
 
-	return qbft.CreateCommit(state, config, test.Value)
+	r, err := qbft.HashDataRoot(test.Value)
+	require.NoError(t, err)
+
+	return qbft.CreateCommit(state, config, r)
 }
 
-func (test *CreateMsgSpecTest) createPrepare() (*qbft.SignedMessage, error) {
+func (test *CreateMsgSpecTest) createPrepare(t *testing.T) (*qbft.SignedMessage, error) {
 	ks := testingutils.Testing4SharesSet()
 	state := &qbft.State{
 		Share: testingutils.TestingShare(ks),
@@ -75,10 +78,13 @@ func (test *CreateMsgSpecTest) createPrepare() (*qbft.SignedMessage, error) {
 	}
 	config := testingutils.TestingConfig(ks)
 
-	return qbft.CreatePrepare(state, config, test.Round, test.Value)
+	r, err := qbft.HashDataRoot(test.Value)
+	require.NoError(t, err)
+
+	return qbft.CreatePrepare(state, config, test.Round, r)
 }
 
-func (test *CreateMsgSpecTest) createProposal() (*qbft.SignedMessage, error) {
+func (test *CreateMsgSpecTest) createProposal(t *testing.T) (*qbft.SignedMessage, error) {
 	ks := testingutils.Testing4SharesSet()
 	state := &qbft.State{
 		Share: testingutils.TestingShare(ks),
@@ -86,14 +92,15 @@ func (test *CreateMsgSpecTest) createProposal() (*qbft.SignedMessage, error) {
 	}
 	config := testingutils.TestingConfig(ks)
 
-	return qbft.CreateProposal(state, config, test.Value[:], test.RoundChangeJustifications, test.PrepareJustifications)
+	return qbft.CreateProposal(state, config, test.Value, test.RoundChangeJustifications, test.PrepareJustifications)
 }
 
-func (test *CreateMsgSpecTest) createRoundChange() (*qbft.SignedMessage, error) {
+func (test *CreateMsgSpecTest) createRoundChange(t *testing.T) (*qbft.SignedMessage, error) {
 	ks := testingutils.Testing4SharesSet()
 	state := &qbft.State{
-		Share: testingutils.TestingShare(ks),
-		ID:    []byte{1, 2, 3, 4},
+		Share:            testingutils.TestingShare(ks),
+		PrepareContainer: qbft.NewMsgContainer(),
+		ID:               []byte{1, 2, 3, 4},
 	}
 	config := testingutils.TestingConfig(ks)
 
@@ -109,7 +116,7 @@ func (test *CreateMsgSpecTest) createRoundChange() (*qbft.SignedMessage, error) 
 		}
 	}
 
-	return qbft.CreateRoundChange(state, config, 1, test.Value[:])
+	return qbft.CreateRoundChange(state, config, 1, test.Value)
 }
 
 func (test *CreateMsgSpecTest) TestName() string {
--- a/qbft/spectest/tests/messages/create_commit.go
+++ b/qbft/spectest/tests/messages/create_commit.go
@@ -5,10 +5,10 @@ import "github.com/bloxapp/ssv-spec/qbft/spectest/tests"
 // CreateCommit tests creating a commit msg
 func CreateCommit() tests.SpecTest {
 	return &tests.CreateMsgSpecTest{
-		CreateType:   tests.CreateCommit,
-		Name:         "create commit",
-		Value:        [32]byte{1, 2, 3, 4},
-		Round:        10,
-		ExpectedRoot: "0834b51f3c87d4aba362d7e2eeb4172d22e0ef18d4dfadd37e8c9ceb62c7719d",
+		CreateType:      tests.CreateCommit,
+		Name:            "create commit",
+		Value:           []byte{1, 2, 3, 4},
+		Round:           10,
+		ExpectedSSZRoot: "a30c2e98de75d97701f3d32026b6a6b67df7bac2aa2804ef7ea481ca03542850",
 	}
 }
--- a/qbft/spectest/tests/messages/create_prepare.go
+++ b/qbft/spectest/tests/messages/create_prepare.go
@@ -5,10 +5,10 @@ import "github.com/bloxapp/ssv-spec/qbft/spectest/tests"
 // CreatePrepare tests creating a prepare msg
 func CreatePrepare() tests.SpecTest {
 	return &tests.CreateMsgSpecTest{
-		CreateType:   tests.CreatePrepare,
-		Name:         "create prepare",
-		Value:        [32]byte{1, 2, 3, 4},
-		Round:        10,
-		ExpectedRoot: "2eba5b18818e0ec94f5c02ff7abc8ca932ed5d1f32a115197fbaa14247a39cb2",
+		CreateType:      tests.CreatePrepare,
+		Name:            "create prepare",
+		Value:           []byte{1, 2, 3, 4},
+		Round:           10,
+		ExpectedSSZRoot: "f1c182dc944840b3cdd182f82ce891a9b527cade2033ff5b371e1ee1086841e2",
 	}
 }
--- a/qbft/spectest/tests/messages/create_proposal.go
+++ b/qbft/spectest/tests/messages/create_proposal.go
@@ -7,9 +7,9 @@ import (
 // CreateProposal tests creating a proposal msg, not previously prepared
 func CreateProposal() tests.SpecTest {
 	return &tests.CreateMsgSpecTest{
-		CreateType:   tests.CreateProposal,
-		Name:         "create proposal",
-		Value:        [32]byte{1, 2, 3, 4},
-		ExpectedRoot: "dfb0a692281b916b1d037df44f5c742f13ac3ee207ea0082cc3ca2afff34e178",
+		CreateType:      tests.CreateProposal,
+		Name:            "create proposal",
+		Value:           []byte{1, 2, 3, 4},
+		ExpectedSSZRoot: "443161d9ea4f2e4abd4c0545aeb4aa99bf40e8a22f59d9917c2dcb7f2f04c9f1",
 	}
 }
--- a/qbft/spectest/tests/messages/create_proposal_not_prev_prepared.go
+++ b/qbft/spectest/tests/messages/create_proposal_not_prev_prepared.go
@@ -13,12 +13,12 @@ func CreateProposalNotPreviouslyPrepared() tests.SpecTest {
 	return &tests.CreateMsgSpecTest{
 		CreateType: tests.CreateProposal,
 		Name:       "create proposal not previously prepared",
-		Value:      [32]byte{1, 2, 3, 4},
+		Value:      []byte{1, 2, 3, 4},
 		RoundChangeJustifications: []*qbft.SignedMessage{
 			testingutils.TestingProposalMessageWithRound(ks.Shares[1], types.OperatorID(1), 2),
 			testingutils.TestingProposalMessageWithRound(ks.Shares[2], types.OperatorID(2), 2),
 			testingutils.TestingProposalMessageWithRound(ks.Shares[3], types.OperatorID(3), 2),
 		},
-		ExpectedRoot: "45f15e2dc3a0c1bd13aa977ead44abfabab094864db4a77dcd0cd9a30e87c3bb",
+		ExpectedSSZRoot: "4a62e3fea5a175243591ebc64728be36ef395c3cd6b2f31e23afe5c7ce1b9811",
 	}
 }
--- a/qbft/spectest/tests/messages/create_proposal_prev_prepared.go
+++ b/qbft/spectest/tests/messages/create_proposal_prev_prepared.go
@@ -13,7 +13,7 @@ func CreateProposalPreviouslyPrepared() tests.SpecTest {
 	return &tests.CreateMsgSpecTest{
 		CreateType: tests.CreateProposal,
 		Name:       "create proposal previously prepared",
-		Value:      [32]byte{1, 2, 3, 4},
+		Value:      []byte{1, 2, 3, 4},
 		RoundChangeJustifications: []*qbft.SignedMessage{
 			testingutils.TestingRoundChangeMessageWithRound(ks.Shares[1], types.OperatorID(1), 2),
 			testingutils.TestingRoundChangeMessageWithRound(ks.Shares[2], types.OperatorID(2), 2),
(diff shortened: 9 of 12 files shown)


ssv/runner.go
```diff
@@ -48,6 +48,7 @@ type BaseRunner struct {
 	BeaconRoleType types.BeaconRole
 
 	// highestDecidedSlot holds the highest decided duty slot and gets updated after each decided is reached
+	// used to validate pre-consensus justification which can not have any lower duty slot than highestDecidedSlot
 	highestDecidedSlot spec.Slot
```
GalRogozinski, line 52: now that the controller's height is also the highest decided slot are we sure we really need this?

GalRogozinski (commented):
I don't see any changes to Baserunner.Decide() like the description said.

ssv/spectest/tests/msgcontainer/duplicate_msg.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+// DuplicateMsg tests adding a duplicate msg (same signer)
+func DuplicateMsg() tests.SpecTest {
+	ks := testingutils.Testing4SharesSet()
+
+	return &SpecTest{
+		Name: "duplicate msg",
+		MsgsToAdd: []*types.SignedPartialSignatureMessage{
+			testingutils.PreConsensusRandaoDifferentSignerMsgV(ks.Shares[1], ks.Shares[1], 1, 1, spec.DataVersionBellatrix),
```
GalRogozinski, line 17: why not use `PreConsensusRandaoMsgV`?

ssv/runner.go
```diff
@@ -156,13 +156,20 @@ func (b *BaseRunner) basePostConsensusMsgProcessing(runner Runner, signedMsg *ty
...
-// basePartialSigMsgProcessing adds an already validated partial msg to the container, checks for quorum and returns true (and roots) if quorum exists
+// basePartialSigMsgProcessing adds an already validated partial msg to the container,
+// checks for quorum and returns true (and roots) if we should process the msg
 func (b *BaseRunner) basePartialSigMsgProcessing(
 	signedMsg *types.SignedPartialSignatureMessage,
 	container PartialSignatureContainer,
 ) (bool, [][32]byte, error) {
-	container[signedMsg.Signer] = signedMsg
 
+	if b.Share.HasQuorum(len(container)) {
```
alonmuroch, line 166: need some explanation here

GalRogozinski: check that quorum happened once

ssv/pre_consensus_justification.go
```diff
@@ -9,23 +9,15 @@ import (
 
 // correctQBFTState returns true if QBFT controller state requires pre-consensus justification
 func (b *BaseRunner) correctQBFTState(msg *qbft.SignedMessage) bool {
```
GalRogozinski, line 11: func name doesn't reflect functionality. 
requiresPreCosensus

ssv/pre_consensus_justification.go
```diff
@@ -9,23 +9,15 @@ import (
...
-
-	return firstHeightNotDecided || notFirstHeightDecided
+	// mind race conditions
+	return b.QBFTController.Height < msg.Message.Height
 }
 
-// shouldProcessingJustificationsForHeight returns true if pre-consensus justification should be processed, false otherwise
-func (b *BaseRunner) shouldProcessingJustificationsForHeight(msg *qbft.SignedMessage) bool {
+// shouldProcessingJustifications returns true if pre-consensus justification should be processed, false otherwise
+func (b *BaseRunner) shouldProcessingJustifications(msg *qbft.SignedMessage) bool {
```
GalRogozinski, line 17: rename shouldProcessJustification

ssv/pre_consensus_justification.go
```diff
@@ -9,23 +9,15 @@ import (
...
-
-	// firstHeightNotDecided is true if height == 0 (special case) and did not start yet
-	firstHeightNotDecided := inst == nil && b.QBFTController.Height == msg.Message.Height && msg.Message.Height == qbft.FirstHeight
-
-	// notFirstHeightDecided returns true if height != 0, height decided and the message is for next height
-	notFirstHeightDecided := decidedInstance && msg.Message.Height > qbft.FirstHeight && b.QBFTController.Height+1 == msg.Message.Height
-
-	return firstHeightNotDecided || notFirstHeightDecided
+	// mind race conditions
+	return b.QBFTController.Height < msg.Message.Height
```
GalRogozinski, line 13: 1. no need for a separate func, just add it to shouldProcessJustifications
2. this means after the first valid pre-consensus justification msg, the next msg will not process pre-consensus justification. This means we need dedicated tests for it

ssv/runner.go
GalRogozinski: we can remove it and use the controller's height instead (please add tests)

MatheusFranco99 (approved):

ssv/runner.go
```diff
@@ -57,23 +54,16 @@ func NewBaseRunner(
 	controller *qbft.Controller,
 	beaconNetwork types.BeaconNetwork,
 	beaconRoleType types.BeaconRole,
-	highestDecidedSlot spec.Slot,
```
alonmuroch, line 60: Explanation: Since we switched to having qbft instances start with duty slot as height there is no need for highest decided anymore

ssv/runner.go
alonmuroch: but why does it need to change?

ssv/runner_state.go
```diff
@@ -12,8 +12,8 @@ import (
 
 // State holds all the relevant progress the duty execution progress
 type State struct {
-	PreConsensusContainer  *PartialSigContainer
-	PostConsensusContainer *PartialSigContainer
+	PreConsensusContainer  PartialSignatureContainer
```
alonmuroch, line 15: I don't fully understand this change? why is this part of this PR?

ssv/partial_sig_container.go
```diff
@@ -1,39 +1,25 @@
 package ssv
 
 import (
-	"encoding/hex"
+	"bytes"
```
alonmuroch, line 4: why are all those changes (partial sig container) relevant to this PR?

ssv/pre_consensus_justification.go
```diff
@@ -1,35 +1,23 @@
...
-	firstHeightNotDecided := inst == nil && b.QBFTController.Height == msg.Message.Height && msg.Message.Height == qbft.FirstHeight
-
-	// notFirstHeightDecided returns true if height != 0, height decided and the message is for next height
-	notFirstHeightDecided := decidedInstance && msg.Message.Height > qbft.FirstHeight && b.QBFTController.Height+1 == msg.Message.Height
-
-	return firstHeightNotDecided || notFirstHeightDecided
-}
-
-// shouldProcessingJustificationsForHeight returns true if pre-consensus justification should be processed, false otherwise
-func (b *BaseRunner) shouldProcessingJustificationsForHeight(msg *qbft.SignedMessage) bool {
```
alonmuroch, line 25: why  did we remove all those checks? not all message carry pre-consensus justification

ssv/spectest/tests/runner/duties/proposer/propose_blinded_decide_regular.go
```diff
@@ -28,9 +28,8 @@ func ProposeBlindedBlockDecidedRegular() tests.SpecTest {
 					[]types.OperatorID{1, 2, 3},
 					qbft.Height(testingutils.TestingDutySlotV(spec.DataVersionBellatrix)),
 					testingutils.ProposerMsgID,
-					testingutils.TestProposerConsensusDataBytsV(spec.DataVersionBellatrix),
-				),
-				nil),
+					testingutils.TestProposerConsensusDataBytsV(ks, spec.DataVersionBellatrix),
```
alonmuroch, line 31: we have a ton of those changes which just add "changed" files to the PR

ssv/runner_state.go
GalRogozinski: See commit 91ba801c

You are its author. So maybe you can answer the question 😅

ssv/partial_sig_container.go
GalRogozinski: https://github.com/bloxapp/ssv-spec/commit/91ba801c3fd2c1734800eefa980e10e2055621b9

You added it here but I agree. Were you working on something else, or you want to keep it here?

ssv/pre_consensus_justification.go
GalRogozinski: We now have simplified this. Compare more closely with `ShouldProcessJustifications`

ssv/spectest/generate/state_comparison/newduty_MultiStartNewRunnerDutySpecTest/duplicate_duty_finished/attester.json
```diff
@@ -147,7 +141,7 @@
 								"CommitteeLength": 128,
 								"CommitteesAtSlot": 36,
 								"ValidatorCommitteeIndex": 11,
-								"ValidatorSyncCommitteeIndices": null
+								"ValidatorSyncCommitteeIndices": []
```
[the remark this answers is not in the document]
GalRogozinski: Looks like there was a problem with passing tests.
Is it important? I remember once I asked you about something similar and you said that changes between null or empty arrays are not so important.

ssv/runner.go
```diff
@@ -234,6 +212,11 @@ func (b *BaseRunner) hasRunningDuty() bool {
 }
 
 func (b *BaseRunner) ShouldProcessDuty(duty *types.Duty) error {
+	// first height special case
```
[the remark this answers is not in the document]
GalRogozinski: *Why do we need it?*

It is a special case because of an implementation detail. Currently we bootstrap the Controller's state with 0. Also the first duty can also be decided with height 0. In all other cases the duty will be decided on a height greater than the controller.

We can do a smart fix in spec so that 0 height will work like any other height and won't need special conditions in our algos (not in this PR).

However, even if we do that maybe we should have first height tests. We can't do check for all heights but first height is suspect for bugs.

*do we test it?*
See `valid_first_height.go`

ssv/spectest/tests/runner/duties/proposer/propose_blinded_decide_regular.go
GalRogozinski: Many of them come from your commit
0266dd81

I'll try to clean it up

ssv/runner_state.go
GalRogozinski: The previous PartialSigMessageContainer can't return the full PartialSigMessage which is a requirement for putting the preconsensus justification in the CD.
I initially thought you did this as part of the spec-rethink but now I think you did this for this PR and I propose we leave those changes

ssv/partial_sig_container.go
GalRogozinski: The previous PartialSigMessageContainer can't return the full PartialSigMessage which is a requirement for putting the preconsensus justification in the CD.
I initially thought you did this as part of the spec-rethink but now I think you did this for this PR and I propose we leave those changes

MatheusFranco99 (approved):
Looks good to me

ssv/runner.go
alonmuroch: ?

alonmuroch:
approved

GalRogozinski (dismissed):

ssv/runner.go
```diff
@@ -155,30 +139,24 @@ func (b *BaseRunner) basePostConsensusMsgProcessing(runner Runner, signedMsg *ty
...
+		container[signedMsg.Signer] = signedMsg
+		// return false if we already have quorum
+		return false, container.Roots(), nil
 	}
 
-	return anyQuorum, roots, nil
+	container[signedMsg.Signer] = signedMsg
+	if b.Share.HasQuorum(len(container)) {
+		return true, container.Roots(), nil
+	}
```
moshe-blox, line 158: Isn't this skipping the check that the signers have signed the same Beacon roots?

I imagine that if operators sign different aggregation messages, and the signing roots aren't checked, then the Runner may falsely think it has quorum and proceed to fail reconstructing the signature. If true, this means that one faulty operator may ruin the duty, no?

On another note, it seems to me that `container.Roots()` may cause another consensus issue. Since it only returns the signing roots from one `SignedPartialSignatureMessage`, this means that if the operators have signed different roots, then we'd only try to reconstruct & submit those from the first message in the container (which is a map so the order is non-deterministic).

In case anyone wants to know why multiple signing roots are necessary:
in sync committee aggregation duties, although extremely rare, one validator may participate in several subcommittees, in which case there will be several roots under `signedMsg.Message.Messages`, one for each subcommittee.

GalRogozinski: So double check me please @moshe-blox:

1. Note that before `basePartialSigMsgProcessing` is called we always call some validation method. See:
https://github.com/bloxapp/ssv-spec/blob/daec5450ce7d808a7e643081a5109b4eb7c55f07/ssv/pre_consensus_justification.go#L20
https://github.com/bloxapp/ssv-spec/blob/0a414236cef5eca4dc0f7da464034325869a6477/ssv/runner_validations.go#L29
https://github.com/bloxapp/ssv-spec/blob/0a414236cef5eca4dc0f7da464034325869a6477/ssv/runner_validations.go#L12

Those validations should take care of the problem you mentioned. I will add method comments though to make this clear though.

2. Having said this it might be smart when it comes to the implementation to add this check as part of a defensive programming strategy. If a bug is introduced in validation you have a fall back mechanism. Or you can add comments and have good tests (your choice, just giving my 2 cents.

3. Same with the problem concerning `container.Roots()`. Not sure if I need to add comments though because this is just a getter method to what was inserted to the container.

GalRogozinski: actually @moshe-blox I think the existing comment is fine:
https://github.com/bloxapp/ssv-spec/blob/ecc6a3f944d3dcb2387be71eb29920276f11a131/ssv/runner.go#L142

If you think it should be clearer you can make a suggestion :-)

moshe-blox: hey @GalRogozinski 

Thanks for the clarifying, I see that this issue can't happen because the roots must be identical to what the verifier got from their own Beacon node.

ssv/partial_sig_container.go
```diff
@@ -43,10 +29,43 @@ func (ps *PartialSigContainer) ReconstructSignature(root [32]byte, validatorPubK
 	return signature.Serialize(), nil
 }
 
-func (ps *PartialSigContainer) HasQuorum(root [32]byte) bool {
-	return uint64(len(ps.Signatures[rootHex(root)])) >= ps.Quorum
+// SignatureForRoot returns a map of signer and signature for a specific root
+func (ps PartialSignatureContainer) SignatureForRoot(root [32]byte) map[types.OperatorID][]byte {
```
y0sher, line 33: not trying to nit pick but I think this function should be named 
```suggestion
func (ps PartialSignatureContainer) SignaturesForRoot(root [32]byte) map[types.OperatorID][]byte {
```

ssv/partial_sig_container.go
```diff
@@ -43,10 +29,43 @@ func (ps *PartialSigContainer) ReconstructSignature(root [32]byte, validatorPubK
...
+	if len(ps) > 0 {
+		ret := make([][32]byte, 0)
+		for _, sigMsg := range ps {
+			for _, msg := range sigMsg.Message.Messages {
+				ret = append(ret, msg.SigningRoot)
+			}
+			break // only need to iterate first msg
+		}
+		return ret
+	}
```
y0sher, line 56: we could do 
```golang
if len(ps) =< 0 {
return nil // or [][32]byte{}
}
```
 to reduce nesting and add readability

ssv/pre_consensus_justification.go
```diff
@@ -39,8 +27,8 @@ func (b *BaseRunner) validatePreConsensusJustifications(data *types.ConsensusDat
 		return errors.New("wrong beacon role")
 	}
 
-	if data.Duty.Slot <= highestDecidedDutySlot {
-		return errors.New("duty.slot <= highest decided slot")
+	if qbft.Height(data.Duty.Slot) <= b.QBFTController.Height && b.QBFTController.Height != qbft.FirstHeight {
```
y0sher, line 30: if we are on first height (which will never happen I guess) then its ok that the duty height is smaller than the controller height? I understand why its ok that its equal but not lower.

ssv/runner.go
```diff
@@ -167,30 +151,24 @@ func (b *BaseRunner) basePostConsensusMsgProcessing(runner Runner, signedMsg *ty
...
-			continue
-		}
-
-		quorum := container.HasQuorum(msg.SigningRoot)
-		if quorum {
-			roots = append(roots, msg.SigningRoot)
-			anyQuorum = true
-		}
+
+	if b.Share.HasQuorum(len(container)) {
```
y0sher, line 161: it seems like `HasQuorum` just checks the number of sigs we have, don't we want to also check validity before we say we have quorum ?

ssv/spectest/generate/state_comparison/tests_MultiMsgProcessingSpecTest/consensus_decided_invalid_value/aggregator.json
```diff
@@ -2,19 +2,152 @@
...
+										"Signer": 2
+								},
+								"3": {
+										"Message": {
+												"Type": 2,
+												"Slot": 12,
+												"Messages": [
+														{
+																"PartialSignature": "qFQDMkcUT9KZK3wO4GHZKlEZnJKWbhULlEMkiYC0igtS+snZxJ5tfnZ5oxKvIPRcBVMI0gP/rhHuEgGzpwBuVfuFYclsirb3DZVgPJcdVBjSkhkAt4polkJ/BKEHBdC5",
+																"SigningRoot": [
```
y0sher, line 108: can we make this hex encoded like other roots/signatures/pubkeys?

ssv/spectest/tests/msgcontainer/quorum_post_consensus_msg.go
```diff
@@ -0,0 +1,27 @@
...
+// QuorumPostConsensusMsg tests adding a quorum of post consensus message to container
+func QuorumPostConsensusMsg() tests.SpecTest {
+	ks := testingutils.Testing4SharesSet()
+
+	return &SpecTest{
+		Name: "quorum post consensus message",
+		MsgsToAdd: []*types.SignedPartialSignatureMessage{
+			testingutils.PostConsensusSyncCommitteeContributionMsg(ks.Shares[1], 1, ks),
+			testingutils.PostConsensusSyncCommitteeContributionMsg(ks.Shares[2], 2, ks),
+			testingutils.PostConsensusSyncCommitteeContributionMsg(ks.Shares[3], 3, ks),
```
y0sher, line 18: should we test with messages other than synccommitteecontrib?

ssv/spectest/tests/msgcontainer/test.go
```diff
@@ -0,0 +1,56 @@
...
+		for i, m := range msg.Message.Messages {
+			sig, lastErr = c.ReconstructSignature(m.SigningRoot, testingutils.TestingValidatorPubKey[:])
+			require.EqualValues(t, test.PostReconstructedSignature[i], hex.EncodeToString(sig))
+		}
+		if len(test.ExpectedErr) == 0 {
+			require.NoError(t, lastErr)
+		} else {
+			require.EqualError(t, lastErr, test.ExpectedErr)
+		}
+	}
```
y0sher, line 55: what about checking ExpectedErr even if postreconstructedsignature is empty? or there's no such test?

ssv/spectest/tests/runner/consensus/valid_decided_10_operators.go
```diff
@@ -19,8 +19,8 @@ func ValidDecided10Operators() tests.SpecTest {
 				Name:                    "sync committee contribution",
 				Runner:                  testingutils.SyncCommitteeContributionRunner(ks),
 				Duty:                    &testingutils.TestingSyncCommitteeContributionDuty,
-				Messages:                testingutils.SSVDecidingMsgsV(testingutils.TestSyncCommitteeContributionConsensusData, ks, types.BNRoleSyncCommitteeContribution),
-				PostDutyRunnerStateRoot: "a991b6470a8c7a55f4ce89aea91925c2d80a9a8c4258545cc2fb17cabc388719",
+				Messages:                testingutils.SSVDecidingMsgsV(testingutils.TestSyncCommitteeContributionConsensusData(ks), ks, types.BNRoleSyncCommitteeContribution),
+				PostDutyRunnerStateRoot: "ec47a2194bab81891bd581eaf427af211b53e13d1f77a6703eeb56a693caa878",
```
y0sher, line 23: this is changed because preconsensus now affects the duty runner state?

ssv/spectest/tests/runner/full_happy_flow.go
```diff
@@ -69,19 +69,12 @@ func FullHappyFlow() tests.SpecTest {
...
-							testingutils.SSVMsgAggregator(nil, testingutils.PostConsensusAggregatorMsg(ks.Shares[2], 2)),
-							testingutils.SSVMsgAggregator(nil, testingutils.PostConsensusAggregatorMsg(ks.Shares[3], 3)),
-						}...,
-					)...,
+					testingutils.SSVDecidingMsgsV(testingutils.TestAggregatorConsensusData(ks), ks, types.BNRoleAggregator), // consensus
+					[]*types.SSVMessage{ // post consensus
+						testingutils.SSVMsgAggregator(nil, testingutils.PostConsensusAggregatorMsg(ks.Shares[1], 1)),
+						testingutils.SSVMsgAggregator(nil, testingutils.PostConsensusAggregatorMsg(ks.Shares[2], 2)),
+						testingutils.SSVMsgAggregator(nil, testingutils.PostConsensusAggregatorMsg(ks.Shares[3], 3)),
+					}...,
```
y0sher, line 77: this duty now doesn't have pre consensus msgs?

ssv/spectest/tests/runner/pre_consensus_justifications/inconsisten_root_count.go
y0sher, line 1: not from this pr but typo in file name

y0sher (changes requested):
Big PR. see my comments, mostly clarifications.

ssv/pre_consensus_justification.go
GalRogozinski: If it is small or equal it returns an error.
On first height it can only be equal.

GalRogozinski: you mean if firstHeight would be a number different than 0?
If it will be 1 then we won't need this special condition in all those places in the code

ssv/partial_sig_container.go
GalRogozinski: done

ssv/partial_sig_container.go
GalRogozinski: done

ssv/runner.go
GalRogozinski: for a message to enter the container it should pass validity checks

ssv/spectest/generate/state_comparison/tests_MultiMsgProcessingSpecTest/consensus_decided_invalid_value/aggregator.json
GalRogozinski: not for this PR

ssv/spectest/tests/msgcontainer/quorum_post_consensus_msg.go
GalRogozinski: maybe...
when I look at this now not sure why this test is part of the PR (it is something Alon added)...
The PR should concentrate on preconsensus...

nevertheless I am adding you point in #264

ssv/spectest/tests/msgcontainer/test.go
GalRogozinski: Currently there is a trick in tests that Alon did to have a non empty `PostReconstructedSignature` filled with empty strings.

In another test he kept the `PostReconstructedSignature` empty to not have an error...
So I'd rather not change his design at this point

ssv/spectest/tests/runner/consensus/valid_decided_10_operators.go
GalRogozinski: the runner contains all msg containers that includes messages with preconsensus justifications, yes...
But I don't see in the logic where we add prec justification for this test... will look into it...

ssv/spectest/tests/runner/full_happy_flow.go
GalRogozinski: `testingutils.SSVDecidingMsgsV(testingutils.TestSyncCommitteeConsensusData, ks, types.BNRoleSyncCommittee), // consensus
`

this line adds precon messages
see `pre_consensus_full_happy_flow`

ssv/spectest/tests/runner/pre_consensus_justifications/inconsisten_root_count.go
GalRogozinski: on one hand the fix for this doesn't belong in this PR...
on the other hand getting 3 busy people to approve a typo fix on a dedicated PR will take ages...

ssv/spectest/tests/runner/consensus/valid_decided_10_operators.go
GalRogozinski: the commit 66c1585f has this [change](https://github.com/bloxapp/ssv-spec/commit/66c1585ff0a1ef79b31ccd2453a37314cb5a0ea8#diff-28956438e46e71f8b726f9fe106f47ac7cfca29ae303a20aacc6e7837a29b506R21)

This should be overridden by deneb imo and the root may need to change again...
I will resolve conflicts and then regenrate jsons and see what's going on

ssv/spectest/tests/runner/pre_consensus_justifications/inconsisten_root_count.go
GalRogozinski: but it will add more conflicts so I won't fix

ssv/partial_sig_container.go
```diff
@@ -43,10 +29,44 @@ func (ps *PartialSigContainer) ReconstructSignature(root [32]byte, validatorPubK
...
+		break // only need to iterate first msg
+	}
+	return ret
 }
 
-func rootHex(r [32]byte) string {
-	return hex.EncodeToString(r[:])
+// AllSorted returns ordered by signer array of signed messages
+func (ps PartialSignatureContainer) AllSorted() []*types.SignedPartialSignatureMessage {
+	ret := make([]*types.SignedPartialSignatureMessage, 0)
```
moshe-blox, line 64: allocate with capacity of `len(ps)`

ssv/partial_sig_container.go
```diff
@@ -43,10 +29,44 @@ func (ps *PartialSigContainer) ReconstructSignature(root [32]byte, validatorPubK
...
+	for _, m := range ps {
+		ret = append(ret, m)
+	}
+	sort.Slice(ret, func(i, j int) bool {
```
moshe-blox, line 68: for less memory allocations:
```suggestion
	ret := make([]*types.SignedPartialSignatureMessage, len(ps))
	for i, m := range ps {
		ret[i] = m
	}
```

ssv/aggregator.go
moshe-blox, line 1: rename `quorum` to `shouldProcess` or the other way around to keep consistency

personally i like `quorum` bc its self-explanatory

moshe-blox (approved):

ssv/partial_sig_container.go
GalRogozinski: will do because you copy it to impl even though ideally spec shouldn't care

ssv/partial_sig_container.go
GalRogozinski: why delete sort?

ssv/aggregator.go
GalRogozinski: the problem is that sometimes you have a quorum but you don't process

PR #162, opened 2023-02-22
ssz: fix ssv sanity tests

nkryuchkov:

PR #243, opened 2023-05-18
Version block support capella sanity tests

olegshmuelov:
### Description
Version block sanity tests.
reference to sanity tests https://github.com/bloxapp/ssv-spec/pull/162

split to #246 #247 #248


olegshmuelov:
split to #246 #247 #248. Closing

PR #259, opened 2023-05-24
Ssv fork update

alonmuroch:
Updated according to https://github.com/bloxapp/SIPs/pull/26


alonmuroch:
Can we trigger fork by simply calculating epoch? 

- What happens to running instances?
- weak subjectivity: sync still needs to provide fork to verify msgs

types/chains.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func (domainType DomainType) GetForksData() []*ForkData {
+	switch domainType.GetChain() {
+	case MainnetChain:
+		return mainnetForks()
+	default:
+		return []*ForkData{}
+	}
+}
```
GalRogozinski, line 44: shouldn't we have forkdata for our public testnet, Jato?
I assume from now instead of changing testnets. we will start forking our public testnet?

types/chains.go
```diff
@@ -0,0 +1,54 @@
...
+	ShifuChain   = 0x2
+	JatoChain    = 0x3
+)
+
+// DomainTypes represent specific forks for specific chains, messages are signed with the domain type making 2 messages from different domains incompatible
+var (
+	GenesisMainnet = DomainType{0x0, 0x0, MainnetChain, 0x0}
+	PrimusTestnet  = DomainType{0x0, 0x0, PrimusChain, 0x0}
+	ShifuTestnet   = DomainType{0x0, 0x0, ShifuChain, 0x0}
+	ShifuV2Testnet = DomainType{0x0, 0x0, ShifuChain, 0x1}
```
GalRogozinski, line 21: do we need to have deprectated nets here?

GalRogozinski (commented):
Since we only have the genesis fork you decided to not have triggering logic for now?

types/chains.go
alonmuroch: no forks yet, if we will we can add them

types/chains.go
alonmuroch: we can later

types/chains.go
```diff
@@ -0,0 +1,54 @@
...
+// DomainType is a unique identifier for signatures, 2 identical pieces of data signed with different domains will result in different sigs
+type DomainType [4]byte
+
+// Chains are intended to separate different SSV chains. A chain can have many forks in it.
+const (
+	MainnetChain = 0x0
+	PrimusChain  = 0x1
+	ShifuChain   = 0x2
+	JatoChain    = 0x3
+)
```
GalRogozinski, line 14: I think we should have a convention for consts. For example if mainnet prefix with 0 and if testnet prefix with 1:

```
// mainnet
const (
 MainnetFork0 = 0x00
 MainnetFork1 = 0x01
 MainnetFork2 = 0x02
 )
 
 // testnet (Jato)
 const (
    JatoFork0 = 0x10
    JatoFork1 = 0x11
 )
```

types/chains.go
```diff
@@ -0,0 +1,54 @@
...
+	Epoch phase0.Epoch
+	// Domain for the new fork
+	Domain DomainType
+}
+
+func (domainType DomainType) GetChain() byte {
+	return domainType[2]
+}
+
+func (domainType DomainType) GetForksData() []*ForkData {
```
y0sher, line 37: where is this used?

types/chains.go
y0sher: what does the 2 first bytes stand for ?

types/chains.go
GalRogozinski: It should from Runner according to the SIP: https://github.com/bloxapp/SIPs/pull/26/files#diff-ef3380c5833665d8dd7ac5bc173946735e412f9da5ca4c0380d39399b5458af6L75

but it isn't part of this PR for some reason...
@alonmuroch is there a reason for this?

types/chains.go
alonmuroch: future TBD use (just have space for it)

types/chains.go
alonmuroch: not used yet

types/chains.go
```diff
@@ -0,0 +1,54 @@
...
+	JatoChain    = 0x3
+)
+
+// DomainTypes represent specific forks for specific chains, messages are signed with the domain type making 2 messages from different domains incompatible
+var (
+	GenesisMainnet = DomainType{0x0, 0x0, MainnetChain, 0x0}
+	PrimusTestnet  = DomainType{0x0, 0x0, PrimusChain, 0x0}
+	ShifuTestnet   = DomainType{0x0, 0x0, ShifuChain, 0x0}
+	ShifuV2Testnet = DomainType{0x0, 0x0, ShifuChain, 0x1}
+	JatoTestnet    = DomainType{0x0, 0x0, JatoChain, 0x1}
```
olegshmuelov, line 22: The JatoTestnet should ideally be initialized with `DomainType{0x0, 0x0, JatoChain, 0x0}` 
instead of `DomainType{0x0, 0x0, JatoChain, 0x1}` to maintain consistency with other testnet configurations.

MatheusFranco99 (approved):

types/chains.go
alonmuroch: Sure thing, please make the change

alonmuroch: pleas also add Jato V2

GalRogozinski (changes requested):
Several Changes:

1. Stop using "chains" use "networks" or "networkIDs"
2. "DomainType" has a collision with attestant.io lib we are using
3. We should have `func (networkID NetworkId) GetForksData() []*ForkData` and not use domainType.
4. Note that currently the `identifier` field in controller is initialized once, but it has `DomainType` data that include protocol versions that are a function of the epoch.

GalRogozinski:
replaced by #310

PR #260, opened 2023-05-24
Disable Pre-consensus Liveness

moshe-blox:
This PR comments the call to `processPreConsensusJustification` together with any pre-consensus spec tests.


qbft/round_change.go
```diff
@@ -30,7 +30,7 @@ func (i *Instance) uponRoundChange(
 		roundChangeMsgContainer,
 		valCheck)
 	if err != nil {
-		return errors.Wrap(err, "could not get proposal justification for leading ronud")
```
GalRogozinski, line 33: what does this has to do with precon-liveness?

ssv/runner.go
```diff
@@ -114,8 +114,11 @@ func (b *BaseRunner) baseConsensusMsgProcessing(runner Runner, msg *qbft.SignedM
...
 	}
 
-	if err := b.processPreConsensusJustification(runner, b.highestDecidedSlot, msg); err != nil {
-		return false, nil, errors.Wrap(err, "invalid pre-consensus justification")
+	// TODO: revert `if false` after pre-consensus justification is fixed.
+	if false {
+		if err := b.processPreConsensusJustification(runner, b.highestDecidedSlot, msg); err != nil {
+			return false, nil, errors.Wrap(err, "invalid pre-consensus justification")
+		}
 	}
```
GalRogozinski, line 122: nit:
why not just comment out like you did in `all_tests`? I just find the `if false` funny and weird.
The request changes is because of the other comment.

GalRogozinski (changes requested):

qbft/round_change.go
lior-blox: fixed typo

GalRogozinski (approved):

qbft/round_change.go
GalRogozinski: btw,
@moshe-blox and @lior-blox 
If we later do `git revert` on this commit (the easiest way) this would be undone as well

alonmuroch (approved):

lior-blox (approved):

PR #268, opened 2023-06-02
[Audit Suggestion 6] Update dependencies

nkryuchkov:
Update dependencies to latest released versions reducing the amount of warnings returned by `go list -json -m all | nancy sleuth` and `go list -u -m -json all | go-mod-outdated`.

`github.com/libp2p/go-libp2p-core/peer` was deprecated and causes linter to fail, hence `github.com/libp2p/go-libp2p/core/peer` is used


GalRogozinski:
Did you see #265?
I upgraded only what was easy to update.

We need to think how much time we should spend on this since spec isn't really production code.
Notice some dependencies are not really used by spec atm like libp2p (there is unfinished msg validation in spec atm).

nkryuchkov:
@GalRogozinski 

> Did you see https://github.com/bloxapp/ssv-spec/pull/265?

Yes, but I would update other dependencies as well

> I upgraded only what was easy to update.

Did you face any complexities with upgrading the rest?

> We need to think how much time we should spend on this since spec isn't really production code.

The spec itself isn't but the implementation imports the spec, so some spec code is production code. I spent not much time on that and both linter/tests passed locally. Unfortunately, CI fails right now, I'll try to fix it quickly. If it will require time, I'll revert some upgrades

> Notice some dependencies are not really used by spec atm like libp2p (there is unfinished msg validation in spec atm).

The approach of starting/continuing things like msg validation with the new libp2p version sounds better to me than using an old version and then rewriting it later

types/spectest/generate/tests.json
nkryuchkov, line 1: This files gets changed after running `make generate-jsons` on `main` too, so

GalRogozinski: you can revert this...
We have a test that generates random keys so it walwyas changes the test

nkryuchkov: Oh, I didn't think about this. Thanks! Reverted

GalRogozinski (approved):
Good job

lior-blox (approved):

PR #274, opened 2023-06-15
Remove QBFT instance requirement for an instance to decide for next one to start

alonmuroch:
Removes the condition that a new QBFT instance (and therefore a new runner duty) can't start without the previous one decided.

**Why many SSV tests changed?** Controller height is not determined by the Duty's slot which in the test cases is not equal to FirstHeight like it was previously. This causes many tests to change post root

Following [SIP-X](https://github.com/bloxapp/SIPs/pull/28)


qbft/instance.go
```diff
@@ -90,6 +90,10 @@ func (i *Instance) Broadcast(msg *SignedMessage) error {
 
 // ProcessMsg processes a new QBFT msg, returns non nil error on msg processing error
 func (i *Instance) ProcessMsg(msg *SignedMessage) (decided bool, decidedValue []byte, aggregatedCommit *SignedMessage, err error) {
+	if i.State.Round == CutoffRound {
+		return false, nil, nil, errors.New("round > cutoff round")
+	}
```
GalRogozinski, line 95: Didn't see this is the SIP... only the `timeout` section was changed.
Doesn't it make better sense to finish the `runner` instead?

In any case just to be safer:
```suggestion
	if i.State.Round >= CutoffRound {
		return false, nil, nil, errors.Errorf("round %d > cutoff round %d", i.State.Round, CutoffRound)
	}
```

qbft/timeout.go
```diff
@@ -4,7 +4,14 @@ import (
 	"github.com/pkg/errors"
 )
 
+// CutoffRound which round the instance should stop its timer and progress no further
+const CutoffRound = 20
+
 func (i *Instance) UponRoundTimeout() error {
+	if i.State.Round == CutoffRound {
+		return errors.New("round > cutoff round")
+	}
```
GalRogozinski, line 13:
```suggestion
	if i.State.Round == CutoffRound {
		return errors.Errof("round %d > cutoff round %d", i.State.Round, CutoffRound)
	}
```

qbft/instance.go
alonmuroch: This is the instance, not the runner

alonmuroch:
@GalRogozinski I'm leaving future message processing intact to get tests passing easily. It needs to be removed later on.
Opened an issue #280

qbft/spectest/tests/timeout/round_15.go
```diff
@@ -1,15 +1,14 @@
...
 	"github.com/bloxapp/ssv-spec/qbft/spectest/tests"
 	"github.com/bloxapp/ssv-spec/types"
 	"github.com/bloxapp/ssv-spec/types/testingutils"
 	"github.com/bloxapp/ssv-spec/types/testingutils/comparable"
 )
 
-// Round20 tests calling UponRoundTimeout for round 20, testing state and broadcasted msgs
-func Round20() tests.SpecTest {
+// Round15 tests calling UponRoundTimeout for round 15, testing state and broadcasted msgs
+func Round15() tests.SpecTest {
```
GalRogozinski, line 11: It looks like we are still using round 20 inside

qbft/controller.go
```diff
@@ -42,18 +42,26 @@ func NewController(
...
-func (c *Controller) StartNewInstance(value []byte) error {
-	if err := c.canStartInstanceForValue(value); err != nil {
-		return errors.Wrap(err, "can't start new QBFT instance")
+func (c *Controller) StartNewInstance(height Height, value []byte) error {
+	if err := c.GetConfig().GetValueCheckF()(value); err != nil {
+		return errors.Wrap(err, "value invalid")
+	}
+
+	if height < c.Height {
+		return errors.New("attempting to start an instace with a past or current height")
```
y0sher, line 51: the error says past or current but if `height == `c.Height` we will not fall into this codepath. should we change the error or the check to `=<`?

qbft/controller.go
```diff
@@ -42,18 +42,26 @@ func NewController(
...
 	}
 
 	// only if current height's instance exists (and decided since passed can start instance) bump
-	if c.StoredInstances.FindInstance(c.Height) != nil {
-		c.bumpHeight()
+	if c.StoredInstances.FindInstance(height) != nil {
+		return errors.New("instance already running")
 	}
 
+	// TODO maybe add height param to add and stpre new instance and update controller later
```
y0sher, line 59: typo
```suggestion
	// TODO maybe add height param to add and store new instance and update controller later
```

qbft/instance.go
```diff
@@ -47,6 +49,10 @@ func NewInstance(
 	}
 }
 
+func (i *Instance) ForceStop() {
```
y0sher, line 52: When I saw `ForceStop` I got the impression there's a regular `Stop` and force is kinda stronger. Why not call it just `Stop`?

GalRogozinski: There is a regular stop if we hit round 15

qbft/instance.go
```diff
@@ -202,3 +212,8 @@ func (i *Instance) Encode() ([]byte, error) {
 func (i *Instance) Decode(data []byte) error {
 	return json.Unmarshal(data, &i)
 }
+
+// CanProcessMessages will return true if instance can process messages
+func (i *Instance) CanProcessMessages() bool {
+	return int(i.State.Round) < CutoffRound && !i.forceStop
```
y0sher, line 218: first check forceStop before comparing other vars. 
```suggestion
	return !i.forceStop && int(i.State.Round) < CutoffRound
```

y0sher: Also shouldn't we use forcestop to prevent the instance from broadcasting messages as well?

GalRogozinski: we can give them a chance to broadcast imo

does @alonmuroch agree with me?

qbft/instance.go
y0sher: I see. it just that we not proceed after round 15, but there's not active "stop"

y0sher:
Do we want to keep Sync? for example in `UponDecided` - https://github.com/alonmuroch/ssv-spec/blob/ed3311db734079e9fb26435c76ce25db9a6c3aa3/qbft/decided.go#L45

same in `UponFutureMsg` - https://github.com/alonmuroch/ssv-spec/blob/ed3311db734079e9fb26435c76ce25db9a6c3aa3/qbft/future_msg.go#L16

qbft/instance.go
GalRogozinski: not proceed means we stop I think 🤷

qbft/controller.go
GalRogozinski: This will be fixed when we merge the PR from Alon's repo.
The reason is that tests have an `ExpectedError` field and this change encompasses files not from this PR

qbft/controller.go
GalRogozinski: removed the TODO now

qbft/controller.go
GalRogozinski: https://github.com/alonmuroch/ssv-spec/pull/4/commits/71d9fbe2cb8ba7cb4bd136a7885b20e812323334

GalRogozinski (approved):

qbft/instance.go
y0sher: why tho? if the (a term I just invented) "effectiveness threshold" has past, will broadcasting give them or the network any benefit ?
Anyway it seems it won't cause any harm as well apart from resources, so I'm ok with leaving this now until we know there's a good reason to stop them from broadcasting (resources or other).

y0sher (approved):

qbft/spectest/tests/timeout/round_15.go
GalRogozinski: fixed

qbft/instance.go
GalRogozinski: opened https://github.com/bloxapp/ssv-spec/issues/288

qbft/spectest/generate/state_comparison/*tests.ControllerSpecTest/qbft controller late commit past instance.json
```diff
@@ -2641,56 +2641,6 @@
 																				"PrepareJustification": null
 																		},
 																		"FullData": null
-																},
```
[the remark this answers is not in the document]
GalRogozinski: If you look at the old SC the reason will become clear.
This is a latemsg and we don't process late messages anymore. So it doesn't enter the commit container.

This breaks the old maximal commits on quorum feature of the explorer...
@lior-blox said it was fine before, but let's remind him again.

https://github.com/bloxapp/ssv-spec/pull/274/files/51b5bbd90e3007bdfce00b2d286a968d2e72ec25#diff-7ec4f6d7bd60a0e03ae21b4dcbbd5376744fa5ceaa8f3d9a790e23e015f07bb1L105-L108

lior-blox (approved):

PR #275, opened 2023-06-19
State comparison json

alonmuroch:
Introducing json state comparison per test.

**Goal**

- Readable post state 
- Comparable (via json diff) post state
- Easy to spot changes (json will change) 

**Overview**
- Post test the state is written into a per test file
- When test loads, it will override (for supported test types) the expected post root and state
- The above will be used to verify the test 

(Supports only msg processing tests for now, can be expanded later)

PR #276, opened 2023-06-20
Serialize post states to JSON for QBFT tests

GalRogozinski:
### Description

Replaces #275 

Add infrastructure that enables one to generate the post states of each spec test in json format with `make generate-jsons` command. All generated jsons can be found in `spectest/generate/state_comparison` folder.
It will override the files for each run.

The generated jsons aren't added here because we want to review each test that we add jsons to.

### How was this tested.

1. Jsons were generated correctly
2. Tests all ran correctly

### DIFF
--- a/qbft/spectest/generate/main.go
+++ b/qbft/spectest/generate/main.go
@@ -4,11 +4,14 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/bloxapp/ssv-spec/qbft/spectest/tests"
+	"github.com/bloxapp/ssv-spec/types"
+	"log"
 	"os"
 	"path/filepath"
 	"reflect"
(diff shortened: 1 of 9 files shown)


qbft/spectest/generate/main.go
```diff
@@ -4,25 +4,33 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/bloxapp/ssv-spec/qbft/spectest/tests"
+	"github.com/bloxapp/ssv-spec/types"
+	"log"
```
GalRogozinski, line 8: @lior-blox or @y0sher 
I think the simple log is good enough but maybe we have reasons to use a different one?

alonmuroch (approved):

Issue #270, opened 2023-06-06
Validator Registration Runner needlessly uses qbft controller


GalRogozinski:
### Description
  
 Currently `ssv` has a hotfix that looks like #267. It should be solved in a different way.
 
 Do one of the following:
  
1) Write different logic for `func (r *ValidatorRegistrationRunner) StartNewDuty(duty *types.Duty)`. In the long run refactor more deep and have frameworks for tasks that require only sig aggregation with no consensus.
2) Create a `dumbController` struct that implements the interface to make it clear that we don't run QBFT (it will also stop us from ever accidentally running qbft).

PR #277, opened 2023-06-21
refactor: skip `QBFTController` assertion in `ValidatorRegistrationRunner.StartNewDuty`

moshe-blox:
This PR tackles #267 with a different approach, instead of passing a dummy `QBFTController` to `ValidatorRegistrationRunner`, it overrides the default behavior of `StartNewDuty` to not use the `QBFTController` at all.


ssv/validator_registration.go
```diff
@@ -41,7 +42,13 @@ func NewValidatorRegistrationRunner(
 }
 
 func (r *ValidatorRegistrationRunner) StartNewDuty(duty *types.Duty) error {
-	return r.BaseRunner.baseStartNewDuty(r, duty)
+	// Note: Unlike the other runners, this doesn't call BaseRunner.baseStartNewDuty because
+	// that requires a QBFTController which ValidatorRegistrationRunner doesn't have.
+	if r.HasRunningDuty() {
+		return errors.New("already running duty")
+	}
```
GalRogozinski, line 49: Good addition per our conversation :-)

GalRogozinski (approved):

GalRogozinski:
Fixes #270

alonmuroch:
Why not just replace return b.QBFTController.CanStartInstance() with

if b.QBFTController != nil {
     return b.QBFTController.CanStartInstance()
}
return nil

GalRogozinski:
@moshe-blox 
I think the problem is currently resolved due to changes in #274 
can you validate and close the PR if that is so?

GalRogozinski:
Actually this PR is needed to resolve something else I am working on..
@alonmuroch I don't like your solution because it is error-prone. 

qbft instance shouldn't be nil for duties

GalRogozinski:
Will be merged as part of #295

PR #281, opened 2023-06-27
Additional SSV GetPostState implementations

GalRogozinski:
Add more `getPostState` implementations.
Generate accompanying jsons with `go run ssv/spectest/generate/main.go`

Also adds overrideStateComparisons

### DIFF
--- a/ssv/spectest/generate/main.go
+++ b/ssv/spectest/generate/main.go
@@ -102,7 +102,10 @@ func writeSingleSCJson(path string, testType string, post interface{}) {
 	basedir = filepath.Join(strings.TrimSuffix(basedir, "main.go"), "state_comparison", testType)
 	file := filepath.Join(basedir, fmt.Sprintf("%s.json", path))
 	// try to create directory if it doesn't exist
-	os.MkdirAll(filepath.Dir(file), 0700)
+	if err := os.MkdirAll(filepath.Dir(file), 0700); err != nil {
+		panic(err.Error())
+	}
+
 	if err := os.WriteFile(file, byts, 0644); err != nil {
 		panic(err.Error())
 	}
--- a/ssv/spectest/generate/main.go
+++ b/ssv/spectest/generate/main.go
@@ -100,11 +100,9 @@ func writeSingleSCJson(path string, testType string, post interface{}) {
 		panic("no caller info")
 	}
 	basedir = filepath.Join(strings.TrimSuffix(basedir, "main.go"), "state_comparison", testType)
-
-	// try to create directory if it doesn't exist
-	_ = os.Mkdir(basedir, 0700)
 	file := filepath.Join(basedir, fmt.Sprintf("%s.json", path))
-
+	// try to create directory if it doesn't exist
+	os.MkdirAll(filepath.Dir(file), 0700)
 	if err := os.WriteFile(file, byts, 0644); err != nil {
 		panic(err.Error())
 	}
--- a/ssv/spectest/generate/main.go
+++ b/ssv/spectest/generate/main.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/bloxapp/ssv-spec/ssv/spectest/tests"
+	"github.com/bloxapp/ssv-spec/types"
 	"log"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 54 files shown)


ssv/spectest/generate/main.go
```diff
@@ -72,12 +73,23 @@ func clearStateComparisonFolder() {
 }
 
 func writeJsonStateComparison(name, testType string, post interface{}) {
+	postMap, ok := post.(map[string]types.Root)
```
y0sher, line 76: why not pass it as `map[string]types.Root` to the function then ?

ssv/spectest/generate/main.go
```diff
@@ -72,12 +73,23 @@ func clearStateComparisonFolder() {
...
+
+	if ok {
+		for subTestName, postState := range postMap {
+			writeSingleSCJson(filepath.Join(name, subTestName), testType, postState)
+		}
+	} else {
+		writeSingleSCJson(name, testType, post)
+	}
```
y0sher, line 84:
```suggestion
	if !ok {
	    writeSingleSCJson(name, testType, post)
	    return
	}
        for subTestName, postState := range postMap {
	    writeSingleSCJson(filepath.Join(name, subTestName), testType, postState)
	}
```

ssv/spectest/generate/main.go
y0sher: ideally I'd split it to two functions and check before calling. go is meant to be statically typed. Generics introduced not long ago, I think we should either stick to using types or used generics.

types/testingutils/comparable/helpers.go
```diff
@@ -31,3 +35,21 @@ func FixIssue178(input *types.ConsensusData, version spec2.DataVersion) *types.C
 
 	return ret
 }
+
+// UnmarshalSSVStateComparison reads a json derived from 'test' and unmarshals it into 'targetState'
+func UnmarshalSSVStateComparison(testName string, folderName string, targetState types.Root) (types.Root, error) {
```
alonmuroch, line 40: why is this used by just one function? don't we have a bunch of those calls?

ssv/spectest/generate/main.go
GalRogozinski: This is more problematic than it first seems.
`post` is received from `GetPostState()` that returns an `interface{}`

`GetPostState()` is an interface function that sometimes returns a `map[string]types.Root`, and sometimes returns `types.Root` (it may also return a slice for now, but it should be deprecated).

Since the point is to not use `interface{}` at all I think to change `getPostState` to always return a `map`. Where we originally returned `Root` it will simply be a map size of 1.

What do you think?

ssv/spectest/generate/main.go
```diff
@@ -88,10 +100,11 @@ func writeJsonStateComparison(name, testType string, post interface{}) {
 		panic("no caller info")
 	}
 	basedir = filepath.Join(strings.TrimSuffix(basedir, "main.go"), "state_comparison", testType)
```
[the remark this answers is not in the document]
GalRogozinski: sure
even though it is a small change in code it is a huge change in the generated files and may cause confusion while we review #274.

So I will do this in another PR

types/testingutils/comparable/helpers.go
GalRogozinski: While adding missing `getPostState()` I was laying down infra for farther refactoring.
Didn't want to do it all now to not distract from #274

ssv/spectest/generate/main.go
GalRogozinski: Opened #283 to track refactoring changed.
I want to solve each issue in ints own PR to cause less disruptions and ensure quality

types/testingutils/comparable/helpers.go
GalRogozinski: Opened #283 to track refactoring changed.
I want to solve each issue in ints own PR to cause less disruptions and ensure quality

ssv/spectest/generate/main.go
y0sher: you could also return some interface or struct that could have both of them. staying with only map sounds ok aswell.

ssv/spectest/generate/main.go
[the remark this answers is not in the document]
y0sher: yeah we can do it in separate PR, though the node might need this to depend on ssv-spec after #274 , currently its not able to load these file. we might find some workaround but its better to get this done quickly.

y0sher (approved):

alonmuroch (approved):

PR #286, opened 2023-07-09
Fix SC dir names and files

GalRogozinski:
### Description

- [x] removes '*' and '.' from dir names
- [x] 'multi_msg_processing_spectest' should create multiple jsons


ssv/spectest/generate/main.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"github.com/bloxapp/ssv-spec/ssv/spectest/tests"
 	"github.com/bloxapp/ssv-spec/types"
+	comparable2 "github.com/bloxapp/ssv-spec/types/testingutils/comparable"
```
y0sher, line 8: comparable2 could be just comparable

qbft/spectest/generate/main.go
```diff
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/bloxapp/ssv-spec/qbft/spectest/tests"
+	comparable2 "github.com/bloxapp/ssv-spec/types/testingutils/comparable"
```
y0sher, line 7: comparable2 could be just comparable

y0sher (approved):

ssv/spectest/generate/main.go
GalRogozinski: The thing is that `comparable` is a built in type in go.
So that's a bit confusing and that's why the IDEA automatically gives it a different name

There are places in the code where we call it `typescomparable` (because it is in `types` lib).
I think in another PR we need to choose a good name to use thorughout the project.

Any suggestions?

qbft/spectest/generate/main.go
GalRogozinski: ditto

ssv/spectest/generate/main.go
y0sher: since it compares states I guess `state` could be used as name, then using a function from the pkg could look like `state.Comparison`, which is nice. otherwise or if state is used, "statecomp", "compare/comp" just to clash with the stdlib.

GalRogozinski: will do in other PR since it adds lots of changes adding to https://github.com/bloxapp/ssv-spec/issues/283

y0sher (approved):

PR #291, opened 2023-07-13
Json encoding

GalRogozinski:
### Description 

Ensures that `[N]byte` arrays will be marshaled to concise hex strings.
Also the impl will have to 

### Issue
We added custom types. Not sure if it looks so elegant.
I think `ssv` impl will have to use them as well which might be a little weird.
We now use casting to `go-eth2` types.


GalRogozinski:
closing for now

PR #295, opened 2023-07-25
Don't start lower height duty

### LINKED ISSUE #282: When receiving a duty that has a slot smaller or equal to the respective runner's height, don't start it.
### Description

When a duty comes in, it first tries to finish the precon phase (if needed) before entering the consensus phase.
If it stops immediately at the consensus phase because of the height then the node burnt resources for no reason.

Should also stop postconsensus.
This relates to bloxapp/SIPS#28

GalRogozinski:
Fixes #282 

Added to `baseStartNewDuty`  a `ShouldProcessDuty` call  that ensures we are not trying to run a duty of a previous height.


GalRogozinski:
To take this out of draft mode:

- [ ] merge #277 
- [ ] generate json

moshe-blox (approved):

alonmuroch (approved):

PR #296, opened 2023-07-26
Don't broadcast messages if you can't process them

### LINKED ISSUE #288: Stop all runner processing if an instance is stopped
### Description

If an instance is stopped, we should also stop/prevent:
- [ ] broadcasting messages
- [ ] postconsensus processing
- [ ] preconsensus processing

GalRogozinski:
fix #288

### DIFF
--- a/qbft/instance.go
+++ b/qbft/instance.go
@@ -78,6 +78,9 @@ func (i *Instance) Start(value []byte, height Height) {
 }
 
 func (i *Instance) Broadcast(msg *SignedMessage) error {
+	if !i.CanProcessMessages() {
+		return errors.New("instance stopped processing messages")
+	}
 	byts, err := msg.Encode()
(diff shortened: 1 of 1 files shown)


GalRogozinski:
Tests will come at a later PR

y0sher:
> Tests will come at a later PR

IMO if you planning to have tests its better to have them aside the code.

y0sher (approved):
Approving code but again IMO you should add at least one test.

GalRogozinski:
@y0sher 

When we did #274 we didn't add tests. Now we are closing some holes and another PR will concentrate all tests

GalRogozinski:
This doesn't have to be a part of the spec tat attempts to be minimalistic.
A node can choose to not broadcast a message and it won't be a fork.

PR #297, opened 2023-07-30
Stop instances tests

### LINKED ISSUE #287: Add missing tests for instance_decided_enformcement SIP
https://github.com/bloxapp/ssv-spec/pull/274/commits/1f7049ebf3a3cc49e0d3d63c63dd27317b8e5998

GalRogozinski:
Fixes #287 

Also adds smart marshaling `instance.forceStop` field so it will only serialize to JSON on relevant tests.
Adds smart marshaling `Height` field to `runInstanceData`


qbft/spectest/tests/startinstance/force_stop_previous.go
```diff
@@ -1,8 +0,0 @@
-package startinstance
-
-import "github.com/bloxapp/ssv-spec/qbft/spectest/tests"
-
-// ForceStopPrevious tests stoping previous running instances when starting a new instance
-func ForceStopPrevious() tests.SpecTest {
```
GalRogozinski, line 6: This is deleted because prev_decided and prev_not_decided cover this

MatheusFranco99 (approved):

y0sher (approved):

qbft/instance.go
```diff
@@ -217,3 +220,41 @@ func (i *Instance) Decode(data []byte) error {
 func (i *Instance) CanProcessMessages() bool {
 	return !i.forceStop && int(i.State.Round) < CutoffRound
 }
+
+/////////////////////// JSON Marshalling for Tests ///////////////////////
```
alonmuroch, line 224: if it's for testing only:
1) why do we need it? which test depends on it?
2) let's separate it to a dedicated file

alonmuroch (changes requested):

qbft/instance.go
GalRogozinski: 1. It only prints `forceStop` when the value is `true`. All tests that have been forcedStopped depend on it.
2. done

There are some test that have nothing in their state change but `forceStop` field. If we don't print it we lose coverage.

In general our jsons are too verbose and hard to review. I would like to switch to an approach that we only print fields that change their default value.

alonmuroch (approved):

PR #299, opened 2023-08-01
Remove backwards compatibility for state comparison

GalRogozinski:
Backwards compatibility makes amending the spec ridiculously slow.
So we remove it. 

When we see a test that still has a root in the json, we know it wasn't still properly vetted


MatheusFranco99 (approved):
Good change!

alonmuroch (approved):

PR #300, opened 2023-08-01
Remove future msg and syncing requirements

### LINKED ISSUE #280: Remove future msg handling
Post merge of https://github.com/bloxapp/SIPs/pull/28 we do not need to handle future messages anymore.
Remove:

- future message container 
- future_msg.go
- if c.isFutureMessage(msg) from processMessage
- c.UponFutureMsg

GalRogozinski:
Fixes #280 

Depends on #299

Also removes syncing


MatheusFranco99 (approved):

qbft/controller.go
```diff
@@ -82,10 +79,12 @@ func (c *Controller) ProcessMsg(msg *SignedMessage) (*SignedMessage, error) {
 	if IsDecidedMsg(c.Share, msg) {
 		return c.UponDecided(msg)
 	} else if c.isFutureMessage(msg) {
```
alonmuroch, line 81: why not remove the condition?

qbft/future_msg.go
```diff
@@ -5,19 +5,6 @@ import (
 	"github.com/pkg/errors"
 )
```
alonmuroch, line 7: why not remove the entire file?

qbft/decided.go
```diff
@@ -41,8 +41,6 @@ func (c *Controller) UponDecided(msg *SignedMessage) (*SignedMessage, error) {
 	}
 
 	if isFutureDecided {
-		// sync gap
-		c.GetConfig().GetNetwork().SyncDecidedByRange(types.MessageIDFromBytes(c.Identifier), c.Height, msg.Message.Height)
```
alonmuroch, line 45: did you remove SyncDecidedByRange from the interface?

alonmuroch (changes requested):

qbft/controller.go
GalRogozinski: The answer is given within the comment in the code:
```
 else if c.isFutureMessage(msg) {
		// No error is returned if msg is future because it is a valid case.
		// The implementation may choose to create optimizations for this case.
		return nil, nil
	}
```

If we delete this segment then the code will move to `c.UponExistingInstanceMsg(msg)` which will not be always true in case a future message does show up...
It can happen if the eth2 node is lagging.


There are 2 approaches we can take:
1. Implicit - simply don't write cases the spec won't handle and let the implementor figure out that the edge case is missing.
2. Explicit - remind the implementor that the spec doesn't cover some edge case.

I think Explicit is better.

qbft/future_msg.go
GalRogozinski: Currently the remaining function is needed so that logic in `msg_validation.go` won't error.

Once @MatheusFranco99 will declare that Msg Validation he was working on is completed, we can delete it in another PR.

qbft/decided.go
GalRogozinski: yes, checkout this commit:
https://github.com/bloxapp/ssv-spec/pull/300/commits/23718342351af66ce5c31054bfecf3d981dae4d4

qbft/controller.go
alonmuroch: I'm not totally convinced. 

We shouldn't create "optimization" hocks in spec for implementation. 

Why not return error saying it's a future message, controller can't process?

GalRogozinski: fixed

qbft/controller.go
```diff
@@ -82,10 +79,10 @@ func (c *Controller) ProcessMsg(msg *SignedMessage) (*SignedMessage, error) {
...
-		return c.UponFutureMsg(msg)
-	} else {
-		return c.UponExistingInstanceMsg(msg)
+		return nil, fmt.Errorf("future msg from height %d, could not process", msg.Message.Height)
```
alonmuroch, line 82: I tend to like less errors with variables in them.. up to you but lets use one format and not a mix

qbft/future_msg.go
alonmuroch: what msg_validation.go file?

qbft/controller.go
GalRogozinski: done

qbft/future_msg.go
GalRogozinski: https://github.com/bloxapp/ssv-spec/blob/main/p2p/validation/msg_validation.go

alonmuroch: then lets move it there and remove the file... no need for non used code in spec

GalRogozinski: fixed

alonmuroch (approved):

moshe-blox (approved):

PR #305, opened 2023-08-20
drop blinded block rejection

moshe-blox:
This PR drops the conditional rejection of blinded blocks proposals together with the `SupportsBlinded` parameter which would no longer be needed.

Quoting from a conversation i had with @MatheusFranco99:
> this PR just removes a node configuration that rejects blinded block proposals
> it was a consensus configuration you could say
> the node would pass this boolean to the spec, according to it's configuration
> and the idea was that if your node is configured with BlindedProposals: false then you would not only propose non-blinded blocks only, but also reject others' blinded blocks as well
> in the spec there's still this boolean left: https://github.com/bloxapp/ssv-spec/blob/55bfe7a89a75ae728347e55cafb856171ff13662/ssv/proposer.go#L85 
> which decides whether you're gonna produce blinded or non-blinded blocks when you're the leader 
> this PR should not affect this, and should only remove the option to reject blinded blocks from others when you're not the leader (SupportsBlinded)


alonmuroch:
How do we test hybrid cluster with say 2 supporting blinded blocks and 2 not?

MatheusFranco99:
Regarding the test with different nodes (2 supporting blinded blocks and 2 not supporting), since the test design allows us only to test the functionality from the perspective of one node, my approach was to see if:
- A **normal** runner would accept a **blinded** block and complete the full proposal with a happy flow
- A **blinded** runner would accept a **normal** block and complete the full proposal with a happy flow
I think it's enough since the node can be abstracted from whether others support blinded blocks or not, and the critical point is whether it accepts blinded blocks or not. What do you think?

GalRogozinski (approved):

y0sher (approved):

ssv/spectest/tests/runner/full_happy_flow.go
```diff
@@ -205,8 +231,34 @@ func FullHappyFlow() tests.SpecTest {
 		}
 	}
 
+	// proposerBlindedReceivingNormalBlockV creates a test specification for versioned blinded proposer receiving a normal block.
```
alonmuroch, line 234: I would separate it into a separate test, testing specific cases in which some of the cluster members propose blinded blocks and others don't

MatheusFranco99: I took it out of the happy flow test and put it in a file of its own in runner/proposer. What do you think?

GalRogozinski (changes requested):
Run `make generate-jsons`

types/spectest/generate/tests.json
```diff
@@ -1 +1 @@
...J+EOfa27qiBRTzFjvNEM4/9BHg86cKMLjc4EqZRV2Q/Z57ZmrNfQCTm7jGHflNrA8OGFvzMmTNlFD3a/845yAWzkWdAcOmTxkZBVgQ6hCZoYMdp3iGLtfVpj0x6m3QUJTXLzAjVs/dHy/an37tsewzMWK9IhrxEFVhJbpyE2AZgEXd38BzbhMCi07Dy2Kbqz/Wg5VvRvmOHw3k66PUjHFlpeukUiUpJs64To6Eky/bOozt+tXXPbNE+YHPd9NAzo/hzZsJwia/ANwfvmkTIKDiHM9cPCbwI3ldNB88ZPwomoIzoJTv9sm1TBmMgEsba4ZR4PesxtyrDW71X8H8WZ3RthfXVshMriF+MsCBpSXgPlAYwc5bcfs55OCJfxaVcZe02CMJ/ZoupsIh1l54kllXRWkw9d7FDOrJvVtgOLKkXjYhQHuaCoU0HsXY16ZaNotmKqetKWuY55C7tDHc1MT0zCP0HlYnVyy3Tgakpj2cyWXam3NPjj4NuS8MQaIX2NIxXyQvq3Xy8oXVwuQEwqq3H6w7Sv7l8XIp5HQtrJzbUOw9ESQlEnm6jK6cG3f4oxAeoXoLj8iBkaW0v17N7M+AeABatkQR/lfDPXF1qvI/XaYxBVcKQoaHbhC535lbGmti8Bs6p4A2tvsiGymSvX+MEXryMVJvyPnHGNPAtqLJQYYwWmlTkr0XXmfLeZ0e94BOVoT8aYF6b5ZztBKUed+0HMPNCBXEWTqUkemcLli6/a0U2YHSMpQIAAAAAAAAAAwAAAAAAAAAAAQIDBAUGBwgJCgsMDQ4PEBESEwAAZKeztuANAgAAAAAAAAC4m+vGmXaXJqMYyOmXG9MXEpfGGupKZXinpPlLVH3LpbrBaokQi2tqH+NpXRqHSgsAAQIDBAUGBwgJCgsMDQ4PEBESEwABAgMEBQYHCAkKCwwNDg8QERITFBUWFxgZGhscHR4fICEiIyQlJicoKSorLC0uLzAxMjM0NTY3ODk6Ozw9Pj9AQUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVpbXF1eXw==","ExpectedRoot":[214,6,133,147,123,255,148,105,121,81,22,60,70,207,145,67,235,242,216,185,1,10,169,67,203,52,170,87,177,181,68,232],"ExpectedError":""}}
```
GalRogozinski: I think nothing in this file changed besides the key? so you can revert this change

MatheusFranco99: Reverted 👍

PR #310, opened 2023-09-13
Spec - Fork support

MatheusFranco99:
Updated according to https://github.com/bloxapp/SIPs/pull/26


GalRogozinski (approved):
💯

GalRogozinski:
@lior-blox @olegshmuelov and @y0sher 

Notice that in the spec's current implementation `Share` is part of `Validator`.
Since there are no tests for `Validator` you can still be spec compliant if the implementation's validator will be different than what you see in spec.

alonmuroch (approved):

types/domain_type.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+	"github.com/pkg/errors"
+)
+
+// NetworkID are intended to separate different SSV networks. A network can have many forks in it.
+type NetworkID [1]byte
+
+var (
+	MainnetNetworkID = NetworkID{0x0}
```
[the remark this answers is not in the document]
GalRogozinski: When we do spec rethink we can move it all to config

qbft/controller.go
```diff
@@ -31,7 +30,6 @@ func NewController(
 	return &Controller{
```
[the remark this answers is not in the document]
MatheusFranco99: Indeed. Forgot about that. Thanks! 😅

types/domain_type.go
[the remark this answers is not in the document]
y0sher: IMO this is something that shouldn't be changed often hence it can be part of the code, as a hard-coded config, this is how it's done at least in other clients. though it introduces an issue of being less dynamic like creating a specific network config for a on-demand spun-up test-network, but I assume we can solve it with adding a test network id.

types/domain_type.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	return current_fork, nil
+}
+
+func (d DomainType) GetNetworkID() NetworkID {
+	return NetworkID{d[2]}
+}
+
+func (f ForkData) GetRoot() ([]byte, error) {
+	byts, err := json.Marshal(f)
```
y0sher, line 90: consider using ssz here, as we use it mostly around the project, IMO we should stick to one deterministic serialization format, for example imagine a different language's json marshal will results in different order or some char (like ' instead of "), it will results in different hash.

qbft/controller.go
```diff
@@ -211,3 +208,7 @@ func (c *Controller) broadcastDecided(aggregatedCommit *SignedMessage) error {
 func (c *Controller) GetConfig() IConfig {
 	return c.config
 }
+
+func (c *Controller) SetIdentifier(identifier []byte) {
+	c.Identifier = identifier
```
y0sher, line 213: I can't understand how is this related to this PR

y0sher (changes requested):
left some comments

qbft/controller.go
GalRogozinski: Maybe if it isn't used in spec it can be deleted.
Implementation can add a simple setter as needed even if it isn't in spec

types/domain_type.go
GalRogozinski: I just noticed we use `json` in at least one other place and opened issue #312.

@MatheusFranco99 please fix for this PR and if you see more places where `json` is used for getting roots please add to the issue.

types/domain_type.go
```diff
@@ -0,0 +1,113 @@
...
+	case JatoV2NetworkID:
+		return []*ForkData{{Epoch: 0, Domain: JatoV2Testnet}}
+	default:
+		return []*ForkData{}
+	}
+}
+
+func (n NetworkID) DefaultFork() *ForkData {
+	return n.GetForksData()[0]
+}
```
olegshmuelov, line 63: In [SIP discussion](https://github.com/bloxapp/SIPs/pull/34#discussion_r1308932572), it's mentioned that the list of forks should be in order. The DefaultFork() method also assumes this. It’s a good idea to make sure this is always the case to avoid any problems later on.

I suggest adding a comment in the `func (n NetworkID) GetForksData() []*ForkData` method to mention that the network fork list should be ordered. It might also be helpful to have a test that checks the forks are listed in order.

also, I'm trying to understand the purpose of the DefaultFork() function. Is it supposed to return the current fork we're on, or is it meant to return the original genesis fork?

types/domain_type.go
```diff
@@ -0,0 +1,113 @@
...
+type DomainType [4]byte
+
+// DomainTypes represent specific forks for specific chains, messages are signed with the domain type making 2 messages from different domains incompatible
+var (
+	GenesisMainnet = DomainType{0x0, 0x0, MainnetNetworkID.Byte(), 0x0}
+	PrimusTestnet  = DomainType{0x0, 0x0, PrimusNetworkID.Byte(), 0x0}
+	ShifuTestnet   = DomainType{0x0, 0x0, ShifuNetworkID.Byte(), 0x0}
+	ShifuV2Testnet = DomainType{0x0, 0x0, ShifuNetworkID.Byte(), 0x1}
+	JatoTestnet    = DomainType{0x0, 0x0, JatoNetworkID.Byte(), 0x1}
+	JatoV2Testnet  = DomainType{0x0, 0x0, JatoV2NetworkID.Byte(), 0x1}
```
olegshmuelov, line 31: There is an inconsistency in the fork version values assigned to `JatoTestnet` and `JatoV2Testnet`. Specifically, both have been assigned a fork version value of 0x1 instead of a sequential versioning like 0x0 and 0x1. I understand that this inconsistency emerged from an error when these networks were first introduced.

```go
JatoTestnet    = DomainType{0x0, 0x0, byte(JatoNetworkID), 0x1}
JatoV2Testnet  = DomainType{0x0, 0x0, byte(JatoV2NetworkID), 0x1}
```
To prevent confusion for individuals who might be reviewing or working with this code in the future, I suggest adding a comment in the code explaining this historical mistake. This comment could explain why JatoTestnet and JatoV2Testnet both have a fork version value of 0x1, and the historical reason behind it.

Suggested comment:
```go
// Historical Note:
// The fork version values for JatoTestnet and JatoV2Testnet are both set to 0x1 due to an error
// when these networks were initially introduced. This inconsistency does not align with the sequential
// versioning observed in other network and fork definitions. It's retained to maintain historical accuracy
// and to avoid any unforeseen issues that might arise from changing these established values.
// Future references and modifications should acknowledge this historical inconsistency.
var (
	GenesisMainnet = DomainType{0x0, 0x0, byte(MainnetNetworkID), 0x0}
	PrimusTestnet  = DomainType{0x0, 0x0, byte(PrimusNetworkID), 0x0}
	ShifuTestnet   = DomainType{0x0, 0x0, byte(ShifuNetworkID), 0x0}
	ShifuV2Testnet = DomainType{0x0, 0x0, byte(ShifuNetworkID), 0x1}
	JatoTestnet    = DomainType{0x0, 0x0, byte(JatoNetworkID), 0x1} // Note the fork version value
	JatoV2Testnet  = DomainType{0x0, 0x0, byte(JatoV2NetworkID), 0x1} // Note the fork version value
)
```

olegshmuelov (changes requested):

types/domain_type.go
MatheusFranco99: I agree. Adding the comment. Thanks!

types/domain_type.go
MatheusFranco99: I'll add it. Thanks :)
Regarding the `DefaultFork` function, initially, it was being used for initialization because the Runner was holding fork info from the network. But this was dropped, and, also, since we want to remove as many functionalities from the spec as possible, I'll remove this function.
Thanks for noticing that.

olegshmuelov (approved):

olegshmuelov (approved):

types/generate.go
```diff
@@ -17,3 +17,6 @@ package types
 
 //go:generate rm -f ./consensus_data_encoding.go
 //go:generate go run github.com/ferranbt/fastssz/sszgen --path consensus_data.go --include ./operator.go,./signer.go,./partial_sig_message.go,./beacon_types.go,$GOPATH/pkg/mod/github.com/attestantio/go-eth2-client@v0.16.3/spec/phase0,$GOPATH/pkg/mod/github.com/attestantio/go-eth2-client@v0.16.3/spec,$GOPATH/pkg/mod/github.com/attestantio/go-eth2-client@v0.16.3/spec/altair --exclude-objs Contributions,BeaconNetwork,BeaconRole
+
+//go:generate rm -f ./domain_type_encoding.go
+//go:generate go run github.com/ferranbt/fastssz/sszgen --path domain_type.go --include $GOPATH/pkg/mod/github.com/attestantio/go-eth2-client@v0.16.3/spec/phase0 --exclude-objs NetworkID,DomainType
```
GalRogozinski, line 22: why NetworkID and DomainType excluded?

types/spectest/tests/forkdata/encoding.go
```diff
@@ -0,0 +1,25 @@
+package forkdata
+
+import (
+	"github.com/bloxapp/ssv-spec/types/testingutils"
+)
+
+// Encoding tests encoding of a ssv message
```
GalRogozinski, line 7: bad comment

types/domain_type.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (f ForkData) GetRoot() ([]byte, error) {
+	byts, err := f.MarshalSSZ()
+	if err != nil {
+		return nil, errors.Wrap(err, "could not marshal ForkData")
+	}
+	ret := sha256.Sum256(byts)
+	return ret[:], nil
+}
```
GalRogozinski, line 98: excuse my silliness because I am reviewing at a late hour.
Why do we need this method again? We ususally need this for signing.

When should we sign the ForkData?

GalRogozinski (changes requested):

GalRogozinski (approved):

y0sher (approved):

MatheusFranco99:
@y0sher 
I think the code in the spec is intended to be implementation-blinded, providing some flexibility on implementation decisions. But I agree that we should have discussed it more. Both the SIP and this spec PR underwent many changes because there was no common ground on how it should be implemented.

Regarding your question, my opinion is that the decision on the `DomainType` value should not be in the broadcast. It's determined by the `duty.Slot`'s epoch and, thus, a message for a duty of epoch $e$, even if broadcasted during $e+1$ which may have a fork, contains the `DomainType` associated with $e$.

I think the easiest way to implement this is, once a `Runner` starts a new duty, it checks if the `DomainType` changed and, if so, it updates the QBFT `controller` and `instance` identifiers.

What do you think about this approach?

PR #315, opened 2023-10-19
fix: (`GetSCDir`) don't alter base directory

moshe-blox:


moshe-blox:
Closed in favor of #313

PR #317, opened 2023-11-01
Spec - New ValidatorRegistration/ValidatorExit duties stop old ones

GalRogozinski:
# Description

Stalling ValidatorRegistration/ValidatorExit duties prevented the execution of new duties. Now when a duty that fits a higher slot comes in it stops the old one.

# Affected tests

 `post_wrong_decided` - Deleted. Seems to be a duplicate of `post_future_decided`. Unclear what "wrong" means in the context.
 
 `post_future_decided` - renamed to `old_slot_duty`. Now that we don't support future messages this is a more fitting name. Cases for missing duties added.

Added missing cases to:
`consensus_not_started`, `duplicate_duty_finished`, `duplicate_duty_not_finished`, and `finished`.


MatheusFranco99 (approved):
LGTM

moshe-blox (approved):

ssv/runner.go
```diff
@@ -234,9 +234,10 @@ func (b *BaseRunner) hasRunningDuty() bool {
 }
 
 func (b *BaseRunner) ShouldProcessDuty(duty *types.Duty) error {
-	if b.QBFTController.Height >= qbft.Height(duty.Slot) && b.QBFTController.Height != 0 {
+	// assume StartingDuty is not nil if state is not nil
+	if b.State != nil && b.State.StartingDuty.Slot >= duty.Slot {
```
alonmuroch, line 238: Why not just add b.QBFTController != nill check? 
Why change it to the above which changes the whole logic

ssv/validator_registration.go
```diff
@@ -42,13 +42,7 @@ func NewValidatorRegistrationRunner(
...
 
 func (r *ValidatorRegistrationRunner) StartNewDuty(duty *types.Duty) error {
-	// Note: Unlike the other runners, this doesn't call BaseRunner.baseStartNewDuty because
-	// that requires a QBFTController which ValidatorRegistrationRunner doesn't have.
-	if r.HasRunningDuty() {
-		return errors.New("already running duty")
-	}
-	r.BaseRunner.baseSetupForNewDuty(duty)
-	return r.executeDuty(duty)
+	return r.BaseRunner.baseStartNewDuty(r, duty)
```
alonmuroch, line 45: Why is r.executeDuty removed?

alonmuroch (changes requested):

ssv/runner.go
GalRogozinski: While fixing the issue I thought about making things better.

1. In general each validator duty runner can run only one duty at a time.
2. There is no real reason to make rule (1) dependent on QBFT controller. If we do separation of concerns and processes it is unclear why the runner has to even have access to the internal state of QBFT controller.
3. With my current change all duties (including the ones that aren't beacon duties) follow the exact same pattern. Because of the special exemptions we did to duties with no QBFT we ran into this issue.

ssv/validator_registration.go
GalRogozinski: `baseStartNewDuty` calls it

ssv/runner.go
GalRogozinski: added a fallback that uses QBFTController just to get this approved

MatheusFranco99: From my perspective, indeed, the change makes module separation better and avoids the problem with the validator registration and exit duties

GalRogozinski:
Replaced by #318

PR #321, opened 2023-11-20
Fix double proposal

GalRogozinski:
# Description

Makes sure that any new round change message that is processes after RC quorum is reached won't trigger sending a proposal


GalRogozinski:
There are several ways to solve this:

1. Check it the PROPOSAL was committed the way it was done in PREPARE with COMMITs. 
https://github.com/bloxapp/ssv-spec/blob/2af3a73b3fe7a8f44ec154b71c67ca407e5d68bb/qbft/prepare.go#L28-L30.
But this checks inside the container... There could be a scenario where the COMMIT was broadcasted but another PREPARE was processed, resulting in 2 commits.
2. Flag once a proposal was broadcasted for a certain round and height. This will cover all scenarios but will farther complicate the state.
3. *The chosen solution*: ensure that only upon the first detection of RC quorum PROPOSAL will be broadcasted.

ssv/spectest/generate/state_comparison/tests_MsgProcessingSpecTest/qbft message processing valid justification.json
```diff
@@ -0,0 +1,283 @@
+{
```
GalRogozinski, line 1: This file existed before

qbft/spectest/tests/roundchange/valid_justification_late.go
```diff
@@ -0,0 +1,42 @@
+package roundchange
+
+import (
+	"github.com/bloxapp/ssv-spec/qbft"
+	"github.com/bloxapp/ssv-spec/qbft/spectest/tests"
+	"github.com/bloxapp/ssv-spec/types"
+	"github.com/bloxapp/ssv-spec/types/testingutils"
+)
+
+// ValidJustificationLate tests a valid rc quorum justification
```
alonmuroch, line 10: explanation is not complete, please write better detail

alonmuroch (changes requested):

y0sher (approved):

qbft/spectest/tests/roundchange/valid_justification_late.go
GalRogozinski: :+1:

alonmuroch (approved):

ssv/spectest/generate/state_comparison/tests_MsgProcessingSpecTest/qbft message processing valid justification.json
nkryuchkov: @GalRogozinski 
This file was created under `ssv` directory, so the `qbft/.../qbft message processing valid justification.json` was unchanged. Did you intend to make changes to the `qbft/.../qbft message processing valid justification.json` and not to create this one (`ssv/.../qbft message processing valid justification.json`)?

Issue #219, opened 2023-05-08
Multi proposer spec test support for all (upcoming) Beacon forks


olegshmuelov:
support version beacon block for testing

PR #324, opened 2023-12-20
Deneb Support

MatheusFranco99:
## Major changes
- Added logical support to Deneb's beacon block in consensus data.
- Note: When submitting a proposer duty output (`VersionedProposal` or `VersionedBlindedProposal`) to the Beacon node, Deneb  uses:
    - `VersionedBlindedProposal`: a `BlindedBeaconBlock`, similarly to other forks.
    - `VersionedProposal`: `BlockContents`, differently from older forks which used `BeaconBlocks`. Nonetheless, the validator signature is still over the beacon block (`BlockContents.Block`), i.e. the signing root is kept the same.

## Minor changes
- Updated Attestantio lib from 0.16.3 to [0.19.7](https://github.com/attestantio/go-eth2-client/tree/v0.19.7). Purpose: use up-to-date Deneb spec types.
- Updated Fastssz from 0.1.2 to [0.1.3](https://github.com/ferranbt/fastssz/tree/v0.1.3) as a consequence of the Attestantio update.
- Deneb testing utils. To-Do: define Deneb fork epoch.
- Added type tests.
- Remove all Bellatrix objects and use Deneb.
- Update ssz-max sizes for `ConsensusData` and `FullData` fields, since Deneb uses a `BlockContents` object which contains a Beacon block, KZGProofs (288 bytes) and Blobs of 786432 bytes.

### SSZ generation notes
- Fastssz 0.1.3 is now deterministic. Because of this, when you run `make generate-ssz` you may see new logs `INFO: Skipped ssz generated object:`.
- Fastssz 0.1.3 seems to have an error which occurs in the `consensus_data_encoding.go` file. It imports an unused library. For that, we created an [issue](https://github.com/ferranbt/fastssz/issues/136) in Fastssz which remains unsolved. As a temporary arrangement, we added, in the `generate.go` in `./types`, a command to remove the unused import.

### Attestantio update notes
- The default `DataVersion` used to be `phase0`. This was was changed to become `unknown`. Therefore, some testing utils that didn't define the `DataVersion` in the `ConsensusData` had to be updated.

### Leftovers To-Do
- [x] Set the correct Deneb fork epoch.

#### References
- [Deneb Consensus-Spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/beacon-chain.md)


types/ssz.go
```diff
@@ -152,3 +153,30 @@ func (b SSZ32Bytes) MarshalSSZ() ([]byte, error) {
...
+		return err
+	}
+	subIndx := hh.Index()
+	for _, i := range b {
+		hh.PutBytes(i[:])
+	}
+	numItems := uint64(len(b))
+	hh.MerkleizeWithMixin(subIndx, numItems, 4096)
+	return nil
+}
```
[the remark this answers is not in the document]
olegshmuelov: Please see previous discussion
https://github.com/bloxapp/ssv-spec/pull/234#discussion_r1195452213

types/spectest/tests/ssz/capella_marshaling_validation.go
```diff
@@ -0,0 +1,22 @@
...
+
+// CapellaBeaconBlockMarshalling tests a valid capella marshaling
+func CapellaBeaconBlockMarshalling() *SSZSpecTest {
+
+	root, err := testingutils.TestingBeaconBlockCapella.HashTreeRoot()
+	if err != nil {
+		panic(err.Error())
+	}
+
+	return &SSZSpecTest{
```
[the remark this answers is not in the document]
olegshmuelov: https://github.com/bloxapp/ssv-spec/pull/324#discussion_r1435865363

go.mod
```diff
@@ -47,8 +47,8 @@ require (
 	go.uber.org/atomic v1.10.0 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
 	go.uber.org/zap v1.24.0 // indirect
-	golang.org/x/crypto v0.7.0 // indirect
-	golang.org/x/sys v0.7.0 // indirect
+	golang.org/x/crypto v0.10.0 // indirect
```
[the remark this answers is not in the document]
MatheusFranco99: That's true. We shall do performance tests when included. Just to remember, for the RSA case, we're actually using microsoft's openssl so it's less of an issue.

GalRogozinski:
There seems to be some epoch [activation dates here](
https://weekinethereumnews.com/#:~:text=Testnet%20upgrade%20dates%3A%20Goerli%20Jan%2017%20(epoch%20231680)%2C%20then%20assuming%20no%20issues%2C%20single%20client%20releases%20for%20Sepolia%20Jan%2030%20(epoch%20132608)%20%26%20Hole%C5%A1ky%20Feb%207%20(epoch%2029696))

Not sure if it is final but maybe let's put this in and later change it so we can all approve this PR

MatheusFranco99:
@GalRogozinski  If I update it, I start getting:
> panic: can't start new duty runner instance for duty: input data invalid: duty invalid: duty epoch is into far future

because of the dutyValueCheck.

types/spectest/tests/consensusdata/proposer/versioned_block_spectest.go
```diff
@@ -56,14 +56,14 @@ func (test *ProposerSpecTest) Run(t *testing.T) {
...
 		switch vBlk.Version {
-		case spec.DataVersionBellatrix:
-			require.NotNil(t, vBlk.Bellatrix)
-			blkSSZ, err = vBlk.Bellatrix.MarshalSSZ()
-			require.NoError(t, err)
 		case spec.DataVersionCapella:
 			require.NotNil(t, vBlk.Capella)
 			blkSSZ, err = vBlk.Capella.MarshalSSZ()
 			require.NoError(t, err)
+		case spec.DataVersionDeneb:
```
GalRogozinski, line 63: if you scroll up you see a comment for issue 219. IMO you can delete the comment and close the issue

GalRogozinski:
Closes #219

types/testingutils/beacon_node.go
```diff
@@ -101,12 +103,12 @@ var Transactions = func() []bellatrix.Transaction {
 	return res.Transactions
 }()
 
-var TestingBeaconBlock = &bellatrix.BeaconBlock{
+var TestingBeaconBlock = &deneb.BeaconBlock{
```
olegshmuelov, line 106: This section of code appears to be unused now. It can be safely deleted as we have transitioned to using the `go-eth2-client` blocks, as specified in our `beacon_node_consts.go`. For reference, see [beacon_node_consts](https://github.com/bloxapp/ssv-spec/blob/3d553a7de7b6ddeaf46bcc3a6d86258e20ae2730/types/testingutils/beacon_node_consts.go).

types/testingutils/beacon_node.go
```diff
@@ -129,33 +131,33 @@ var TestingBeaconBlock = &bellatrix.BeaconBlock{
...
-			BaseFeePerGas: [32]byte{},
+			BaseFeePerGas: &uint256.Int{37},
 			BlockHash:     phase0.Hash32{},
 			Transactions:  Transactions,
 		},
 	},
 }
 
-var TestingBlindedBeaconBlock = func() *apiv1bellatrix.BlindedBeaconBlock {
+var TestingBlindedBeaconBlock = func() *apiv1deneb.BlindedBeaconBlock {
```
olegshmuelov, line 152: This section of code appears to be unused now. It can be safely deleted as we have transitioned to using the `go-eth2-client` blocks, as specified in our `beacon_node_consts.go`. For reference, see [beacon_node_consts](https://github.com/bloxapp/ssv-spec/blob/3d553a7de7b6ddeaf46bcc3a6d86258e20ae2730/types/testingutils/beacon_node_consts.go).

types/testingutils/beacon_node.go
```diff
@@ -129,33 +131,33 @@ var TestingBeaconBlock = &bellatrix.BeaconBlock{
...
 			LogsBloom:     [256]byte{},
 			PrevRandao:    [32]byte{},
 			BlockNumber:   100,
 			GasLimit:      1000000,
 			GasUsed:       800000,
 			Timestamp:     123456789,
-			BaseFeePerGas: [32]byte{},
+			BaseFeePerGas: &uint256.Int{37},
 			BlockHash:     phase0.Hash32{},
 			Transactions:  Transactions,
```
olegshmuelov, line 147: I suggest removing the `transactions` and `withdrawals` structures, along with their related components, including the custom SSZ logic and associated unnecessary tests. This simplification is feasible if we adopt the `go-eth-client` blinded block in a manner similar to how we currently use the regular block in our [`beacon_node_consts.go`](https://github.com/bloxapp/ssv-spec/blob/3d553a7de7b6ddeaf46bcc3a6d86258e20ae2730/types/testingutils/beacon_node_consts.go).

This approach is supported by our unit test which verifies that the regular block root matches the blinded block root. This can be reviewed in [`beacon_node_test.go`](https://github.com/bloxapp/ssv-spec/blob/3d553a7de7b6ddeaf46bcc3a6d86258e20ae2730/types/testingutils/beacon_node_test.go).

By adopting it, we can eliminate a significant amount of unnecessary code found in [`ssz.go`](https://github.com/bloxapp/ssv-spec/blob/3d553a7de7b6ddeaf46bcc3a6d86258e20ae2730/types/ssz.go), specifically within the `SSZTransactions` and `SSZWithdrawals` sections.
Additionally, this change allows us to remove unnecessary [SSZ tests](https://github.com/bloxapp/ssv-spec/tree/3d553a7de7b6ddeaf46bcc3a6d86258e20ae2730/types/spectest/tests/ssz).

olegshmuelov (changes requested):

types/testingutils/beacon_node.go
MatheusFranco99: That will be good! Thanks!
But I think it would be better to make it in another PR. What do you think? @olegshmuelov @GalRogozinski

olegshmuelov: > That will be good! Thanks! But I think it would be better to make it in another PR. What do you think? @olegshmuelov @GalRogozinski

Sure.
Btw the [transactions](https://github.com/bloxapp/ssv-spec/blob/49e2bdd9f013276c93355e5908506b63df79d1d9/types/testingutils/beacon_node.go#L95) func can be removed in the current PR as well

types/spectest/tests/consensusdata/deneb_blinded_block_validation.go
```diff
@@ -0,0 +1,6 @@
+package consensusdata
+
+// DenebBlindedBlockValidation tests a valid consensus data with deneb blinded block
+func DenebBlindedBlockValidation() *SpecTest {
```
y0sher, line 4: do we want to merge without all these tests?

y0sher (approved):

types/testingutils/beacon_node_versioned.go
```diff
@@ -5,46 +5,50 @@ import (
...
 	"github.com/attestantio/go-eth2-client/spec/phase0"
 	ssz "github.com/ferranbt/fastssz"
 
 	"github.com/bloxapp/ssv-spec/types"
 )
 
 const (
+
+	//TO-DO: Define Deneb Fork Epoch
+	ForkEpochPraterDeneb = 225431
```
moshe-blox, line 21: I'm not sure what's the significance of that, but that fork epoch is wrong, I think this is the right one:
https://github.com/ethereum/execution-specs/blob/559ba54829a5d726ca95fe5701486f7bd303bf7f/network-upgrades/mainnet-upgrades/cancun.md

types/testingutils/beacon_node.go
MatheusFranco99: Done

olegshmuelov (approved):

types/spectest/tests/consensusdata/deneb_blinded_block_validation.go
MatheusFranco99: `ConsensusData` has no implemented tests yet, only for versioned blocks inside the `proposer` folder, which were altered to use Deneb instead of Capella.
However, there's the PR https://github.com/bloxapp/ssv-spec/pull/344 for that (`ConsensusData` tests).
We can also merge it before the spec release.
What do you think?

ssv/spectest/tests/runner/preconsensus/post_decided.go
```diff
@@ -21,7 +21,10 @@ func PostDecided() tests.SpecTest {
 	decideRunner := func(r ssv.Runner, duty *types.Duty, decidedValue *types.ConsensusData, preMsgs []*types.SignedPartialSignatureMessage) ssv.Runner {
 		r.GetBaseRunner().State = ssv.NewRunnerState(3, duty)
 		for _, msg := range preMsgs {
-			r.ProcessPreConsensus(msg)
+			err := r.ProcessPreConsensus(msg)
+			if err != nil {
+				panic(err.Error())
```
GalRogozinski, line 26: should we panic or handle the error?
Do we panic in other places?

types/testingutils/beacon_node_versioned.go
GalRogozinski: @MatheusFranco99 
This may be a very bad bug in our value check... not sure
are you debigging this?
do you want me to look at this with you?

ssv/spectest/tests/runner/preconsensus/post_decided.go
MatheusFranco99: I decided to use `panic` because this is just for test generation. If there's an error, the generation will fail. Also, `panic` will facilitate to see the error and where it happened (since we weren't catching it before).

types/testingutils/beacon_node_versioned.go
MatheusFranco99: I'm looking into it.
But, basically it's triggered by [the first duty value check](https://github.com/bloxapp/ssv-spec/blob/main/ssv/value_check.go#L20-L21) which uses [some network functions](https://github.com/bloxapp/ssv-spec/blob/main/types/beacon_types.go#L116-L199).

types/spectest/tests/consensusdata/deneb_blinded_block_validation.go
y0sher: I approved #334 so we can merge it , but then we'll have to impl these tests?  we want to get this merged and release asap

MatheusFranco99: Yes, after merging, we'll need to implement the missing tests for Deneb but it shall be fast to do

types/consensus_data.go
```diff
@@ -185,7 +185,7 @@ func (ci *ConsensusData) GetAttestationData() (*phase0.AttestationData, error) {
 }
 
 // GetBlockData ISSUE 221: GetBlockData/GetBlindedBlockData return versioned block only
-func (ci *ConsensusData) GetBlockData() (*api.VersionedProposal, ssz.HashRoot, error) {
+func (ci *ConsensusData) GetBlockData() (blk *api.VersionedProposal, signingRoot ssz.HashRoot, err error) {
```
GalRogozinski, line 190: the impl depends on this in anyway or you just changed because you thought it is nice?

GalRogozinski (approved):

types/consensus_data.go
MatheusFranco99: Impl does not depend. We got confused by the function return values and decided to clarify it

types/consensus_data.go
```diff
@@ -127,7 +129,7 @@ type ConsensusData struct {
 	// Current 30M gas limit produces 30M / 16 (call data cost) = 1,875,000 bytes (https://eips.ethereum.org/EIPS/eip-4488)
 	// we can upper limit transactions to 2^21, together with the rest of the data 2*2^21 = 2^22 = 4,194,304 bytes
 	// Exmplanation on why transaction sizes are so big https://github.com/ethereum/consensus-specs/pull/2686
-	DataSSZ []byte `ssz-max:"4194304"` // 2^22
+	DataSSZ []byte `ssz-max:"4981024"` // 2^22 + 288 + 786432
```
olegshmuelov, line 132: There is a slight discrepancy in the `ssz-max` value for `DataSSZ`. The current `ssz-max` is set to `4,981,024` bytes, but based on a detailed breakdown of the components (including `KZG Proofs`, `Blobs`, basic `BeaconBlock` components, and the fixed-size elements in `BeaconBlockBody`), the calculated maximum size should be `4,981,464` bytes.

This difference of `440 bytes` seems to arise from not fully accounting for:
- The fixed-size components within `BeaconBlockBody` (`RANDAOReveal`, `ETH1Data`, `Graffiti`, `SyncAggregate`), which total `360 bytes`.
- The basic `BeaconBlock` size (components `Slot`, `ProposerIndex`, `ParentRoot`, `StateRoot`), adding up to `80 bytes`.

Additionally, we need to adjust the full-data `ssz-max` in SignedMessage accordingly

types/spectest/tests/consensusdata/deneb_blinded_block_validation.go
GalRogozinski: done

types/testingutils/beacon_node_versioned.go
GalRogozinski: So this can only be changed around the time the fork happens
The implementation shouldn't use this value

types/testingutils/ssv_msgs.go
```diff
@@ -398,7 +406,7 @@ var TestSelectionProofWithJustificationsConsensusData = func(ks *TestKeySet) *ty
 
 	return &types.ConsensusData{
 		Duty:                       TestingAggregatorDuty,
-		Version:                    spec.DataVersionBellatrix,
+		Version:                    spec.DataVersionDeneb,
```
olegshmuelov, line 409: IMO the version should be reverted to `DataVersionBellatrix`
The duty not related to proposal here

types/testingutils/ssv_msgs.go
```diff
@@ -801,7 +809,7 @@ var TestContributionProofWithJustificationsConsensusData = func(ks *TestKeySet)
 
 	return &types.ConsensusData{
 		Duty:                       TestingSyncCommitteeContributionDuty,
-		Version:                    spec.DataVersionBellatrix,
+		Version:                    spec.DataVersionDeneb,
```
olegshmuelov, line 812: IMO the version should be reverted to `DataVersionBellatrix`
The duty not related to proposal here

types/testingutils/ssv_msgs.go
GalRogozinski: it caused failures on ssv due to a bug that @moshe-blox just fixed
anyhow we shouldn't have Bellatrix in our code anymore so nothing wrong with this

types/consensus_data.go
MatheusFranco99: I've fixed some fields that we missed and I actually arrived at a bigger value: 5305388.
@GalRogozinski We would like a third opinion on that. Could you calculate your ssz-max value for DataSSZ (Deneb.BlockContents)?
If you'll push anything, pls don't forget to generate the ssz and extend the ssz-max values to FullData in SignedMessage

GalRogozinski: Looking into the calculation:
`Attestation(2048*8 + 96 + AttestationData(2*8 + 32 + 2*(8+32)))`

This is derived from:
```
type Attestation struct {
	AggregationBits bitfield.Bitlist `ssz-max:"2048"`
	Data            *AttestationData
	Signature       BLSSignature `ssz-size:"96"`
}
```
So not sure why we multiply 2048 by 8?
I am not even sure why `ssz-max:"2048"` is correct?
There should be max 2048 bits in the `Bitlist` because this is the max number of committee members. But ssz sizes are given in bytes everywhere so it should be 2048/8 = 256...

Probably the go-eth2-client lib is right and I am wrong but I don't understand why...

GalRogozinski: The following python code prints:
`Total Size without ExecutionPayload: 3207032`

That's slightly different than what Matheos calculated. This doesn't take my previous comment into account

```python
# Constants
KZG_PROOFS_SIZE = 6 * 48  # KZGProofs size
BLOBS_SIZE = 6 * 131072  # Blobs size
BEACON_BLOCK_OVERHEAD = 2 * 32 + 2 * 8  # Additional overhead for BeaconBlock

# Components of BeaconBlockBody
ETH1_DATA_SIZE = 96 + 2 * 32 + 8  # ETH1Data
PROPOSER_SLASHING_SIZE = 16 * (2 * (96 + 3 * 32 + 2 * 8))  # ProposerSlashing
ATTESTER_SLASHING_SIZE = 2 * (2 * (2048 * 8 + 96 + (2 * 8 + 32 + 2 * (8 + 32))))  # AttesterSlashing
ATTESTATION_SIZE = 128 * (2048 * 8 + 96 + (2 * 8 + 32 + 2 * (8 + 32)))  # Attestation
DEPOSIT_SIZE = 16 * (33 * 32 + 48 + 32 + 8 + 96)  # Deposit
SIGNED_VOLUNTARY_EXIT_SIZE = 16 * (96 + 2 * 8)  # SignedVoluntaryExit
SYNC_AGGREGATE_SIZE = 64 + 96  # SyncAggregate
BLS_TO_EXECUTION_CHANGES_SIZE = 16 * (96 + (8 + 48 + 20))  # BLSToExecutionChanges
KZG_COMMITMENT_SIZE = 4096 * 48  # KZGCommitment

# BeaconBlockBody total size without ExecutionPayload
beacon_block_body_size_without_execution_payload = (
    ETH1_DATA_SIZE + PROPOSER_SLASHING_SIZE + ATTESTER_SLASHING_SIZE + 
    ATTESTATION_SIZE + DEPOSIT_SIZE + SIGNED_VOLUNTARY_EXIT_SIZE + 
    SYNC_AGGREGATE_SIZE + BLS_TO_EXECUTION_CHANGES_SIZE + KZG_COMMITMENT_SIZE
)

# Total size of Deneb.BlockContents and BeaconBlock without ExecutionPayload
total_size_without_execution_payload = KZG_PROOFS_SIZE + BLOBS_SIZE + BEACON_BLOCK_OVERHEAD + beacon_block_body_size_without_execution_payload

# Output the total size
print("Total Size without ExecutionPayload:", total_size_without_execution_payload)
```

GalRogozinski: When I take the comment into account I get 1,078,136

```python
# Constants
KZG_PROOFS_SIZE = 6 * 48  # KZGProofs size
BLOBS_SIZE = 6 * 131072  # Blobs size
BEACON_BLOCK_OVERHEAD = 2 * 32 + 2 * 8  # Additional overhead for BeaconBlock

# Components of BeaconBlockBody
ETH1_DATA_SIZE = 96 + 2 * 32 + 8  # ETH1Data
PROPOSER_SLASHING_SIZE = 16 * (2 * (96 + 3 * 32 + 2 * 8))  # ProposerSlashing
ATTESTER_SLASHING_SIZE = 2 * (2 * (2048 / 8 + 96 + (2 * 8 + 32 + 2 * (8 + 32))))  # AttesterSlashing
ATTESTATION_SIZE = 128 * (2048/8 + 96 + (2 * 8 + 32 + 2 * (8 + 32)))  # Attestation
DEPOSIT_SIZE = 16 * (33 * 32 + 48 + 32 + 8 + 96)  # Deposit
SIGNED_VOLUNTARY_EXIT_SIZE = 16 * (96 + 2 * 8)  # SignedVoluntaryExit
SYNC_AGGREGATE_SIZE = 64 + 96  # SyncAggregate
BLS_TO_EXECUTION_CHANGES_SIZE = 16 * (96 + (8 + 48 + 20))  # BLSToExecutionChanges
KZG_COMMITMENT_SIZE = 4096 * 48  # KZGCommitment

# BeaconBlockBody total size without ExecutionPayload
beacon_block_body_size_without_execution_payload = (
    ETH1_DATA_SIZE + PROPOSER_SLASHING_SIZE + ATTESTER_SLASHING_SIZE + 
    ATTESTATION_SIZE + DEPOSIT_SIZE + SIGNED_VOLUNTARY_EXIT_SIZE + 
    SYNC_AGGREGATE_SIZE + BLS_TO_EXECUTION_CHANGES_SIZE + KZG_COMMITMENT_SIZE
)

# Total size of Deneb.BlockContents and BeaconBlock without ExecutionPayload
total_size_without_execution_payload = KZG_PROOFS_SIZE + BLOBS_SIZE + BEACON_BLOCK_OVERHEAD + beacon_block_body_size_without_execution_payload

# Output the total size
print("Total Size without ExecutionPayload:", total_size_without_execution_payload)
```

GalRogozinski: Also notice 2^21 is actually above the estimation so not sure how strict we should be

types/testingutils/beacon_node_versioned.go
```diff
@@ -163,18 +191,28 @@ var TestingSignedBeaconBlockV = func(ks *TestKeySet, version spec.DataVersion) s
...
 		}
-
+	case spec.DataVersionDeneb:
+		if vBlk.Deneb == nil {
+			panic("empty block contents")
+		}
+		if vBlk.Deneb.Block == nil {
+			panic("empty block")
+		}
+		return &deneb.SignedBeaconBlock{
```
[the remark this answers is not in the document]
MatheusFranco99: Well noticed. I'm fixing it now. At first, I used `version.SignedBeaconBlock` because it was already being used for the blinded proposer beacon roots, strangely enough. But the correct one should be the `SignedBlindedBeaconBlock` indeed.

types/consensus_data.go
MatheusFranco99: `ssz-max:"2048"` is also strange for me. But, since it's in go-eth2-client, I believe we should, for now, take that into account in the calculation.
On the other hand, 2048 * 8 seems to be a mistake for me. I think we should use just 2048. I.e.:
```python
ATTESTER_SLASHING_SIZE = 2 * (2 * (2048 + 96 + (2 * 8 + 32 + 2 * (8 + 32))))  # AttesterSlashing
ATTESTATION_SIZE = 128 * (2048 + 96 + (2 * 8 + 32 + 2 * (8 + 32)))  # Attestation
```
I think in the script, it's missing the graffiti:
```python
ETH1_DATA_SIZE = 96 + (2 * 32 + 8) + 32 # Randao + eth1data + graffiti
```
And if we consider the `ExecutionPayload` without the transactions, we would have:
```python
EXECUTION_PAYLOAD_NO_TRANSACTIONS = 32 + 20 + 2*32 + 256 + 32 + 4*8 + 3*32 + 16 * (2*8 + 20 + 8) + 8 + 8
```
Running it, I get 1,315,964 for everything but transactions. Adding 2^21 as the transactions upper limit, I get 3,413,116.
What do you think @GalRogozinski ?

y0sher (approved):

types/consensus_data.go
```diff
@@ -112,23 +111,51 @@ type ConsensusData struct {
...
+	//			ExecutionPayload(32 + 20 + 2*32 + 256 + 32 + 4*8 + 3*32 + 1048576*1073741824 + 16 * (2*8 + 20 + 8) + 8 + 8) +
+	//			BLSToExecutionChanges 16 * (96 + (8 + 48 + 20)) +
+	//			KZGCommitment 4096 * 48
+	// = 1315964 (everything but transactions) + 2^50 (transaction list)
 	// We do not need to support such a big DataSSZ size as 2^50 represents 1000X the actual block gas limit
 	// Current 30M gas limit produces 30M / 16 (call data cost) = 1,875,000 bytes (https://eips.ethereum.org/EIPS/eip-4488)
-	// we can upper limit transactions to 2^21, together with the rest of the data 2*2^21 = 2^22 = 4,194,304 bytes
-	// Exmplanation on why transaction sizes are so big https://github.com/ethereum/consensus-specs/pull/2686
-	DataSSZ []byte `ssz-max:"4194304"` // 2^22
+	// we can upper limit transactions to 2^21, together with the rest of the data 1315964 + 2^21  = 3,413,116 bytes
```
moshe-blox, line 132: Why did we go down from 2*2^21 to 2^21? If anything, blob size should have increased our ssz-max, no?

GalRogozinski: See this comment where we corrected a previous mistake regarding attestations:
https://github.com/bloxapp/ssv-spec/pull/324#discussion_r1455227526

checkout the commented python script

GalRogozinski: > they're thinking of increasing gas size from 30M to 40M very soon
When they are done thinking we'll fix this

it will only be on the next hard fork (best case) so there is time

types/consensus_data.go
moshe-blox: @MatheusFranco99 @GalRogozinski 2048 is bits I believe, not bytes.

```
64 attester committees * 2048 members per committee * 32 slots per epoch * 32 eth = 134,217,728
```

There's not enough ETH supply to have 2048 members per attester committee, so I'd guess the size is `2048 / 8` (bits to bytes).

moshe-blox: Additionally, see the Beacon spec increasing GOSSIP_MAX_SIZE from 1mb to 10mb post-bellatrix: https://github.com/ethereum/consensus-specs/blob/3727a757450feb6110e2d4e403604572a478343d/specs/bellatrix/p2p-interface.md?plain=1#L147-L153

moshe-blox: https://github.com/ethereum/consensus-specs/blob/3727a757450feb6110e2d4e403604572a478343d/specs/bellatrix/p2p-interface.md?plain=1#L155-L161

```
Considering that `BeaconBlock` max size is on the
order of 128 KiB in the worst case and the current gas limit (~30M) bounds max blocksize to less
than 2 MiB today, this marginal difference in theoretical bounds will have zero
impact on network functionality and security.
```

GalRogozinski: Yesterday I opened https://github.com/attestantio/go-eth2-client/issues/98 because of the 2048 bytes issue.
For the calculation we carried over the issue because we rely on the lib... but if all 3 of us agree it is a mistake on go-eth2-client side we can fix the calculation already.

Besides this I think the current calculation is more precise than yours because we go into details about the block structure.

About the 10% margin. According to the spec: 
`the current gas limit (~30M) bounds max blocksize to less
than 2 MiB today`
So we already have some margin by assuming 2MiB.

About 40M gas, for eth to change that I think it must be in a hard-fork. So I don't see a need to prepare in advanced... am I missing something?

types/consensus_data.go
moshe-blox: > > they're thinking of increasing gas size from 30M to 40M very soon
> > When they are done thinking we'll fix this
> 
> it will only be on the next hard fork (best case) so there is time

I think maybe we'd like to support a probable gas increase without having to fork?

types/consensus_data.go
MatheusFranco99: Personally, though I think it's correct, I don't feel comfortable using 2048/8 while their ssz-max is 2048 bytes.

About the transactions and gas, I could go either way, as you guys prefer. I think it's both easy to keep 30M and to switch to 40.

moshe-blox (approved):

olegshmuelov (approved):

PR #327, opened 2023-12-27
QBFT tests leftovers

MatheusFranco99:
#### Commit
- [ ] Decided full data
- [ ] Invalid full data
- [ ] Valid full data

#### Controller - Decided
- [ ] Valid full data

#### Messages
- [ ] Invalid hash data root
- [ ] Invalid prepare justification
- [ ] Invalid round change justifications unmarshalling
- [ ] Marshal justifications
- [ ] Prepare justifications unmarshalling
- [ ] Round change justifications unmarshalling
- [ ] Round change prepared
- [ ] SSZ Marshalling
- [ ] Valid Hash data root

#### Prepare
- [ ] Duplicate message quorum
- [ ] Set prepared full data

#### Proposal
- [ ] Valid full data

#### Round Change
- [x] Invalid Justification Full Data
- [ ] Valid Justification Full Data

#### Start Instance
- [ ] Equal height no running instance


qbft/spectest/tests/prepare/set_prepared_full_data.go
```diff
@@ -1,8 +1,36 @@
...
+
+	msgs := []*qbft.SignedMessage{
+		testingutils.TestingPrepareMessageWithRound(ks.Shares[1], 1, 10),
+	}
+	return &tests.MsgProcessingSpecTest{
+		Name:          "set prepared full data",
+		Pre:           pre,
+		InputMessages: msgs,
+	}
 }
```
GalRogozinski, line 36: I actually don't understand what state transition this tests

GalRogozinski (changes requested):

qbft/spectest/tests/commit/decided_full_data.go
```diff
@@ -1,8 +1,47 @@
...
+		testingutils.TestingCommitMessage(ks.Shares[3], 3),
+	}
+
+	outMsgs := []*qbft.SignedMessage{
+		testingutils.TestingPrepareMessage(ks.Shares[1], 1),
+		testingutils.TestingCommitMessage(ks.Shares[1], 1),
+		testingutils.TestingCommitMultiSignerMessage([]*bls.SecretKey{ks.Shares[1], ks.Shares[2], ks.Shares[3]}, []types.OperatorID{1, 2, 3}),
+	}
+
+	return &tests.ControllerSpecTest{
```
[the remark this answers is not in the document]
MatheusFranco99: I used the `ControllerSpecTest` because the I needed the `controller` structure since it is the one that broadcasts the decided messages.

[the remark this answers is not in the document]
MatheusFranco99: What should be the correct comment / instruction for this test?

qbft/spectest/tests/prepare/set_prepared_full_data.go
MatheusFranco99: Yeah. I was to ask you about it. I also don't get this test.
The instruction is to test "setting state.LastPreparedValue to the full data not the root". But this is how it is normally set. Maybe the instruction is for an outdated version. What do you think?

qbft/spectest/tests/commit/invalid_full_data.go
```diff
@@ -1,8 +1,34 @@
...
+	return &tests.MsgProcessingSpecTest{
+		Name:          "commit with invalid full data",
+		Pre:           pre,
+		InputMessages: msgs,
+		OutputMessages: []*qbft.SignedMessage{
+			testingutils.TestingPrepareMessage(ks.Shares[1], 1),
+			testingutils.TestingCommitMessage(ks.Shares[1], 1),
+		},
+	}
 }
```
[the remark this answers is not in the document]
MatheusFranco99: Yes, I agree 100% with your observation.
I talked about it in the PR description. May you take a look at it?
Since we don't use the `FullData` field when receiving the commit message, there shouldn't be a problem related to the QBFT execution.
However, regarding security, one can add a `FullData` field with maximum length to slow down the node processing.
For that, we can add a check that asserts that the field is empty in the commit message. What do you think?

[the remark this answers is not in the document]
GalRogozinski: opened #329 
add a comment in code that refers to this issue please

qbft/spectest/tests/prepare/set_prepared_full_data.go
GalRogozinski: What the comment says is to check that pre.State.LastPreparedValue gets updated correctly after a prepare quorum was reached...

But the happyflow test already tests this...

So I think delete this test and add a github comment to why

MatheusFranco99:
The test "Set prepared full data" was dropped because it's already tested in the happy flow test.

https://github.com/bloxapp/ssv-spec/pull/327#discussion_r1438870719

qbft/spectest/tests/commit/valid_full_data.go
```diff
@@ -1,8 +1,34 @@
...
+	ks := testingutils.Testing4SharesSet()
+
+	msgs := []*qbft.SignedMessage{
+		testingutils.TestingProposalMessage(ks.Shares[1], 1),
+
+		testingutils.TestingPrepareMessage(ks.Shares[1], 1),
+		testingutils.TestingPrepareMessage(ks.Shares[2], 2),
+		testingutils.TestingPrepareMessage(ks.Shares[3], 3),
+
+		testingutils.TestingCommitMessageWithFullData(ks.Shares[1], 1, testingutils.TestingQBFTFullData),
```
GalRogozinski, line 21: do we need this? It is enough to have a proposals + prepares, no?

qbft/spectest/tests/commit/decided_full_data.go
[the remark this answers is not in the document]
GalRogozinski: I think that the comment should simply be moved to `controller/valid_full_data.go` and here it can be deleted imho.
I don't see much difference between the 2.

If you can think of a way to make a meaningful `MsgProcessingTest` for this be my guest... but I am not sure how

qbft/spectest/tests/commit/valid_full_data.go
```diff
@@ -1,8 +1,34 @@
...
+		testingutils.TestingCommitMessageWithFullData(ks.Shares[2], 2, testingutils.TestingQBFTFullData),
+		testingutils.TestingCommitMessageWithFullData(ks.Shares[3], 3, testingutils.TestingQBFTFullData),
```
GalRogozinski, line 23: why do we need this? enough to have proposal +prepares, no?

GalRogozinski (changes requested):

qbft/spectest/tests/commit/valid_full_data.go
MatheusFranco99: By the test desc:
> tests the signed commit with a valid full data field (H(full data) == root)

Isn't it supposed to process at least one commit message as described? I could make it process only one, though I included three of them.

qbft/spectest/tests/commit/decided_full_data.go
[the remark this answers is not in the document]
MatheusFranco99: Yup, agree 100%

qbft/spectest/tests/commit/valid_full_data.go
GalRogozinski: I see the comment saying:
> // ValidFullData tests signed decided with a valid full data field

so I thought you can just use the decided message... actually it should be able to terminate the instance w.o proposals and prepares, no?

MatheusFranco99: Wait. In this file I see

> // ValidFullData tests the signed COMMIT with a valid full data field (H(full data) == root)

Sure you see "decided"?

GalRogozinski: Looked at wrong test

GalRogozinski (approved):

PR #328, opened 2023-12-29
QBFT Tests - Instance & Message

MatheusFranco99:
#### Messages
- [x] Invalid hash data root - `message.Validate()` doesn't test `root` field. Should this test be dropped?
- [x] Invalid prepare justification
- [x] Invalid round change justifications unmarshalling
- [x] Marshal justifications
- [x] Prepare justifications unmarshalling
- [x] Round change justifications unmarshalling
- [x] Round change prepared
- [x] SSZ Marshalling
- [x] Valid Hash data root - `message.Validate()` doesn't test `root` field. Should this test be dropped?

#### Start Instance
- [x] Equal height no running instance


qbft/spectest/tests/messages/invalid_prepare_justification_unmarshaling.go
```diff
@@ -1,8 +1,33 @@
...
-import "github.com/bloxapp/ssv-spec/qbft/spectest/tests"
+import (
+	"github.com/bloxapp/ssv-spec/qbft"
+	"github.com/bloxapp/ssv-spec/qbft/spectest/tests"
+	"github.com/bloxapp/ssv-spec/types"
+	"github.com/bloxapp/ssv-spec/types/testingutils"
+)
 
 // InvalidPrepareJustificationsUnmarshalling tests unmarshalling invalid prepare justifications (during message.validate())
 func InvalidPrepareJustificationsUnmarshalling() tests.SpecTest {
```
GalRogozinski, line 11: @alonmuroch
It seems this is mainly testing ssz logic that is the repsonsibility of a 3rd party lib.

We can have this basic test but just wondering what is your view of testing such parts of the code?

qbft/spectest/tests/messages/prepare_justification_unmarshaling.go
```diff
@@ -1,8 +1,63 @@
...
+		}
+		root2, err := prepareMsgs[idx].GetRoot()
+		if err != nil {
+			panic(err)
+		}
+		if !bytes.Equal(root1[:], root2[:]) {
+			panic("Unmarshalled message is different")
+		}
+	}
+
```
GalRogozinski, line 56: @moshe-blox will this work when you run on the node?

@MatheusFranco99
Let's wait to see what @alonmuroch says regarding ssz logic

GalRogozinski: Anyhow , it makes more sense to use the `ExpectedRoots` in `MsgSpecTest`

qbft/spectest/tests/messages/round_change_justification_unmarshaling.go
```diff
@@ -1,8 +1,56 @@
...
+			panic(err)
+		}
+		root2, err := rcMsgs[idx].GetRoot()
+		if err != nil {
+			panic(err)
+		}
+		if !bytes.Equal(root1[:], root2[:]) {
+			panic("Unmarshalled message is different")
+		}
+	}
```
GalRogozinski, line 48: ditto
should use `ExpectedRoots`

but let see what Alon says

qbft/spectest/tests/messages/round_change_prepared.go
```diff
@@ -1,8 +1,31 @@
...
+	msg := testingutils.TestingRoundChangeMessageWithParams(
+		ks.Shares[1], types.OperatorID(1), 2, qbft.FirstHeight, testingutils.TestingQBFTRootData, 1, prepareMarshalled)
+
+	return &tests.MsgSpecTest{
+		Name: "round change prepared",
+		Messages: []*qbft.SignedMessage{
+			msg,
+		},
+	}
 }
```
GalRogozinski, line 31: I actually don't understand the point of this test. We're checking our `testingutils`?

I am not sure what test makes sense here, @alonmuroch?

qbft/spectest/tests/messages/ssz_marshaling.go
```diff
@@ -1,8 +1,66 @@
...
+	}
+	root2, err := unmarshalledMsg.GetRoot()
+	if err != nil {
+		panic(err.Error())
+	}
+
+	if !bytes.Equal(root1[:], root2[:]) {
+		panic("Unmarshalled message is different.")
+	}
+
```
GalRogozinski, line 59: ditto `ExpectedRoots`

qbft/spectest/tests/messages/valid_hash_data_root.go
```diff
@@ -1,8 +1,30 @@
...
+
+	msg.FullData = testingutils.TestingQBFTFullData
+
+	return &tests.MsgSpecTest{
+		Name: "valid hash data root",
+		Messages: []*qbft.SignedMessage{
+			msg,
+		},
+	}
 }
```
GalRogozinski, line 30: I don't think this is what @alonmuroch meant to implement... but I am not sure what he did mean

GalRogozinski (changes requested):
We need to discuss whether the `MsgSpecTest` needs to be changed to have better validity checks.

Also we need to understand why we need `CreateMsgTest` and `MsgSpecTest`

qbft/spectest/tests/messages/invalid_hash_data_root.go
```diff
@@ -1,8 +1,30 @@
...
+		MsgType:    qbft.ProposalMsgType,
+		Height:     qbft.FirstHeight,
+		Round:      qbft.FirstRound,
+		Identifier: []byte{1, 2, 3, 4},
+		Root:       testingutils.DifferentRoot,
+	})
+
+	msg.FullData = testingutils.TestingQBFTFullData
+
+	return &tests.MsgSpecTest{
```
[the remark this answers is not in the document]
MatheusFranco99: You mean the instance `BaseMsgValidation` or the controller one?
In the test structure we don't have any of these objects btw. Should I add it?
My only concern is that the other tests will probably need to change.
Another way would be to move this single test to ssv. What do you think?

qbft/spectest/tests/messages/ssz_marshaling.go
```diff
@@ -1,8 +1,66 @@
...
+	"bytes"
+
+	"github.com/bloxapp/ssv-spec/qbft"
+	"github.com/bloxapp/ssv-spec/qbft/spectest/tests"
+	"github.com/bloxapp/ssv-spec/types"
+	"github.com/bloxapp/ssv-spec/types/testingutils"
+)
 
 // SSZMarshaling tests a valid marshaling of a signed message
 func SSZMarshaling() tests.SpecTest {
```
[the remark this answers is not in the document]
MatheusFranco99: Yup they are basically the same, though were created with different purposes I guess.

qbft/spectest/tests/messages/valid_hash_data_root.go
MatheusFranco99: I'm also confused on this. :/

qbft/spectest/tests/messages/invalid_hash_data_root.go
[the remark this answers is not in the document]
GalRogozinski: Maybe we open an issue and just remember to fix this when we do spec rethink

qbft/spectest/tests/messages/ssz_marshaling.go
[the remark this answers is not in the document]
GalRogozinski: I am just annoyed
I don't care if we leave or delete it even

in spec rethink we can make sure we have at least less duplicates

GalRogozinski (commented):
So we just need @alonmuroch to explain us the test and we're done

qbft/spectest/tests/controller_spectest.go
```diff
@@ -216,6 +216,10 @@ func (test *ControllerSpecTest) overrideStateComparison(t *testing.T) {
 func (test *ControllerSpecTest) GetPostState() (interface{}, error) {
 	contr := test.generateController()
 
+	if test.StartHeight != nil {
+		contr.Height = *test.StartHeight
+	}
+
```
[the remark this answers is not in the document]
MatheusFranco99: I added it in the `Run` function but forgot to add it in the `GetPostState`
None of these tests failed but another test that I did for another PR (in which I tried to use the StartHeight) would fail (that's actually why I remembered about it)

qbft/spectest/tests/messages/ssz_marshaling.go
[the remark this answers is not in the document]
MatheusFranco99: Sure. As you want :)

GalRogozinski (approved):

y0sher (approved):

PR #330, opened 2024-01-01
Clarify `DataRound` with comment

### LINKED ISSUE #320: RoundChangePrepared() doesn't check if there was a preparation
https://github.com/bloxapp/ssv-spec/blob/61f5b994c10eef3d71d8d95ab9fbff3859f92841/qbft/messages.go#L91-L98

GalRogozinski:
# Description

Closes #320

Add comments to `DataRound` so it will be clear why `RoundChangePrepared()`  works as intended

### DIFF
--- a/qbft/messages.go
+++ b/qbft/messages.go
@@ -51,7 +51,7 @@ type Message struct {
 	Identifier []byte `ssz-max:"56"` // instance Identifier this msg belongs to
 
 	Root                     [32]byte `ssz-size:"32"`
-	DataRound                Round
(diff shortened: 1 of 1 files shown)


MatheusFranco99 (approved):

moshe-blox (approved):
makes sense.

i guess it's not worth changing the field name because it'd break the tests?

GalRogozinski:
@moshe-blox 
it will cause changes in jsons...
Maybe it is actually a good idea to change the name of the field

PR #332, opened 2024-01-01
Aggregator Consensus data with no justifications

GalRogozinski:
# Description

Encoding test for Aggregator Consensus data with no justifications


types/spectest/tests/consensusdata/aggregator_no_justifications.go
```diff
@@ -1,8 +1,26 @@
 package consensusdata
 
-import "github.com/bloxapp/ssv-spec/types/spectest/tests"
+import (
+	"github.com/bloxapp/ssv-spec/types/spectest/tests"
+	"github.com/bloxapp/ssv-spec/types/testingutils"
+)
 
-// AggregatorNoJustifications tests an invalid consensus data with no aggregator pre-consensus justifications
```
GalRogozinski, line 3: Note that it is no issue to have an empty pre-con justification field here

GalRogozinski:
What is in #344 is good enough
opened #353 to complete this

Issue #264, opened 2023-05-29
Missing Tests


GalRogozinski:
### Description

In general make sure that after errors we can recover

- [ ] runner/postconsensus - DuplicateMsg and then quorum (make sure error doesn't stop us)
- [ ] runner/consensus - finish
- [ ] [latemsg/late_proposal_past_instance.go](https://github.com/bloxapp/ssv-spec/pull/192/files/88af956de13f40c81cfe094170fd3092d15f3a4b#diff-1f397f33fe5123cb2bb6cdafd9bab3e7b1871cd49bb47e809e4c49b208b62b8b) - create a version that sends a late non-duplicate message

PR #333, opened 2024-01-02
Post-consensus test: Duplicate message then quorum test

MatheusFranco99:
- [x] Post-consensus - DuplicateMsg and then quorum. The test asserts that the error doesn't disallow terminating a duty after a quorum is reached.


GalRogozinski (approved):
Nice!

GalRogozinski:
refers to #264

moshe-blox (approved):
LGTM

y0sher (approved):

GalRogozinski (changes requested):
Hey @MatheusFranco99 

Don't want to do changes in your PR myself (besides merging and resolving conflicts).
Shouldn't the tests here be "Duplicate message with different roots then quorum"...

It is just that we have `duplicate_msg.go` and `duplicate_msg_different_root.go` and this test seems like an extension of the latter...

maybe other good tests will be to create quorum of dup messages and a test with one dup message that has the same root (not for this PR)

MatheusFranco99:
Hey @GalRogozinski 
The test is indeed a duplicate message (from an already received signer) with a different signing root and then the operator receives a quorum successfully.

You mean that a more appropriate test name (and file name) should include something like "msg different root", correct?
I agree. Will wait on your response to change it :)

GalRogozinski:
@MatheusFranco99 yes

MatheusFranco99:
@GalRogozinski done

GalRogozinski (approved):

PR #337, opened 2024-01-02
Type test: share

MatheusFranco99:
## Spec Test
New test type to verify share's functionality of checking quorums.

Tests:
- [x] Message with partial quorum
- [x] Message with quorum
- [x] Message with 3f+1 quorum
- [x] Message with no partial quorum
- [x] Message with no partial quorum with duplicate signers
- [x] Message with partial quorum with duplicate signers
- [x] Message with no quorum
- [x] Message with no quorum with duplicate signers
- [x] Message with quorum with duplicate signers

Solving one element of [issue 25](https://github.com/bloxapp/ssv-spec/issues/25).


types/spectest/tests/share/no_quorum.go
```diff
@@ -1,6 +1,24 @@
...
+	msg := testingutils.TestingCommitMultiSignerMessage([]*bls.SecretKey{ks.Shares[1], ks.Shares[2]}, []types.OperatorID{1, 2})
+
+	return &ShareTest{
+		Name:                     "no quorum",
+		Share:                    *share,
+		Message:                  *msg,
+		ExpectedHasPartialQuorum: true,
+		ExpectedHasQuorum:        false,
+		ExpectedFullCommittee:    false,
+	}
```
GalRogozinski, line 23: The logic is pretty much the same as `has_partial_quroum`?
If so maybe we might as well delete this test and change the name of `has_partial_quorum` to account for 2 scenarios?

Having the same test for 2 different scenarios might also be an ok approach, but I think we have enough clutter in our tests anyhow

MatheusFranco99: Agree

GalRogozinski (approved):

y0sher (approved):

PR #340, opened 2024-01-03
Test - Validator

MatheusFranco99:


ssv/spectest/tests/validator/valid_consensus.go
```diff
@@ -0,0 +1,76 @@
+package validator
+
+import (
+	"github.com/bloxapp/ssv-spec/qbft"
+	"github.com/bloxapp/ssv-spec/ssv/spectest/tests"
+	"github.com/bloxapp/ssv-spec/types"
+	"github.com/bloxapp/ssv-spec/types/testingutils"
+)
+
+func ValidConsensus() tests.SpecTest {
```
GalRogozinski, line 10: add comment

GalRogozinski (commented):
hmm...
I see Alon did add this as an issue but on another instance he told me that we don't need this in spec... I also didn't plan doing such tests for the spec rethink... putting this on hold for now

ssv/spectest/tests/validator/test.go
```diff
@@ -0,0 +1,128 @@
...
+	for _, r1 := range test.BeaconBroadcastedRoots {
+		found := false
+		for _, r2 := range broadcastedRoots {
+			if r1 == hex.EncodeToString(r2[:]) {
+				found = true
+				break
+			}
+		}
+		require.Truef(t, found, "broadcasted beacon root not found")
+	}
```
GalRogozinski, line 98: same issue as below I guess

ssv/spectest/tests/validator/test.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/bloxapp/ssv-spec/types/testingutils"
+	"github.com/stretchr/testify/require"
+)
+
+type ValidatorTest struct {
+	Name                   string
+	Duties                 []*types.Duty
+	Messages               []*types.SSVMessage
+	OutputMessages         []*types.SSVMessage
+	BeaconBroadcastedRoots []string
```
GalRogozinski, line 21: I noticed you don't really use this field for the tests...
Not much harm that it is here I suppose

your choice whether to remove it or not

ssv/spectest/tests/validator/test.go
```diff
@@ -0,0 +1,128 @@
...
+				if msg.GetType() == msg2.GetType() && msg.GetID() == msg2.GetID() {
+					found = true
+					usedIndexes[index2] = true
+					break
+				}
+			}
+		}
+		if !found {
+			panic(fmt.Sprintf("Broadcasted message not found in output messages: %v", msg))
+		}
```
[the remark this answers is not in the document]
MatheusFranco99: Before it I do:
```go
require.Len(t, broadcastedMsgs, len(test.OutputMessages))
```
So I think they are tested to be equal. What do you think?

ssv/spectest/tests/validator/test.go
MatheusFranco99: Before it I also check the length. But indeed I should add an "already used" map as in the other check. Adding it.

ssv/spectest/tests/validator/test.go
MatheusFranco99: Which field?

ssv/spectest/tests/validator/test.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+}
+
+func (test *ValidatorTest) GetPostState() (interface{}, error) {
+	return nil, nil
```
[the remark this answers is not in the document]
MatheusFranco99: Adding. Thanks!

ssv/spectest/tests/validator/test.go
GalRogozinski: BeaconBroadcastedRoots

GalRogozinski: it is just that in the tests itself I noticed it was always empty
but arguably that's also a check

MatheusFranco99: Oh, I see. Will take a look. Thx!

MatheusFranco99:
@GalRogozinski 
This PR added tests for the validator structure, which we almost didn't have. I think it may be useful to have such tests. But the structure changed a lot since then. To merge it, I would need to do some work here

Issue #235, opened 2023-05-11
Future decide with no running instance panic


olegshmuelov:
when there is no running instance and we try to check if the future msg decided correctly we will get a panic (attack vector).
https://github.com/bloxapp/ssv-spec/blob/b8be13458287f1099a8edfd8115d0814eb85fb2f/ssv/runner.go#L191

solution:
check `RunningInstance` for nil
```go
decidedRunningInstance := decided && b.State.RunningInstance != nil && decidedMsg.Message.Height == b.State.RunningInstance.GetHeight()
```

add tests for specific scenario for all related duty runners

btw the implementation has the `nil` check

GalRogozinski:
@olegshmuelov 
now that #300 removed future message you think this is an issue?
I can add the nil check but not sure if there are relevant tests

PR #342, opened 2024-01-05
Fix - Future decided no instance nil check

MatheusFranco99:
Fix https://github.com/bloxapp/ssv-spec/issues/235 by adding a `nil` check.

New test:
- [x] Future decided with no running instance

### DIFF
--- a/ssv/runner.go
+++ b/ssv/runner.go
@@ -195,15 +195,18 @@ func (b *BaseRunner) basePartialSigMsgProcessing(
 
 // didDecideCorrectly returns true if the expected consensus instance decided correctly
 func (b *BaseRunner) didDecideCorrectly(prevDecided bool, decidedMsg *qbft.SignedMessage) (bool, error) {
-	decided := decidedMsg != nil
-	decidedRunningInstance := decided && b.State.RunningInstance != nil && decidedMsg.Message.Height == b.State.RunningInstance.GetHeight()
-
-	if !decided {
+	if decidedMsg == nil {
 		return false, nil
 	}
-	if !decidedRunningInstance {
+
+	if b.State.RunningInstance == nil {
 		return false, errors.New("decided wrong instance")
 	}
+
+	if decidedMsg.Message.Height != b.State.RunningInstance.GetHeight() {
+		return false, errors.New("decided wrong instance")
+	}
+
 	// verify we decided running instance only, if not we do not proceed
 	if prevDecided {
 		return false, nil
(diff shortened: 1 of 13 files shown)


GalRogozinski (approved):

ssv/spectest/tests/runner/consensus/future_decided_no_instance.go
```diff
@@ -0,0 +1,114 @@
...
+		ssvMessage := &types.SSVMessage{
+			MsgType: types.SSVConsensusMsgType,
+			MsgID:   *(*types.MessageID)(getID(role)),
+			Data:    byts,
+		}
+
+		return ssvMessage
+	}
+
+	multiSpecTest := &tests.MultiMsgProcessingSpecTest{
```
y0sher, line 47: we don't need this test for proposer?

MatheusFranco99: The proposer and blinded proposer tests are in the end of the file

ssv/runner.go
```diff
@@ -196,7 +196,7 @@ func (b *BaseRunner) basePartialSigMsgProcessing(
 // didDecideCorrectly returns true if the expected consensus instance decided correctly
 func (b *BaseRunner) didDecideCorrectly(prevDecided bool, decidedMsg *qbft.SignedMessage) (bool, error) {
 	decided := decidedMsg != nil
```
y0sher, line 198: I suggest to write it differently, this is how we write stuff lie this in go and I also think is more readable and easier to reason about
```golang
if decidedMsg == nil {
 return false, nil
}
if b.State.RunningInstance == nil {
 return false, errors.New("decided wrong instance")
}

if decidedMsg.Message.Height != b.State.RunningInstance.GetHeight() {
 return false, errors.New("decided wrong instance")
}
```

y0sher: let me know what you guys think, I know its not a part of this PR but I couldn't help but see this
@MatheusFranco99 , @GalRogozinski , @moshe-blox . its ok also to decide not to do it here.

y0sher (approved):
small code style comment, approved in case you don't want to do it now

ssv/runner.go
MatheusFranco99: I 100% agree. @GalRogozinski what do you think?

ssv/runner.go
```diff
@@ -195,15 +195,18 @@ func (b *BaseRunner) basePartialSigMsgProcessing(
...
-	decidedRunningInstance := decided && decidedMsg.Message.Height == b.State.RunningInstance.GetHeight()
-
-	if !decided {
+	if decidedMsg == nil {
 		return false, nil
 	}
-	if !decidedRunningInstance {
+
+	if b.State.RunningInstance == nil {
+		return false, errors.New("decided wrong instance")
```
olegshmuelov, line 203: we should return different err message here

MatheusFranco99: I agree

PR #345, opened 2024-01-10
SignedSSVMessage

MatheusFranco99:
## Overview
Introduce `SignedSSVMessage`, which encapsulates `SSVMessage` along with a `signature` and a sender `identifier` fields.

### Changes
- New `SignedSSVMesage` structure in `types`
- `DecodePubsubMsg` now returns a `SignedSSVMesage`
- `MsgValidation` adapted to receive a `SignedSSVMesage` instead of a `SSVMessage`.

Note: though `SSVMessage` is no longer the transmitted message in the network, the `Runner` still processes the type `SSVMessage` in the `ProcessMessage` function. 

### Tests
- [x] Encoding
- [x] Valid SignedSSVMessage
- [x] Invalid SignedSSVMessage with no data
- [x] Invalid SignedSSVMessage with empty signature
- [x] Invalid SignedSSVMessage with zero signer
- [x] Invalid SignedSSVMessage with wrong data


p2p/validation/msg_validation.go
```diff
@@ -15,11 +16,19 @@ type MsgValidatorFunc = func(ctx context.Context, p peer.ID, msg *pubsub.Message
 
 func MsgValidation(runner ssv.Runner) MsgValidatorFunc {
 	return func(ctx context.Context, p peer.ID, msg *pubsub.Message) pubsub.ValidationResult {
-		ssvMsg, err := DecodePubsubMsg(msg)
+		signedSSVMsg, err := DecodePubsubMsg(msg)
```
GalRogozinski, line 19: maybe this is a complaint to myself.. but we should have deleted MsgValidation since it is definitely not what the code does at the moment

types/generate.go
```diff
@@ -7,7 +7,7 @@ package types
 //go:generate go run github.com/ferranbt/fastssz/sszgen --path share.go --include ./operator.go,./messages.go,./signer.go,./domain_type.go
 
 //go:generate rm -f ./messages_encoding.go
-//go:generate go run github.com/ferranbt/fastssz/sszgen --path messages.go --exclude-objs ValidatorPK,MessageID,MsgType
+//go:generate go run github.com/ferranbt/fastssz/sszgen --path messages.go --include ./operator.go --exclude-objs ValidatorPK,MessageID,MsgType
```
GalRogozinski, line 10: not sure why this changed?

types/messages.go
```diff
@@ -125,3 +127,62 @@ func (msg *SSVMessage) Encode() ([]byte, error) {
 func (msg *SSVMessage) Decode(data []byte) error {
 	return msg.UnmarshalSSZ(data)
 }
+
+// SSVMessage is the main message passed within the SSV network. It encapsulates the SSVMessage structure and a signature
+type SignedSSVMessage struct {
+	OperatorID OperatorID
+	Signature  []byte `ssz-max:"512"`     // Current signature max size allow keys up to 512*8 = 4096 bits
```
GalRogozinski, line 134: maybe it is just nice to comment that this is created by the operator's private key

GalRogozinski (commented):
Looks great!

types/generate.go
GalRogozinski: Because of `OperatorID` addition

p2p/validation/msg_validation.go
MatheusFranco99: Sure, let's remove it in another PR

MatheusFranco99:
@GalRogozinski 
The current tests only check `msg.Validate()` which doesn't check the signature. This is because we don't specify here how this signature should be created or validated.
I didn't add such a restriction on the type of signature because the message validation module (in p2p/validation) will be removed from spec in a future PR (right?) and the validator or runner don't have access to this signature.
What do you think?

GalRogozinski:
@MatheusFranco99 

The spec should achieve 2 objectives in my eyes:
1. Make sure that the node implementation doesn't accidentally change the protocol w.o detection
2. Allow anyone to implement a node, ideally w.o looking at any other source.

Currently the spec doesn't exactly reach those objectives and in the spec rewrite I am hoping to change this.
In the meanwhile I wish to do things correctly for new additions :-)

Imagine someone changes the RSA parameters (for better security) and then creates an incompatibility.
This should be caught by the spec

MatheusFranco99:
@GalRogozinski
Sure, I'll be adding a check in message validation to verify the RSA signature and tests for it.

types/spectest/tests/signedssvmsg/test.go
```diff
@@ -0,0 +1,55 @@
...
+		// check RSA signature
+		if err == nil {
+			var pk *rsa.PublicKey
+			pk, err = types.PemToPublicKey(test.RSAPublicKey)
+			if err != nil {
+				panic(err.Error())
+			}
+
+			messageHash := sha256.Sum256(msg.Data)
+			err = rsa.VerifyPKCS1v15(pk, crypto.SHA256, messageHash[:], msg.Signature)
```
y0sher, line 46: for a future PR, something we are working in ssv-node is encapsulating all operator pubkey/private key operation to make them less dependent on specific functions. so if for example we'll want to change our RSA implementation it will be easy.

y0sher (approved):
FYI this is probably a fork. we are creating signed messages in the node a bit differently right now.
https://github.com/bloxapp/ssv/blob/main/network/commons/common.go#L43

it might be good idea to consider using the same approach in the node unless this fork has some significant advantages.

types/spectest/tests/signedssvmsg/test.go
MatheusFranco99: That's great. I will try to encapsulate it here too.

MatheusFranco99:
@y0sher 
Yes, I'm aware of that. The point of this PR is not only to define a type for this buffer approach that we're using on the node, but it also is a pre-requisite for the next "reduce signatures" PR. This signed message type will now be used in the runner module.

GalRogozinski (changes requested):
@y0sher @MatheusFranco99 

This is a fork because the implementation puts signature first and the PR puts operator first, right?

If so we should change this to not be a fork because I don't think there is a real advantage here...

y0sher:
> @y0sher @MatheusFranco99
> 
> This is a fork because the implementation puts signature first and the PR puts operator first, right?
> 
> If so we should change this to not be a fork because I don't think there is a real advantage here...

Its a fork because the implementation uses its own way of encoding the sig and operators. even if you create the struct in the PR in the same order it doesn't mean the final encoded data will be the same.

GalRogozinski:
@moshe-blox @y0sher 

There are 2 things we can do here:
1. Delete the ssz encoding and use the same encoding as impl
2. Make this part of the upcoming fork


The only reason I would use SSZ is just for the sake of consistency, since we use SSZ everywhere. Besides this I don't see an advantage. But this is good enough for me to opt for 2.
If you guys don't want to change to SSZ then we can go for 1.

GalRogozinski (approved):
Currently we fork so changed order is fine

Issue #303, opened 2023-08-14
Fix error on proposal/post-decided


GalRogozinski:
# Description

Currently we receive "proposal is not valid with current state" error. This sounds ok, but it isn't because the error doesn't stem from the fact that the instance is decided.

PR #347, opened 2024-01-12
Fix: proposal / post_decided error

MatheusFranco99:
Fix issue: https://github.com/bloxapp/ssv-spec/issues/303

- Changed error message and logic.
- Updated tests.


GalRogozinski (commented):
I think my thoughts for this issue were that there was too much processing going on for a decided instance. Also the fact that the same error happens for an active and a decided instance bothered.

This current fix isn't solving this and there's also a new untested error.
However, I don't think the issue is important enough to be put in priority so I'll just be closing the pull request

PR #348, opened 2024-01-16
Remove encryption functionality and tests

MatheusFranco99:
Encryption functions were not used anywhere except from their own tests. Thus, we deprecate this functionality and its respective tests.


GalRogozinski:
I am still thinking about this...
@moshe-blox confirmed the test wasn't used...
But there is a point of having such test in the spec, depending on how we define the scope

I'll talk to @lior-blox about this...

MatheusFranco99:
@GalRogozinski 
It seems that this should be closed, right? As it's part of the protocol being used somewhere

PR #352, opened 2024-01-21
Round change according to duty start time

GalRogozinski:
Implements https://github.com/bloxapp/SIPs/pull/37

Changes logic to all duties but proposer

### DIFF
--- a/qbft/timeout.go
+++ b/qbft/timeout.go
@@ -4,6 +4,7 @@ import (
 	"github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/bloxapp/ssv-spec/types"
 	"github.com/pkg/errors"
+	"golang.org/x/exp/constraints"
 )
 
 var (
@@ -43,7 +44,7 @@ func (t *RoundTimer) TimeoutForRound(round Round) uint64 {
 func RoundTimeout(round Round, height Height, baseDuration uint64, network types.BeaconNetwork) uint64 {
 	// Calculate additional timeout in seconds based on round
 	var additionalTimeout uint64
-	additionalTimeout = uint64(round) * quickTimeout
+	additionalTimeout = uint64(min(round, quickTimeoutThreshold)) * quickTimeout
 	if round > quickTimeoutThreshold {
 		slowPortion := uint64(round-quickTimeoutThreshold) * slowTimeout
 		additionalTimeout += slowPortion
@@ -100,4 +101,12 @@ func (i *Instance) UponRoundTimeout() error {
 	}
 
 	return nil
+}
+
+// min returns the minimum of two values
+func min[T constraints.Ordered](a, b T) T {
+	if a < b {
+		return a
+	}
+	return b
 }
--- a/qbft/timeout.go
+++ b/qbft/timeout.go
@@ -1,17 +1,51 @@
 package qbft
 
 import (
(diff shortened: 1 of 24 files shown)


qbft/timeout.go
```diff
@@ -1,18 +1,84 @@
...
+		return DefaultTimeout(round)
+	}
+}
+
+// RoundTimeout returns the unix epoch time (seconds) in which we should send a RC message
+// Called for all beacon duties other than proposals
+func RoundTimeout(round Round, height Height, baseDuration int64, network types.BeaconNetwork) int64 {
+	// Calculate additional timeout in seconds based on round
+	var additionalTimeout int64
+	additionalTimeout = int64(min(round, quickTimeoutThreshold)) * quickTimeout
```
GalRogozinski, line 47: different than impl

qbft/timeout.go
```diff
@@ -1,18 +1,84 @@
...
+	//current unix epoch time in seconds
+	CurrentTime int64
+}
+
+func (t *RoundTimer) TimeoutForRound(round Round) int64 {
+	switch t.Role {
+	case types.BNRoleAttester, types.BNRoleSyncCommittee:
+		return max(AttestationOrSyncCommitteeTimeout(round, t.Height, t.Network)-t.CurrentTime, 0)
+	case types.BNRoleAggregator, types.BNRoleSyncCommitteeContribution:
+		return max(AggregationOrContributionTimeout(round, t.Height, t.Network)-t.CurrentTime, 0)
```
GalRogozinski, line 36: different than impl

not sure if this can happen even if the system clock is bad...

MatheusFranco99 (approved):
LGTM

qbft/timeout.go
```diff
@@ -1,18 +1,85 @@
...
 
+type RoundTimer struct {
+	// role is the role of the QBFT instance
+	Role types.BeaconRole
+	// height is the current height of the instance
+	Height Height
+	// network is the beacon network
+	Network types.BeaconNetwork
+	//current unix epoch time in seconds
+	CurrentTime int64
```
moshe-blox, line 29: I think we can use `time.Time` and `time.Duration` instead of integers, makes it a bit more friendly.

For tests, you can use `time.Unix()` to create time.Time from an integer.

qbft/timeout.go
```diff
@@ -1,18 +1,85 @@
...
-	slowTimeout           = 2 * time.Minute //nolint
+	// quickTimeoutThreshold is the round after which the timeout duration increases
+	quickTimeoutThreshold = Round(6) //nolint
+	// quickTimeout is the timeout in seconds for the first 6 rounds
+	quickTimeout int64 = 2 // 2 seconds
+	// slowTimeout is the timeout in seconds for rounds after the first 86
+	slowTimeout int64 = 120 // 2 minutes
 	// CutoffRound which round the instance should stop its timer and progress no further
-	CutoffRound = 15 // stop processing instances after 8*2+120*6 = 14.2 min (~ 2 epochs)
+	CutoffRound = 14 // stop processing instances after 6*2+120*7 = 14.2 min (~ 2 epochs)
```
moshe-blox, line 18: Should we make the cutoff change we discussed here or it wouldn't fit this PR?

qbft/timeout.go
GalRogozinski: I intentionally don't want the spec to rely on go standard lib

PR #356, opened 2024-01-31
Performance benchmark tests

MatheusFranco99:
## Overview

Benchmark tests to:
- measure performance
- keep track of performance impact for each new protocol change

It includes performance tests for:
- [x] Full duty
- [x] Consensus full round
- [x] Message type by round and occurrence number

### DIFF
--- a/ssv/spectest/benchmark/happy_flow_test.go
+++ b/ssv/spectest/benchmark/happy_flow_test.go
@@ -1,0 +1,201 @@
+package benchmark
+
+import (
+	"fmt"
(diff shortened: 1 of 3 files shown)


MatheusFranco99:
@GalRogozinski 
I think this PR is super useful for us to have a view about our protocol and how our changes are improving it
Not only that, I find it super useful to have these views for later optimizations.
But, since I created this with not much care (just as some util for me), I would prefer putting it into a better shape before merging it, and even adding more analytics

PR #357, opened 2024-01-31
Add jsons

GalRogozinski:
# Description

Add state comparison jsons for types and signedMessage.
Now any change in the tests is easily detectable.


qbft/spectest/tests/create_message_spectest.go
```diff
@@ -66,6 +67,8 @@ func (test *CreateMsgSpecTest) Run(t *testing.T) {
 		require.Fail(t, "post state not equal", diff)
 	}
 	require.EqualValues(t, test.ExpectedRoot, hex.EncodeToString(r[:]))
+
+	typescomparable.CompareWithJson(t, test, test.TestName(), reflect.TypeOf(test).String())
```
moshe-blox, line 71: what is the point of this if the root is being compared just above? just so we can see the diffs clearly?

MatheusFranco99 (approved):

qbft/spectest/tests/create_message_spectest.go
MatheusFranco99: I think the above root test is just a check on the expected root of the created message.
But, the `CompareWithJson` should show us any difference in the test "content".
@GalRogozinski correct me if I'm wrong

PR #359, opened 2024-02-01
Improve performance for test generation and execution

MatheusFranco99:
# Overview

The purpose of this PR is to optimize the performance of test generation and execution with the purpose of improving work quality on future changes.

The `ssv` module is the one that takes the most time. After running a CPU profiling test, we noticed that the `NewTestingKeyManager` function consumes 80% of the total time, which is explained by the fact that it's called multiple times and takes 26 milliseconds to run (according to benchmark tests).

Nonetheless, the `NewTestingKeyManager` function returns a `testingKeyManager` object that is only used to validate and sign data but does not change state throughout a test execution. Thus, we propose adding a singleton pattern for this object.

## Metrics

To illustrate the number of times the `NewTestingKeyManager` function is called, take a look at the following table

Note: the number after the dash (/) is the number of different instances created, which depends on a `slashableDataRoots` parameter.

| Module | Test generation | Test execution |
|--------|-----------------|----------------|
| QBFT   | 368/1           | 390/1          |
| SSV    | 17058/1         | 20470/2        |
| Types  | 48/1            | 48/1           |

## Performance comparison

With the singleton pattern, if the appropriate object already has been created, the function takes 5 microseconds, rather than 26 milliseconds.

All time values shown below are in seconds (or in minutes:seconds if it has ":" ).

For test generation, we got the following performance results.

|     | QBFT | SSV  | Types |
|-----|------|------|-------|
| Old | 15   | 7:36 | 3     |
| New | 5    | 20   | 3     |

For test execution, we got the following performance results.

|     | QBFT | SSV  | Types |
|-----|------|------|-------|
| Old | 16   | 1:22 | 3     |
| New | 4    | 5    | 2     |

### DIFF
--- a/types/testingutils/keymanager.go
+++ b/types/testingutils/keymanager.go
@@ -46,10 +46,11 @@ func NewTestingKeyManager() *testingKeyManager {
 
 func NewTestingKeyManagerWithSlashableRoots(slashableDataRoots [][]byte) *testingKeyManager {
 
+	hash := getHash(slashableDataRoots)
+
 	mu.Lock()
 	defer mu.Unlock()
 
-	hash := getHash(slashableDataRoots)
 	if instance, ok := instancesMap[hash]; ok {
 		return instance
 	}
--- a/types/testingutils/keymanager.go
+++ b/types/testingutils/keymanager.go
@@ -4,8 +4,10 @@ import (
 	"bytes"
 	"crypto/ecdsa"
 	"crypto/rsa"
+	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"sync"
 
 	spec "github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/bloxapp/ssv-spec/types"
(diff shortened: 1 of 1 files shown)


types/testingutils/keymanager.go
```diff
@@ -25,11 +27,33 @@ type testingKeyManager struct {
 	slashableDataRoots [][]byte
 }
 
+var (
+	instancesMap = make(map[string]*testingKeyManager)
+	mu           sync.Mutex
+)
+
+func getHash(data [][]byte) string {
+	hasher := sha256.New()
```
GalRogozinski, line 36: you can consider using MuHash or something else which is faster

GalRogozinski (approved):
Great!

types/testingutils/keymanager.go
y0sher: fnv is easy fast and simple for tasks like this. built in the go stdlib

types/testingutils/keymanager.go
```diff
@@ -25,11 +27,34 @@ type testingKeyManager struct {
 	slashableDataRoots [][]byte
 }
 
+var (
+	instancesMap = make(map[string]*testingKeyManager)
+	mu           sync.Mutex
+)
```
y0sher, line 33: generally I would say global state is a bad idea, but since this is a test generation code and you are protecting it properly, its seems ok to me.

y0sher (approved):

PR #362, opened 2024-02-01
Revert changes in HashTreeRoot and SSZ encoding for backwards-compatibility

GalRogozinski:
## Description

`fastssz` was upgraded in #324 and it changed the roots for `ConsensusData`, `SignedMessage` and `SSVMessage`, which would be a fork.

This `fastssz` upgrade included a [crucial hashing fix](https://github.com/ferranbt/fastssz/pull/120/files#diff-856514f08a40f56b88f2b5c96750cdc308c1cee9c426c71ad8e9a298a5849c99), replacing `PutBytes` with `Append` in order to prevent double-merkleization in `HashTreeRoot` methods, thereby changing roots for `[]byte` fields exceeding 32 bytes.

It's noted that Beacon spec tests appear unaffected by this change, as they don't contain fields exceeding the threshold for the double-merkleization bug. However, we don't know that Beacon messages are absolutely not impacted.

Therefore, we propose inspecting the changes in `_encoding.go` files between the currently and previously imported versions of go-eth2-client, to ensure that any change from `PutBytes` to `Append` is encoding a field which can not exceed 32 bytes in size.

Two more changes were done to preserve backwards-compatibility:
- Encoding of `types.Contribution` was copied over from `v0.3.4`
- `spec.DataVersion` is now being decremented in encodings and incremented in decodings, because all of the values were incremented by 1 in go-eth2-client.


MatheusFranco99 (approved):
Good work!
Nice description

moshe-blox (approved):

Issue #326, opened 2023-12-24
Optimization Suggestion: Reduce Redundant Calls to SignedMessage.Validate in QBFT Message Processing


olegshmuelov:
**Details**
The method SignedMessage.Validate is being called multiple times across different functions, which may lead to inefficiencies in processing.
The method `SignedMessage.Validate` is redundantly invoked in several places:

1. Within the `Instance.BaseMsgValidation` method:
```go
// Instance.BaseMsgValidation method
if err := msg.Validate(); err != nil {
    return errors.Wrap(err, "invalid signed message")
}
```

2. Also in other functions such as `isValidProposal`, `validSignedPrepareForHeightRoundAndRoot`, `validateCommit`, and `validRoundChangeForData`. For example, in `isValidProposal`:
```go
// isValidProposal function
if err := signedProposal.Validate(); err != nil {
    return errors.Wrap(err, "proposal invalid")
}
```

**Proposed Solution**
We can call `SignedMessage.Validate` only once at the beginning to avoid repeating it.

PR #366, opened 2024-03-01
Reduce redundant validation calls

MatheusFranco99:
## Drop redundant calls

`SignedMessage.Validate()` is called multiple times for the same message, as pointed by https://github.com/bloxapp/ssv-spec/issues/326. This PR aims to solve this duplication.

The current flow is shown below.

current_flow

Notice that `isValidProposal` calls `validRoundChangeForData` and `validSignedPrepareForHeightRoundAndRoot` due to the justification messages.

If we want to keep the higher level validation calls we could get the following

eliminate_low_level

The problem is that the `Prepare` and `Round-Change` messages would still do double validation. And the checks can't be dropped due to the `isValidProposal` function.

If we were to keep the lower level calls, we would get

eliminate_higher_level

This produces no duplicated calls but, from a design point of view, it's bad to do a light validation check late.

Another option is to change the `Validation` function to also validate nested messages (which I think is best), as below.

new_validate

---

Closes https://github.com/bloxapp/ssv-spec/issues/326.

### DIFF
--- a/qbft/decided.go
+++ b/qbft/decided.go
@@ -2,6 +2,7 @@ package qbft
 
 import (
 	"bytes"
+
 	"github.com/bloxapp/ssv-spec/types"
 	"github.com/pkg/errors"
 )
@@ -66,10 +67,6 @@ func ValidateDecided(
 
 	if err := baseCommitValidation(config, signedDecided, signedDecided.Message.Height, share.Committee); err != nil {
 		return errors.Wrap(err, "invalid decided msg")
-	}
-
-	if err := signedDecided.Validate(); err != nil {
-		return errors.Wrap(err, "invalid decided")
 	}
 
 	r, err := HashDataRoot(signedDecided.FullData)


y0sher:
@MatheusFranco99 I agree with you on this that last option is the best. I think another option exists where we only validate the top message. but then when we call `Validate` on underlying message we are only validating the underlying message.
I think the point is not to validate same data more than once. but also a nice to have property is that we do not validate data that we don't need to. 
for example if we deem the message as bad and not go forward after first validate and some other checks, in the last option you mentioned we might be validating internal messages without needing to.

GalRogozinski:
Is there some overlap with #358 ?

MatheusFranco99:
Perhaps, there's overlap but their goals are a bit different. Here the goal is not to verify messages already verified (in proposal construction). In #358 , the goal was to avoid calling SignedSSVmessage.Validate() many times when processing a message

Issue #325, opened 2023-12-24
Optimization Proposal: Eliminate Duplicate Validation Check to Improve Consensus Time


olegshmuelov:
**Description**
There appears to be a redundancy in the validation of consensus data, resulting in the same checks being conducted twice. This redundancy could be unnecessarily extending the time taken for consensus.

**Details**
The validation check for the input data (byts) is performed twice:
1. In the `BaseRunner.decide` method:
```go
// BaseRunner.decide method
if err := runner.GetValCheckF()(byts); err != nil {
    return errors.Wrap(err, "input data invalid")
}
```
2. In the `Controller.StartNewInstance` method:
```go
// Controller.StartNewInstance method
if err := c.GetConfig().GetValueCheckF()(value); err != nil {
    return errors.Wrap(err, "value invalid")
}
```
Given that `BaseRunner.decide` calls `Controller.StartNewInstance`, this results in the same validation check being performed twice on the same data.

**Suggestion**
To optimize the consensus time, I propose removing one of these duplicate checks. My recommendation would be to keep the validation in the `Controller.StartNewInstance` method, as it seems more contextually appropriate for the responsibility of starting a new QBFT instance.

GalRogozinski:
`decide` calls `StartNewInstance` so maybe we should do this check as soon as possible.
Also som "duties" may not have an instance

PR #367, opened 2024-03-01
Drop duplicated value check call

MatheusFranco99:
Drop duplicated value check call as pointed by https://github.com/bloxapp/ssv-spec/issues/325.

Closes https://github.com/bloxapp/ssv-spec/issues/325.

### DIFF
--- a/qbft/controller.go
+++ b/qbft/controller.go
@@ -37,9 +37,6 @@ func NewController(
 
 // StartNewInstance will start a new QBFT instance, if can't will return error
 func (c *Controller) StartNewInstance(height Height, value []byte) error {
-	if err := c.GetConfig().GetValueCheckF()(value); err != nil {
-		return errors.Wrap(err, "value invalid")
-	}
 
 	// can't use <= because of height == 0 case
 	if height < c.Height {


MatheusFranco99:
@GalRogozinski Should we add a check inside `StartNewInstance` just to check if the value is not empty?

GalRogozinski:
@MatheusFranco99 
for your question... shouldn't this check be on decided?

MatheusFranco99:
@GalRogozinski 
What do you mean by "on decided"? Once the instance decides?
Btw, should this PR be into main or dev?

GalRogozinski:
@MatheusFranco99 
Sorry was being super unclear.
I meant that the check you offered should be on `decide()` the function that calls `StartNewInstance`

MatheusFranco99:
@GalRogozinski 
Oh yeah, I agree

MatheusFranco99:
Actually, thinking about it again, we would have something like:
```go
func (b *BaseRunner) decide(runner Runner, input *types.ConsensusData) error {
	byts, err := input.Encode()
	if err != nil {
		return errors.Wrap(err, "could not encode ConsensusData")
	}
        // New check
	if len(byts) == 0 {
		return errors.Wrap(err, "...")
	}

	if err := runner.GetValCheckF()(byts); err != nil {
		return errors.Wrap(err, "input data invalid")
	}
	// ...
```
But `runner.GetValCheckF` should already cover it. What do you think?

GalRogozinski:
@MatheusFranco99 
For example on `BeaconVoteValueCheckF` I don't see an empty check. I see we do a `Decode` that I am not sure will fail (maybe it returns empty struct, unclear from the code and spec should be clear). It will of course fail but with some other error message.

You have a point though, that it is nice that this will be inside the value check. But this probably means that you need to add the check in each function.

From a SW engineer perspective I would make a base function wrapper that does checks that should happen across all value checks and then calls an internal `ValueCheck` function.  But I don't know how readable it is for spec.

So for spec I would either use your snippet or add the empty check in each `valueCheckF`. You can choose. The first one won't translate the check to the postconsensus stage, but this is fine since we should never have consensus on an empty value.

GalRogozinski (commented):
Just a question: we are still covered test wise for empty values?

PR #369, opened 2024-03-21
Partial Signature Verification Aggregation

MatheusFranco99:
# Overview

This PR implements the reduction of signatures optimization for partial signature messages as described in this [SIP](https://github.com/bloxapp/SIPs/pull/46).

## Changes
- Drop `SignedPartialSignatureMessage`'s signature verification.
- Stop verifying each beacon signature in favour of verifying the quorum of signatures through the reconstructed signature.
- Add fall back function in case the reconstructed signature is wrong.
- In case of duplicated different signatures (same signer), verify both and keep the correct one (if any).

## Changed tests
- [x] Unknown signer: error message
- [x] Invalid beacon signature: add a quorum of messages to detect the invalid signature
- [x] Invalid signature: drop expected error regarding `SignedPartialSignatureMessage`'s signature

## New tests
- [x] Invalid quorum then valid quorum: produces an invalid quorum and then sends a fourth valid message to get a valid quorum

## Improvement

New **total duty time** per number of consensus rounds.

| 1 Round | 2 Rounds | 3 Rounds |
| ------- | -------- | -------- |
| 50%     | 80%      | 88%      |

New **partial signature quorum** processing time: 17%.


ssv/runner_signatures.go
```diff
@@ -45,35 +45,34 @@ func (b *BaseRunner) signPostConsensusMsg(runner Runner, msg *types.PartialSigna
...
+	signerInCommittee := false
+	for _, operator := range b.Share.Committee {
+		if operator.OperatorID == signedMsg.Signer {
+			signerInCommittee = true
+			break
 		}
 	}
+	if !signerInCommittee {
+		return errors.New("unknown signer")
+	}
```
GalRogozinski, line 70: hmmm

This is currently checked by MsgValidation `ERR_PSIG_INVALID_SIG_ID`?
Since MsgValidation isn't part of spec it is fine that you added this..

But now I made up my mind that MsgValidation should be an inseperable part of spec.
Once we add tests for this, we won't have to do those double checks

types/messages_encoding.go
```diff
@@ -122,3 +122,153 @@ func (s *SSVMessage) HashTreeRootWith(hh ssz.HashWalker) (err error) {
...
 	return ssz.ProofTree(s)
 }
+
+// MarshalSSZ ssz marshals the SignedSSVMessage object
+func (s *SignedSSVMessage) MarshalSSZ() ([]byte, error) {
+	return ssz.MarshalSSZ(s)
+}
+
+// MarshalSSZTo ssz marshals the SignedSSVMessage object to a target array
+func (s *SignedSSVMessage) MarshalSSZTo(buf []byte) (dst []byte, err error) {
```
GalRogozinski, line 132: ok...
So we need to decide how to do this...

In #345 we added the `SignedSSVMessage` but we merged it to the fork branch because of this very encoding change...
I made this decision because I decided that the ssz encoding is anyhow the correct (consistent) way to go and I didn't take into account that this PR would need this.

I think we should rework #345 to remove the ssz encoding and add flat encoding so no fork

ssv/spectest/tests/runner/postconsensus/invalid_msg_sig.go
```diff
@@ -9,10 +9,9 @@ import (
 	"github.com/bloxapp/ssv-spec/types/testingutils"
 )
 
-// InvalidMessageSignature tests PartialSignatureMessage signature invalid
+// InvalidMessageSignature tests PartialSignatureMessage signature invalid. No error is generated since the SignedPartialSignatureMessage.Signature is no longer checked
```
GalRogozinski, line 12:
```suggestion
// InvalidMessageSignature tests PartialSignatureMessage signature invalid. No error is generated since the SignedPartialSignatureMessage. Signature is no longer checked. Wait for quorum before verifying.
```

ssv/spectest/tests/runner/postconsensus/invalid_quorum_then_valid_quorum_sc.go
```diff
@@ -0,0 +1,288 @@
...
+	"github.com/bloxapp/ssv-spec/ssv"
+	ssvcomparable "github.com/bloxapp/ssv-spec/ssv/spectest/comparable"
+	"github.com/bloxapp/ssv-spec/types"
+	"github.com/bloxapp/ssv-spec/types/testingutils"
+	"github.com/bloxapp/ssv-spec/types/testingutils/comparable"
+)
+
+// invalidThenQuorumSyncCommitteeContributionSC returns state comparison object for the invalid then quorum SyncCommitteeContribution versioned spec test
+// runner should finish since quorum is achieved
+func invalidQuorumThenValidQuorumSyncCommitteeContributionSC() *comparable.StateComparison {
```
GalRogozinski, line 16: Just a friendly reminder that SC files were replaced by generated jsons.
However if you think this is a core test (looks like one) and that coding out the SC helps in verifying then this is great!!

ssv/spectest/tests/runner/postconsensus/invalid_then_quorum.go
```diff
@@ -14,7 +14,6 @@ import (
 // InvalidThenQuorum  tests a runner receiving an invalid message then a valid quorum, terminating successfully
 func InvalidThenQuorum() tests.SpecTest {
```
GalRogozinski, line 15: What's the difference between this and `InvalidQuorumThenQuorum`?

ssv/spectest/tests/runner/preconsensus/invalid_msg_sig.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/bloxapp/ssv-spec/types/testingutils"
 )
 
-// InvalidMessageSignature tests SignedPartialSignatureMessage signature invalid
+// InvalidMessageSignature tests SignedPartialSignatureMessage signature invalid. No error is generated since the SignedPartialSignatureMessage.Signature is no longer checked
```
GalRogozinski, line 11:
```suggestion
// InvalidMessageSignature tests SignedPartialSignatureMessage signature invalid. No error is generated since the SignedPartialSignatureMessage. Signature is no longer checked.
// We wait for quorum to verify sig.
```

ssv/spectest/tests/runner/preconsensus/invalid_then_quorum.go
```diff
@@ -32,7 +32,6 @@ func InvalidThenQuorum() tests.SpecTest {
 				OutputMessages: []*types.SignedPartialSignatureMessage{
 					testingutils.PreConsensusContributionProofMsg(ks.Shares[1], ks.Shares[1], 1, 1), // broadcasts when starting a new duty
 				},
-				ExpectedError: "failed processing sync committee selection proof message: invalid pre-consensus message: could not verify Beacon partial Signature: wrong signature",
```
GalRogozinski, line 35: shouldn't the error happen on the first quorum?

GalRogozinski (changes requested):
What about QBFT messages?

ssv/runner_signatures.go
MatheusFranco99: I agree that message validation should belong to the spec

types/messages_encoding.go
MatheusFranco99: Removed `SignedSSVMessage` from this PR.

ssv/spectest/tests/runner/postconsensus/invalid_msg_sig.go
MatheusFranco99: The "no space" is intended. I wanted it to be: "SignedPartialSignatureMessage. Signature"

ssv/spectest/tests/runner/preconsensus/invalid_msg_sig.go
MatheusFranco99: The "no space" is intended. I wanted it to be: "SignedPartialSignatureMessage. Signature"

ssv/spectest/tests/runner/preconsensus/invalid_then_quorum.go
MatheusFranco99: The first invalid signature is overwritten by the correct one before a quorum is reached. Once the quorum is reached, all signatures stored are correct and no error is created

MatheusFranco99:
@GalRogozinski This PR is for partial signature messages

GalRogozinski (approved):
Looks good. We also need the `SignedSSVMessage` with no encoding change.

ssv/runner_validations.go
```diff
@@ -26,6 +27,18 @@ func (b *BaseRunner) ValidatePreConsensusMsg(runner Runner, signedMsg *types.Sig
...
+// Verify each signature in container removing the invalid ones
+func (b *BaseRunner) FallBackAndVerifyEachSignature(container *PartialSigContainer, root [32]byte) {
+
+	signatures := container.GetSignatures(root)
+
+	for operatorID, signature := range signatures {
+		if err := b.verifyBeaconPartialSignature(operatorID, signature, root); err != nil {
+			container.Remove(operatorID, root)
+		}
+	}
```
y0sher, line 39: shouldn't we do another check here? or after calling FallBackAndVerifyEachSignature, what if we had a quorum of 4, then we realized one is invalid, don't we want to reconstruct again with the 3 left? 
or this is not possible since its a must we'll try first when have 3?

Issue #370, opened 2024-03-24
In ControllerSpecTest, OutputMessages should be used


GalRogozinski:
## Description

Broadcasted messages are not checked

PR #371, opened 2024-03-24
Do not pass value to instance

### LINKED ISSUE #289: Improve RC clock sync
### Description 

There may an issue where different nodes start the consensus process at different times. Causing some nodes to send RC messages earlier than others. This just bloats the number of messages send over the network and may delay consensus.

GalRogozinski:
## Description
Closes #289 

Instance will now fetch consensus data only if it is the round leader.
Timer starts before data is fetched to ensure it is synchronized across committee nodes.


GalRogozinski:
#370 was opened while working on this

qbft/instance.go
```diff
@@ -121,7 +132,7 @@ func (i *Instance) ProcessMsg(msg *SignedMessage) (decided bool, decidedValue []
 			}
 			return err
 		case RoundChangeMsgType:
-			return i.uponRoundChange(i.StartValue, msg, i.State.RoundChangeContainer, i.config.GetValueCheckF())
+			return i.uponRoundChange(msg, i.State.RoundChangeContainer, i.config.GetValueCheckF())
```
olegshmuelov, line 135: I noticed we're passing `i.StartValue`, `i.State.RoundChangeContainer`, and `i.config.GetValueCheckF()` as arguments to the function. Given these are accessed via the instance `i`, could someone help clarify the rationale behind explicitly passing these as parameters instead of accessing them directly within the uponRoundChange method?

Is there a specific design choice or benefit we're aiming for by structuring the call this way? For instance, does this approach enhance testability, readability, or separation of concerns in a way that accessing them directly within the function would not?

olegshmuelov, line 135:
```go
case CommitMsgType:
	decided, decidedValue, aggregatedCommit, err = i.UponCommit(msg, i.State.CommitContainer)
	if decided {
		i.State.Decided = decided
		i.State.DecidedValue = decidedValue
	}
	return err
```
not related to the change, but we should check for err before assign

qbft/instance.go
```diff
@@ -24,6 +24,7 @@ type Instance struct {
 	// forceStop will force stop the instance if set to true
 	forceStop  bool
 	StartValue []byte
+	CdFetcher  *types.DataFetcher `json:"-"`
```
olegshmuelov, line 27: Could we rename CdFetcher to CDFetcher for clarity? 'CD' stands for 'Consensus Data', and capitalizing the acronym could improve readability and align with our naming conventions.

qbft/instance.go
olegshmuelov: ditto
`i.uponProposal`
`i.uponPrepare`
`i.UponCommit`

ssv/attester.go
```diff
@@ -172,28 +172,37 @@ func (r *AttesterRunner) expectedPostConsensusRootsAndDomain() ([]ssz.HashRoot,
...
+	attestationFetcher := attestationFetcher(r, duty)
 
-	input := &types.ConsensusData{
-		Duty:    *duty,
-		Version: ver,
-		DataSSZ: attDataByts,
-	}
-
-	if err := r.BaseRunner.decide(r, input); err != nil {
+	if err := r.BaseRunner.decide(r, attestationFetcher); err != nil {
```
olegshmuelov, line 177: Why are we passing r to r.BaseRunner.decide? Can't it access r internally?

olegshmuelov: same for 
`aggregator`
`proposer`
and more

qbft/spectest/tests/startinstance/empty_value.go
```diff
@@ -5,6 +5,7 @@ import (
 )
 
 // EmptyValue tests a starting an instance for an empty value (not passing value check)
+// Instance starts but with an empty value and no proposal gets broadcasted
```
olegshmuelov, line 8: Is there a way no proposals gets broadcasted?

qbft/spectest/tests/startinstance/invalid_value.go
```diff
@@ -6,6 +6,7 @@ import (
 )
 
 // InvalidValue tests a starting an instance for an invalid value (not passing value check)
+// Instance starts but with an empty value and no proposal gets broadcasted
```
olegshmuelov, line 9: ditto

qbft/spectest/tests/startinstance/nil_value.go
```diff
@@ -5,6 +5,7 @@ import (
 )
 
 // NilValue tests a starting an instance for a nil value (not passing value check)
+// Instance starts but with an empty value and no proposal gets broadcasted
```
olegshmuelov, line 8: ditto

ssv/attester.go
MatheusFranco99: It seems to me that the BaseRunner can't access the runner internally. But, also, the `decide` function doesn't need the runner as far as I understand. So, I think this could be simplified to `r.BaseRunner.decide(fetcher)` too

qbft/round_change.go
```diff
@@ -120,66 +119,81 @@ func hasReceivedProposalJustificationForLeadingRound(
...
+		valueToPropose = roundChangeMsg.FullData
+	} else {
+		cd, err := cdFetcher.GetConsensusData()
+		if err != nil {
+			return nil, err
+		}
+		valueToPropose = cd
+	}
+	return valueToPropose, nil
 }
```
MatheusFranco99, line 197: I believe this function can be called more than once during the Round-Change phase. If so, we would also wait for the Beacon Node more than once. Is there a way to save the data, e.g. storing it in the `i.StartValue`, so it can be returned faster?

MatheusFranco99 (approved):
Good work!! Just left a suggestion

qbft/round_change.go
GalRogozinski: The idea of `cdFetcher` is that implementation can do it in a way where it calls it asynchronously and even caches values. But this is an implementation detail.

@olegshmuelov

qbft/instance.go
MatheusFranco99: I agree. I even prefer the whole `ConsensusDataFetcher`, or simply `DataFetcher`.

qbft/instance.go
```diff
@@ -54,16 +55,26 @@ func (i *Instance) ForceStop() {
...
 		// propose if this node is the proposer
 		if proposer(i.State, i.GetConfig(), FirstRound) == i.State.Share.OperatorID {
+			value, err := cdFetcher.GetConsensusData()
+			if err != nil {
+				fmt.Printf("%s\n", err.Error())
+				return
+			}
+			if err = valueCheckF(value); err != nil {
+				fmt.Printf("%s\n", err.Error())
+				return
```
nkryuchkov, line 75: should we also return if CreateProposal returns an error?

types/consensus_data.go
```diff
@@ -46,6 +46,12 @@ func (c *Contributions) HashTreeRootWith(hh ssz.HashWalker) error {
 	return nil
 }
 
+// DataFetcher asynchronusly fetches data from the beacon node upon instantiation
```
nkryuchkov, line 49:
```suggestion
// DataFetcher asynchronously fetches data from the beacon node upon instantiation
```

types/consensus_data.go
```diff
@@ -46,6 +46,12 @@ func (c *Contributions) HashTreeRootWith(hh ssz.HashWalker) error {
...
+type DataFetcher struct {
```
nkryuchkov, line 50: Why not interface? Using concrete types is not flexible and inconvenient for testing and when spec repo is used by other repositories

qbft/instance.go
```diff
@@ -56,16 +57,26 @@ func (i *Instance) ForceStop() {
 }
 
 // Start is an interface implementation
-func (i *Instance) Start(value []byte, height Height) {
+func (i *Instance) Start(cdFetcher *types.DataFetcher, height Height, valueCheckF ProposedValueCheckF) {
```
MatheusFranco99, line 60: Can't we use the `i.GetConfig().GetValueCheck()` instead of using an argument? It seems a duplication to me

qbft/instance.go
```diff
@@ -56,16 +57,26 @@ func (i *Instance) ForceStop() {
...
 		i.State.Height = height
 
 		i.config.GetTimer().TimeoutForRound(FirstRound)
 
 		// propose if this node is the proposer
 		if proposer(i.State, i.GetConfig(), FirstRound) == i.State.CommitteeMember.OperatorID {
+			value, err := cdFetcher.GetConsensusData()
+			if err != nil {
+				fmt.Printf("%s\n", err.Error())
+				return
```
MatheusFranco99, line 73: The error should be returned and not printed, no?

qbft/round_change.go
```diff
@@ -154,58 +144,84 @@ func hasReceivedProposalJustificationForLeadingRound(
...
+	if roundChangeMsg.RoundChangePrepared() {
+		valueToPropose = preparedValue
+	} else {
+		cd, err := cdFetcher.GetConsensusData()
+		if err != nil {
+			return nil, err
+		}
+		valueToPropose = cd
+	}
+	return valueToPropose, nil
```
MatheusFranco99, line 224: Since the CDFetcher may update its value, we need to make sure that two different proposals for the same round can't be done. One way to do that is to check, in the validation of Round-Change messages, that the `msg.Round` is **strictly** above our current round.

GalRogozinski: I don't understand.. we call this once per round

qbft/round_change.go
```diff
@@ -154,58 +144,84 @@ func hasReceivedProposalJustificationForLeadingRound(
...
-	roundChangeJustifications []*ProcessingMessage,
-	value []byte,
+	roundChangeJustifications []*ProcessingMessage, cdFetcher *types.DataFetcher,
 	valCheck ProposedValueCheckF,
 	newRound Round,
-) error {
+) ([]byte, error) {
+	if proposer(state, config, roundChangeMsg.QBFTMessage.Round) != state.CommitteeMember.OperatorID {
+		return nil, errors.New("not proposer")
+	}
```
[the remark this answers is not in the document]
GalRogozinski: where? I don't see it

qbft/instance.go
GalRogozinski: apparently we printed errors in main before...
not sure why...
Maybe we should add errors and test them?
I will open an issue

GalRogozinski: #460

qbft/instance.go
y0sher: Lets to this, I prefer `DataFetcher` or `Fetcher`. anyway if using acronym always capital letters for all words.

qbft/instance.go
y0sher: I agree, if value check only used here then maybe we don't need it as part of `Config`.

qbft/controller.go
```diff
@@ -33,10 +33,7 @@ func NewController(identifier []byte, committeeMember *types.CommitteeMember, co
 }
 
 // StartNewInstance will start a new QBFT instance, if can't will return error
-func (c *Controller) StartNewInstance(height Height, value []byte) error {
-	if err := c.GetConfig().GetValueCheckF()(value); err != nil {
-		return errors.Wrap(err, "value invalid")
-	}
+func (c *Controller) StartNewInstance(height Height, cdFetcher *types.DataFetcher) error {
```
y0sher, line 36: Why not move the fetching logic outside of controller and pass an already fetched value? this of course also means determining that we are leaders outside of the controller. IMO it makes more sense than passing a fetcher. less changes to controller.

qbft/instance.go
y0sher: @GalRogozinski I agree with this, either only pass and not save (if its only used within these function) or save and don't pass.

qbft/controller.go
y0sher: thinking again this might not work if we need to fetch data after round change? but probably we should fetch data at the beginning of the slot regardless.

qbft/round_change.go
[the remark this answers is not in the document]
y0sher: 196-197

qbft/round_change.go
```diff
@@ -154,58 +144,84 @@ func hasReceivedProposalJustificationForLeadingRound(
...
 	}
 
-	return nil
+	return valueToPropose, nil
+}
+
+// chooseValueToPropose
+// If justifiedRoundChangeMsg has no prepare justification choose state value
+// If justifiedRoundChangeMsg has prepare justification choose prepared value
+func chooseValueToPropose(roundChangeMsg *Message, cdFetcher *types.DataFetcher, preparedValue []byte) ([]byte, error) {
```
y0sher, line 213: since its only used in one place, this shouldn't be a method IMO, or at least the name should reflect that a long-awaited fetch is part of this function.

qbft/spectest/tests/startinstance/empty_value.go
y0sher: @olegshmuelov I assume if CDFetcher fails fetching..

types/consensus_data.go
y0sher: @nkryuchkov @GalRogozinski 
Usually I support using concrete types. As they are more predictable than interfaces. but this time you are using this concrete type as an interface so it might make more sense for it to be an interface ( but then also moved to the runners ).
the rule is -
Return concrete types
Receive interfaces

y0sher (changes requested):
you think its possible to do the fetching outside of the qbft controller?

qbft/instance.go
GalRogozinski: this is out of scope of PR, but can be done

iurii-ssv:
@GalRogozinski (if/once you get back to this code) could you clarify/confirm whether this PR (after finished and merged) will support and be 100% compatible with the solution for the issue I [described here](https://github.com/ssvlabs/ssv/issues/1825#issuecomment-2445279035) (which we'll probably want to address at some point) ?

I haven't looked through the code of this PR too much yet, but from cursory overview I think
- it couples/bundles the start of QBFT instance and the call to fetch starting data (when leader)
- while the more flexible/unassuming approach (for the case when we are QBFT leader especially, but also in general) would be to 1) start QBFT instance with non-blocking `Instance.Start` call and let it finish all the setup it needs for `Instance` to become functional in terms of being able to communicate with other instance in the cluster (and participate round-changes for example) 2) and make it such that `Instance` can just "get injected" with `StartValue` without knowing when/whether/whom by it will happen (ofc this way there will be a new set of conditions to handle - when `StartValue` is needed but is not there yet)

lets consider the following scenario:
- we have block proposer duty, lets say we found a way to start all QBFT instances for the relevant cluster "roughly at the same time" (which btw current SSV node implementation doesn't do as my comment describes it  https://github.com/ssvlabs/ssv/issues/1825#issuecomment-2445279035)
- to make it concrete, lets say cluster is 4 nodes for this example, all 4 have now started counting time to the first round timeout waiting for leader to fetch his data (which he is in the process of)
- leader couldn't fetch his data, and even worse - he is synchronously blocked on a call to Beacon node
- now 3 nodes start round 1 change, and stuck leader of round 1 can't further participate in QBFT for this cluster because he is just stuck (his whole duty is blocked/stuck btw waiting for `Instance.Start` func to return) - all because of QBFT instance took it upon itself to go fetch the data
- and `Instance` can't really recover from this type of scenario (even if `DataFetcher` function has timeout/retry logic) to start/keep participating in QBFT because technically `Instance` has never started and is thus useless to QBFT cluster

GalRogozinski:
Hey @iurii-ssv,

Yeah, this PR is suppose to allow making async calls.
We stopped working on it due to Alan fork... which is a shame in my opinion because I think this change should be part of a fork so all nodes have similar behavior

So currently we need to wait for the next fork

PR #372, opened 2024-03-27
Reduce signatures - drop redundant BLS in QBFT (no fork)

MatheusFranco99:
# Overview

This PR drops redundant BLS signature checks in QBFT.

## Changes
- [x] Drop redundant checks for Proposal, Prepare, Commit and Round-Change messages
- [x] Keep checks for message justification and decided messages
- [x] Add a check to ensure that the signers belong to the committee


qbft/prepare.go
```diff
@@ -108,8 +111,14 @@ func validSignedPrepareForHeightRoundAndRoot(
 		return errors.New("msg allows 1 signer")
 	}
 
-	if err := signedPrepare.Signature.VerifyByOperators(signedPrepare, config.GetSignatureDomainType(), types.QBFTSignatureType, operators); err != nil {
-		return errors.Wrap(err, "msg signature invalid")
+	if !signedPrepare.CheckSignersInCommittee(operators) {
+		return errors.New("signers not in committee")
+	}
```
GalRogozinski, line 116: just from sw engineering point of view
shouldn't we have some `baseValidation` that keeps on calling this

so we never forget to add this?
because it repeats itself

GalRogozinski (approved):

GalRogozinski (approved):
Approved but you can still fix my comment

qbft/round_change.go
```diff
@@ -213,6 +214,7 @@ func validRoundChangeForData(
 	height Height,
 	round Round,
 	fullData []byte,
+	verifySignature bool,
```
moshe-blox, line 217: switch to dontVerifySignature?

moshe-blox: on another thought, maybe we should go for the validateX() and validateXNoVerify() pattern?

qbft/commit.go
```diff
@@ -127,7 +135,7 @@ func validateCommit(
 	proposedMsg *SignedMessage,
 	operators []*types.Operator,
 ) error {
-	if err := baseCommitValidation(config, signedCommit, height, operators); err != nil {
+	if err := baseCommitValidation(config, signedCommit, height, operators, false); err != nil {
```
moshe-blox, line 138: if we go for the NoVerify suffix pattern, validateCommit could be validateCommitNoVerify to be explicit

moshe-blox (approved):

moshe-blox (changes requested):
Let's merge this after #345 is on main

qbft/round_change.go
MatheusFranco99: Agree

qbft/commit.go
MatheusFranco99: Agree

qbft/commit.go
```diff
@@ -111,9 +113,15 @@ func baseCommitValidation(
 		return errors.Wrap(err, "signed commit invalid")
 	}
 
+	if !signedCommit.CheckSignersInCommittee(operators) {
+		return errors.New("signers not in committee")
+	}
```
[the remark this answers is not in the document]
MatheusFranco99: I think it does. But I'm leaving it here because spec doesn't know and it's a light check

qbft/prepare.go
MatheusFranco99: I think we should totally do that. Not only for this check but for many others such as height, msg.Validate(), etc. Since it would require some bigger refactoring, let's do it in a future PR? Just so we can dispatch this to implementation as soon as possible and as simple as possible.

qbft/commit.go
```diff
@@ -94,7 +95,7 @@ func CreateCommit(state *State, config IConfig, root [32]byte) (*SignedMessage,
 	return signedMsg, nil
 }
 
-func baseCommitValidation(
+func baseCommitValidationNoVerification(
```
olegshmuelov, line 98: the config is not used here anymore, should we delete it?

qbft/prepare.go
```diff
@@ -79,7 +80,7 @@ func getRoundChangeJustification(state *State, config IConfig, prepareMsgContain
 
 // validSignedPrepareForHeightRoundAndRoot known in dafny spec as validSignedPrepareForHeightRoundAndDigest
 // https://entethalliance.github.io/client-spec/qbft_spec.html#dfn-qbftspecification
-func validSignedPrepareForHeightRoundAndRoot(
+func validSignedPrepareForHeightRoundAndRootNoVerification(
```
olegshmuelov, line 83: the config is not used here anymore, should we delete it?

MatheusFranco99: Indeed. I'll remove it. Thanks

qbft/commit.go
MatheusFranco99: Indeed. I'll remove it. Thanks

qbft/round_change.go
```diff
@@ -227,12 +228,12 @@ func validRoundChangeForData(
 		return errors.New("msg allows 1 signer")
 	}
 
-	if err := signedMsg.Signature.VerifyByOperators(signedMsg, config.GetSignatureDomainType(), types.QBFTSignatureType, state.Share.Committee); err != nil {
-		return errors.Wrap(err, "msg signature invalid")
+	if err := signedMsg.Validate(); err != nil {
+		return errors.Wrap(err, "roundChange invalid")
```
olegshmuelov, line 232: just to understand, it was a bug that we validated only the inner message?

MatheusFranco99: I think so

olegshmuelov: if so, we have to add a test for `message signers is empty`, `non unique signer` & `signer ID 0 not allowed`

MatheusFranco99: Yup, adding these tests.
Btw, the previous version couldn't produce any errors because, even though it should be `signedMsg.Validate()` and not `signedMsg.Message.Validate()`, the other checks were also verifying the conditions. But, of course, still better to call `signedMsg.Validate()`

olegshmuelov (approved):

qbft/commit.go
```diff
@@ -127,7 +145,7 @@ func validateCommit(
 	proposedMsg *SignedMessage,
 	operators []*types.Operator,
 ) error {
-	if err := baseCommitValidation(config, signedCommit, height, operators); err != nil {
+	if err := baseCommitValidationNoVerification(signedCommit, height, operators); err != nil {
```
olegshmuelov, line 148: one more cleanup can be done here
to not pass the `config` to validateCommit

olegshmuelov:
since https://github.com/bloxapp/ssv-spec/pull/342 was merged already


can we fix small [issue](https://github.com/bloxapp/ssv-spec/pull/342#discussion_r1542515196) in the current pr?

GalRogozinski:
> since #342 was merged already
> 
> can we fix small [issue](https://github.com/bloxapp/ssv-spec/pull/342#discussion_r1542515196) in the current pr?

This PR doesn't change runner at all...
Maybe in a subsequent PR when we change runner

olegshmuelov:
> > since #342 was merged already
> > can we fix small [issue](https://github.com/bloxapp/ssv-spec/pull/342#discussion_r1542515196) in the current pr?
> 
> This PR doesn't change runner at all... Maybe in a subsequent PR when we change runner

created an issue
https://github.com/bloxapp/ssv-spec/issues/376

types/operator.go
```diff
@@ -5,13 +5,19 @@ type OperatorID = uint64
 
 // Operator represents an SSV operator node
 type Operator struct {
-	OperatorID OperatorID
-	PubKey     []byte `ssz-size:"48"`
+	OperatorID    OperatorID
+	BeaconPubKey  []byte `ssz-size:"48"`
```
olegshmuelov, line 9: Can we consider calling it SharePubKey?

MatheusFranco99: Yes, better 👍

qbft/commit.go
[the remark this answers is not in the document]
y0sher: I think we should NEVER assume something happens in msg validation hence we don't need to take care in consensus.

GalRogozinski (changes requested):
I don't see new sig verification tests in `runner/consensus` pkg

qbft/commit.go
MatheusFranco99: Yeah. "IgnoreSignature" is more clear. I'll update it. Thanks!

qbft/round_change.go
MatheusFranco99: It is not a bug, but all other base validation functions for the other message types call validate on the upper message, and this is the only inconsistent one. Even though, all of these calls are indeed duplicated 😅

types/signer.go
```diff
@@ -35,6 +37,7 @@ type BeaconSigner interface {
 // SSVSigner used for all SSV specific signing
 type SSVSigner interface {
 	SignRoot(data Root, sigType SignatureType, pk []byte) (Signature, error)
+	SignNetworkData(data []byte, pk []byte) ([]byte, error)
```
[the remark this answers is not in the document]
MatheusFranco99: I find `OperatorData` confusing

types/testingutils/keymanager.go
```diff
@@ -22,6 +25,7 @@ type testingKeyManager struct {
 	keys           map[string]*bls.SecretKey
 	ecdsaKeys      map[string]*ecdsa.PrivateKey
 	encryptionKeys map[string]*rsa.PrivateKey
+	networkKeys    map[string]*rsa.PrivateKey
```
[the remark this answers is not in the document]
MatheusFranco99: I don't know what these `encryptionKeys` are for. It's never used in spec. Maybe some old deprecated functionality once used it and now it's a leftover.

I find the name `OperatorKey` confusing. The operator has many key types. I think we should include its purpose in the name, e.g. network, share, beacon, etc.

types/testingutils/keymanager.go
```diff
@@ -166,11 +154,30 @@ func (km *testingKeyManager) SignETHDepositRoot(root []byte, address common.Addr
 	panic("implement")
 }
 
+func (km *testingKeyManager) SignNetworkData(data []byte, pk []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
MatheusFranco99: See above comment

types/testingutils/keymanager.go
```diff
@@ -166,11 +154,30 @@ func (km *testingKeyManager) SignETHDepositRoot(root []byte, address common.Addr
...
+	}
+	return signature, nil
+}
+
 func (km *testingKeyManager) AddShare(shareKey *bls.SecretKey) error {
 	km.keys[hex.EncodeToString(shareKey.GetPublicKey().Serialize())] = shareKey
 	return nil
 }
 
+func (km *testingKeyManager) AddNetworkKey(sk *rsa.PrivateKey) error {
```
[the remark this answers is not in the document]
MatheusFranco99: See above comment

qbft/spectest/tests/commit/unknwon_signer.go
```diff
@@ -23,9 +23,9 @@ func UnknownSigner() tests.SpecTest {
 	return &tests.MsgProcessingSpecTest{
 		Name:          "unknown commit signer",
 		Pre:           pre,
-		PostRoot:      "976cd5cecd58bba892a38ec0ef02b3aed4656fb89fef473d8af78fedf095439d",
+		PostRoot:      "f92657c2f9b7225573bcd5b03a0f59dfda3f04c361e022e37357bf22c7a10ab9",
 		InputMessages: msgs,
-		ExpectedError: "invalid signed message: msg signature invalid: unknown signer",
+		ExpectedError: "invalid signed message: signers not in committee",
```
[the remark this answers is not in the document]
MatheusFranco99: Indeed. I'm going to update it. Thanks!

types/testingutils/keymanager.go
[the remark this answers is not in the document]
GalRogozinski: The encryption key is what is used to encrypt on chain validator keys.
It is the same key that we use for signing message.

Why is this part of spec? Well basically it depends on what is the scope of spec... and lately I've come to realize that different people at the company expect different things from spec... We wanted to remove it in #348... But @lior-blox didn't answer my question yet :-\

Conceptually there could be two different keys (for encryption and signing), and one can claim the implementation decides to use the same key for both operations...
So rethinking this, both having 2 keys or one key can pass my review, I won't fuss :-)

But I don't agree an operator has many key types (mostly 2)... The `shareKey` is linked to a validator.
Sure an operators hold validator so I can see why is this confusing.

But I really dislike `networkKey` especially that now in spec we decided to move authentication to `Validator`...
Maybe a name we can agree on is `SSVKey`?
It matches the `SSVSigner` interface. Also it signifies those keys are used specifically by the SSV protocol (unlike the SharedKeys that are Eth specific)

types/signer.go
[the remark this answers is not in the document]
GalRogozinski: Maybe `SSVData`...
I don't know I find `NetworkData` confusing as long as we don't put it in p2p package.
See https://github.com/bloxapp/ssv-spec/pull/372#discussion_r1550196788

types/operator.go
```diff
@@ -5,13 +5,19 @@ type OperatorID = uint64
...
+	NetworkPubKey []byte `ssz-size:"294"`
 }
 
-// GetPublicKey returns the public key with which the node is identified with
-func (n *Operator) GetPublicKey() []byte {
-	return n.PubKey
+// GetSharePublicKey returns the share public key
+func (n *Operator) GetSharePublicKey() []byte {
+	return n.SharePubKey
+}
```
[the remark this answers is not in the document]
GalRogozinski: > But I think this should be done in another PR.
Aite, too many diffs are bad at this point 👍 

> Even though spec may not care about memory, since the implementation is forced to be aligned to spec, I think we should care and implement it in the future PR.
Not important for me, but we can. If @moshe-blox will feel this is important then we will do it for sure.

types/signer.go
```diff
@@ -53,4 +56,6 @@ type KeyManager interface {
 	AddShare(shareKey *bls.SecretKey) error
 	// RemoveShare removes a share key
 	RemoveShare(pubKey string) error
+	// AddNetworkKey saves a network key
+	AddNetworkKey(sk *rsa.PrivateKey) error
```
GalRogozinski, line 60: So after discussing with @y0sher and @olegshmuelov

Get rid of the keyManager interface. `Share` functions should move to `BeaconSigner`. `Network` (rename to SSV) functions should go to SSVSigner

The `KeyManager` in testingutils can implement both to reduce changes (but you can do some other design if you wish)

GalRogozinski (changes requested):

GalRogozinski (changes requested):
Another point is that our Broadcasted messages should be signed.

I think that if you change `testingutils.Broadcast` it is good enough since it will change all the tests. It will also be like impl that currently put in p2p

types/testingutils/keymanager.go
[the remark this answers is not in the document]
MatheusFranco99: The same key is being used for both encryption and signature?

Wait, but the `SSVSigner` according to [comment](https://github.com/bloxapp/ssv-spec/pull/372#discussion_r1550144782) is already a name being used for the BLS shares sign.

GalRogozinski:
Also we need to maybe add an `RSAVerify` to MsgProcessing test on the broadcasted message to ensure that messages were signed correctly

GalRogozinski:
Also we need to drop the double verification at the p2p layer now...

ssv/attester.go
```diff
@@ -227,6 +230,10 @@ func (r *AttesterRunner) GetSigner() types.KeyManager {
 	return r.signer
 }
 
+func (r *AttesterRunner) GetOperatorSigner() types.SSVOperatorSigner {
+	return r.operatorSigner
+}
+
```
GalRogozinski, line 236: does runner use the signer?
in the end we use the network to sign it

MatheusFranco99: Indeed. Removing it. Thanks!

qbft/state.go
```diff
@@ -29,20 +31,26 @@ type IConfig interface {
...
+	Domain         types.DomainType
+	ValueCheckF    ProposedValueCheckF
+	ProposerF      ProposerF
+	Network        Network
+	Timer          Timer
 }
 
 // GetSigner returns a Signer instance
-func (c *Config) GetSigner() types.SSVSigner {
+func (c *Config) GetSigner() types.SSVShareSigner {
```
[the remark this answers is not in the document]
MatheusFranco99: Yes, better. Thanks

types/testingutils/operator_signer.go
```diff
@@ -16,10 +17,14 @@ type testingOperatorSigner struct {
 
 var (
 	testingOperatorSignerInstance *testingOperatorSigner
+	opMu                          sync.Mutex
```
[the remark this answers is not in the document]
MatheusFranco99: The lack of it randomly produces errors around tests

qbft/state.go
```diff
@@ -31,7 +31,7 @@ type IConfig interface {
 }
 
 type Config struct {
-	Signer         types.SSVShareSigner
+	SSVShareSigner types.SSVShareSigner
```
moshe-blox, line 34: this should probably be just `ShareSigner` like below, and the same for method names

tbh im not sure that we even need the word 'SSV' in the structs, but it's not too bad, lmk what you think @GalRogozinski @MatheusFranco99

moshe-blox (changes requested):

qbft/state.go
MatheusFranco99: I agree the word `SSV` here is redundant. I'll remove it, thx

qbft/round_change.go
```diff
@@ -271,6 +272,28 @@ func validRoundChangeForData(
 
 		return nil
 	}
+
+	return nil
+}
+
+func validRoundChangeForDataWithVerification(
```
y0sher, line 279: I don't want to nit pick on the naming, but `WithVerification` and `IgnoreSignature` don't sound related to each other when and fact the functionality should be identical with one difference. 
I would go with either "WithVerification" and "NoVerification"
or "IgnoreSignature" and "WithSignature"/"VerifySignature"
but leaving it like now is a bit confusing imo

ssv/aggregator.go
```diff
@@ -14,10 +14,11 @@ import (
 type AggregatorRunner struct {
 	BaseRunner *BaseRunner
 
-	beacon   BeaconNode
-	network  Network
-	signer   types.KeyManager
-	valCheck qbft.ProposedValueCheckF
+	beacon         BeaconNode
+	network        Network
+	signer         types.KeyManager
```
y0sher, line 19: if I understand correctly at some other places you call this `types.SSVShareSigner`, it's best to decide on similar name/type. unless you create local interfaces for each struct with only the subset of functions they need.

ssv/validator.go
```diff
@@ -10,26 +10,32 @@ import (
...
-	DutyRunners DutyRunners
-	Network     Network
-	Beacon      BeaconNode
-	Share       *types.Share
-	Signer      types.KeyManager
+	DutyRunners       DutyRunners
+	Network           Network
+	Beacon            BeaconNode
+	Share             *types.Share
+	Signer            types.KeyManager
```
y0sher, line 17: ditto about SSVShare signer and keymanager ( and many other places )

qbft/round_change.go
MatheusFranco99: Definitely. Forgot about that. Thanks!

ssv/aggregator.go
MatheusFranco99: The `KeyManager` includes the `SSVShareSigner`. Do you mean renaming the attribute `signer`?

types/signer.go
```diff
@@ -32,13 +34,22 @@ type BeaconSigner interface {
...
 
-// SSVSigner used for all SSV specific signing
-type SSVSigner interface {
+// SSVOperatorSigner used for to sign protocol messages
+type SSVOperatorSigner interface {
+	SignSSVMessage(data []byte, pk []byte) ([]byte, error)
+	// AddSSVOperatorKey saves an SSV operator key
+	AddSSVOperatorKey(sk *rsa.PrivateKey) error
+	// RemoveSSVOperatorKey removes an SSV operator key
+	RemoveSSVOperatorKey(pubKey string)
```
y0sher, line 43: IMO these are redundant, don't make us satisfy this interface (😆); There is always just a single operator key per instance, so there's never a need to add and remove them.

types/signer.go
```diff
@@ -48,7 +59,7 @@ type DKGSigner interface {
 // KeyManager is an interface responsible for all key manager functions
 type KeyManager interface {
 	BeaconSigner
-	SSVSigner
+	SSVShareSigner
 	// AddShare saves a share key
 	AddShare(shareKey *bls.SecretKey) error
 	// RemoveShare removes a share key
```
y0sher, line 65: mixing all this up is wrong if we're looking from SWE point.

y0sher (changes requested):
some small stuff.

types/signer.go
y0sher: oh I see now it does not include the operator signer. sorry.

ssv/aggregator.go
y0sher: Oh I just saw SSVShareSigner is part of keymanager. sorry. I thought you splitted them somehow. resolving.

ssv/validator.go
```diff
@@ -43,29 +49,63 @@ func (v *Validator) StartDuty(duty *types.Duty) error {
...
+	// Decode the nested SSVMessage
+	msg := &types.SSVMessage{}
+	if err := msg.Decode(signedSSVMessage.Data); err != nil {
+		return errors.Wrap(err, "could not decode data into an SSVMessage")
+	}
+
+	// Verify SignedSSVMessage's signature
+	if err := v.SignatureVerifier.Verify(signedSSVMessage, v.Share.Committee); err != nil {
+		return errors.Wrap(err, "SignedSSVMessage has an invalid signature")
+	}
```
moshe-blox, line 68: we should probably decode after verifying

types/messages.go
```diff
@@ -191,7 +191,7 @@ func (msg *SignedSSVMessage) Decode(data []byte) error {
 // - Data length should not be 0
 func (msg *SignedSSVMessage) Validate() error {
 	if msg.OperatorID == 0 {
-		return errors.New("OperatorID in SignedSSVMessage is 0")
+		return errors.New("signer ID 0 not allowed")
 	}
 	if len(msg.Signature) == 0 {
```
moshe-blox, line 196: is there a reason to not require the expected length of RSA signatures, or is that not the right place to do so, or not the right assumption for the spec to make?

MatheusFranco99: Actually, in the beginning, I wanted to let it open. But I've also added the custom decoding function to be used by implementation that sets it to 256 bytes
```go
signatureSize = 256
signature := encoded[signatureOffset : signatureOffset+signatureSize]
```
I have no opinion right now on whether we should check it here or leave the current check. 🤔

ssv/validator.go
MatheusFranco99: Agree. Thanks!

types/messages.go
moshe-blox: @GalRogozinski any reason why we shouldn't check that `len(sig) == signatureSize` in spec?

GalRogozinski: When you decode the data from pubsub you create a byte array of the expected length
maybe change `SignedMessage` to have `Signature [256]byte`

btw @MatheusFranco99 I remember we said the encoded length of the sig we use is actually greater correct?

MatheusFranco99: Ok, I'll change it to 256 bytes
This was about an encrypted key. For the signature, it's indeed 256

p2p/validation/msg_validation.go
```diff
@@ -14,28 +14,13 @@ import (
...
 
-// To-Do: better spec the intended Message Validation behavior. Issue #375: https://github.com/bloxapp/ssv-spec/issues/375
-type MessageValidator interface {
-	ValidateSignedSSVMessage(runner ssv.Runner, signedSSVMessage *types.SignedSSVMessage) error
-	// Verifies the message's RSA signature with PKCSv1.5 encoding. Ref: https://datatracker.ietf.org/doc/html/rfc8017#section-8.2.2
-	VerifySignature(runner ssv.Runner, signedSSVMessage *types.SignedSSVMessage) error
-}
-
-func MsgValidation(runner ssv.Runner, msgValidator MessageValidator) MsgValidatorFunc {
+func MsgValidation(runner ssv.Runner) MsgValidatorFunc {
```
GalRogozinski, line 17: do you see value to keep this file here for now?
if no, you may delete it

ssv/spectest/tests/runner/preconsensus/inconsistent_network_signer.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"github.com/attestantio/go-eth2-client/spec"
+
+	"github.com/bloxapp/ssv-spec/ssv/spectest/tests"
+	"github.com/bloxapp/ssv-spec/types"
+	"github.com/bloxapp/ssv-spec/types/testingutils"
+)
+
+// InconsistentNetworkSigner tests a SignedSSVMessage with inconsistent signer regarding its nested SignedPartialSignatureMessage
+func InconsistentNetworkSigner() tests.SpecTest {
```
GalRogozinski, line 12: change network to operator on all tests I think

GalRogozinski (commented):
just final nits and that's it

p2p/validation/msg_validation.go
GalRogozinski: doesn't have to be in this PR and not a blocker

moshe-blox (approved):

types/signer.go
```diff
@@ -48,7 +54,7 @@ type DKGSigner interface {
 // KeyManager is an interface responsible for all key manager functions
 type KeyManager interface {
 	BeaconSigner
-	SSVSigner
+	ShareSigner
 	// AddShare saves a share key
 	AddShare(shareKey *bls.SecretKey) error
 	// RemoveShare removes a share key
```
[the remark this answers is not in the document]
MatheusFranco99: Yes, other PR

p2p/validation/msg_validation.go
MatheusFranco99: Future PR

y0sher:
Q: are we signing rsa in consensus ? If we are verifying the we should probably also sign.

GalRogozinski:
> Q: are we signing rsa in consensus ? If we are verifying the we should probably also sign.
In spec we sign in validator.go

In impl from my POV you should sign where you pass all spec tests

GalRogozinski (approved):

PR #374, opened 2024-03-27
SignedSSVMessage

MatheusFranco99:
## Overview
Introduce `SignedSSVMessage`, which encapsulates `SSVMessage` along with a `signature` and a sender `identifier` fields.

### Changes
- New `SignedSSVMesage` structure in `types`
- Flat encoding for `SignedSSVMesage`
- `DecodePubsubMsg` now returns a `SignedSSVMesage`
- `MsgValidation` adapted to receive a `SignedSSVMesage` instead of a `SSVMessage`

Note: though `SSVMessage` is no longer the transmitted message in the network, the `Runner` still processes the type `SSVMessage` in the `ProcessMessage` function.

### Tests
- [x] Encoding
- [x] Valid SignedSSVMessage
- [x] Invalid SignedSSVMessage with no data
- [x] Invalid SignedSSVMessage with empty signature
- [x] Invalid SignedSSVMessage with zero signer
- [x] Invalid SignedSSVMessage with wrong data

Note: this is a non-fork version of #345


MatheusFranco99:
@y0sher I changed the encoding to match with the implementation one

p2p/validation/msg_validation.go
```diff
@@ -14,15 +14,23 @@ import (
 // MsgValidatorFunc represents a message validator
 type MsgValidatorFunc = func(ctx context.Context, p peer.ID, msg *pubsub.Message) pubsub.ValidationResult
 
-func MsgValidation(runner ssv.Runner) MsgValidatorFunc {
+type MessageValidator interface {
+	ValidateSignedSSVMessage(signedSSVMessage *types.SignedSSVMessage) error
+	ValidateSignature(signedSSVMessage *types.SignedSSVMessage) error
+}
```
GalRogozinski: Add comments:
1. Open a github issue that we need to add spec for MsgValidation and reference for `ValidateSignedSSVMessage`. Comment with the link to the issue.
2. For `ValidateSignature` comment that it is an RSA sig with PKCSv1.5 padding. You can link to the [RFC](https://datatracker.ietf.org/doc/html/rfc8017#section-8.2.2)... I think the link I gave is good, you can double-check 😛

types/messages.go
```diff
@@ -125,3 +135,90 @@ func (msg *SSVMessage) Encode() ([]byte, error) {
 func (msg *SSVMessage) Decode(data []byte) error {
 	return msg.UnmarshalSSZ(data)
 }
+
+// SSVMessage is the main message passed within the SSV network. It encapsulates the SSVMessage structure and a signature
+type SignedSSVMessage struct {
+	OperatorID OperatorID
+	Signature  []byte `ssz-max:"512"`     // Created by the operator's private key. Max size allow keys up to 512*8 = 4096 bits
+	Data       []byte `ssz-max:"6291893"` // Max size extracted from SSVMessage
+}
```
GalRogozinski, line 144: signature is the first field in impl
let;s do it exactly the same

types/messages_encoding.go
```diff
@@ -122,153 +122,3 @@ func (s *SSVMessage) HashTreeRootWith(hh ssz.HashWalker) (err error) {
...
-	}
-
-	hh.Merkleize(indx)
-	return
-}
-
-// GetTree ssz hashes the SignedSSVMessage object
-func (s *SignedSSVMessage) GetTree() (*ssz.Node, error) {
-	return ssz.ProofTree(s)
-}
```
GalRogozinski: won't this be regenrated automatically?
we either need to make an exception for the type or move it to a file that won't cause automatic generation

GalRogozinski (changes requested):

types/spectest/tests/signedssvmsg/valid.go
```diff
@@ -0,0 +1,32 @@
+package signedssvmsg
+
+import (
+	"github.com/bloxapp/ssv-spec/types"
+	"github.com/bloxapp/ssv-spec/types/testingutils"
+)
+
+// Valid tests a valid SignedSSVMessageTest
+func Valid() *SignedSSVMessageTest {
```
GalRogozinski, line 9: Now that you added `MessageValidator` mayb sig tests should be in `MessageValidator`...

The issue is since our `MessageValidator` isn't aligned with implementation not sure how this will work out for the SSV team.

wdyt @y0sher and @moshe-blox

types/messages_encoding.go
MatheusFranco99: It is not automatically generated because the encoding of the `messages.go` file is not included in the `generate` file. Adding a remark anyway.

types/spectest/tests/signedssvmsg/valid.go
MatheusFranco99: I agree a lot has to be done regarding Message Validation. Plus, tests for it and SignedSSVMessage are related. But I think we could do it in a later PR, since (if I'm not mistaken) it's not a top priority and we're overheaded with other tasks.

p2p/validation/msg_validation.go
```diff
@@ -13,13 +14,30 @@ import (
 // MsgValidatorFunc represents a message validator
 type MsgValidatorFunc = func(ctx context.Context, p peer.ID, msg *pubsub.Message) pubsub.ValidationResult
 
-func MsgValidation(runner ssv.Runner) MsgValidatorFunc {
+// To-Do: better spec the intended Message Validation behavior. Issue #375: https://github.com/bloxapp/ssv-spec/issues/375
+type MessageValidator interface {
+	ValidateSignedSSVMessage(runner ssv.Runner, signedSSVMessage *types.SignedSSVMessage) error
+	ValidateSignature(runner ssv.Runner, signedSSVMessage *types.SignedSSVMessage) error // Verifies the message's RSA signature with PKCSv1.5 encoding. Ref: https://datatracker.ietf.org/doc/html/rfc8017#section-8.2.2
```
moshe-blox, line 20:
```suggestion
	VerifySignature(runner ssv.Runner, signedSSVMessage *types.SignedSSVMessage) error // Verifies the message's RSA signature with PKCSv1.5 encoding. Ref: https://datatracker.ietf.org/doc/html/rfc8017#section-8.2.2
```

moshe-blox (approved):

p2p/validation/msg_validation.go
```diff
@@ -14,9 +14,10 @@ import (
 // MsgValidatorFunc represents a message validator
 type MsgValidatorFunc = func(ctx context.Context, p peer.ID, msg *pubsub.Message) pubsub.ValidationResult
 
+// To-Do: better spec the intended Message Validation behavior. Issue #375: https://github.com/bloxapp/ssv-spec/issues/375
 type MessageValidator interface {
 	ValidateSignedSSVMessage(runner ssv.Runner, signedSSVMessage *types.SignedSSVMessage) error
-	ValidateSignature(runner ssv.Runner, signedSSVMessage *types.SignedSSVMessage) error
+	ValidateSignature(runner ssv.Runner, signedSSVMessage *types.SignedSSVMessage) error // Verifies the message's RSA signature with PKCSv1.5 encoding. Ref: https://datatracker.ietf.org/doc/html/rfc8017#section-8.2.2
```
GalRogozinski, line 20: nit, can you put the comment above the function please?

p2p/validation/msg_validation.go
GalRogozinski: I don't object

I just want to admit that I use the words as synonyms and never understood the difference

types/spectest/tests/signedssvmsg/valid.go
y0sher: @GalRogozinski @MatheusFranco99 
I think this test should just use message validator as a blackbox, this means this test should check that all the behavior  we expect to happen or not happen based on the msg validator response should be tested under SignedSSVMsg tests, all the stuff regarding message validation (signatures and other validations) should happen on its own place.
For this test, what message validator did doesn't matter , it should just test what it expects to see if message validator returned valid/invalid. could use a mock imo.

y0sher (approved):

GalRogozinski (approved):

GalRogozinski (approved):

PR #379, opened 2024-04-08
committee consensus 

lior-blox:


ssv/spectest/tests/runner/consensus/past_msg.go
```diff
@@ -25,8 +26,15 @@ func PastMessage() tests.SpecTest {
...
-		fullData, _ := obj.Encode()
+	pastMsgF := func(cd *types.ConsensusData, beaconVote *types.BeaconVote, id []byte) *types.SignedSSVMessage {
+		var fullData []byte
+		if cd != nil {
+			fullData, _ = cd.Encode()
+		} else if beaconVote != nil {
+			fullData, _ = beaconVote.Encode()
+		} else {
+			panic("no consensus data or beacon vote")
+		}
```
GalRogozinski, line 37: simply use `Encoder` interface

types/messages.go
```diff
@@ -4,28 +4,25 @@ import (
...
-	"fmt"
 
+	"github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/pkg/errors"
 )
 
-// ValidatorPK is an eth2 validator public key
-type ValidatorPK []byte
+// ValidatorPK is an eth2 validator public key 48 bytes long
+// TODO: maybe we can use BLS pubkey simply?
```
lior-blox, line 13: lets remove this comment?

qbft/state.go
```diff
@@ -28,22 +24,22 @@ type IConfig interface {
 	GetNetwork() Network
 	// GetTimer returns round timer
 	GetTimer() Timer
+	// GetSignatureVerifier returns the signature verifier for operator signatures
+	GetSignatureVerifier() types.SignatureVerifier
+	// GetRoundCutOff returns the round cut off
```
nkryuchkov, line 29:
```suggestion
	// GetCutOffRound returns the round cut-off
```

qbft/commit.go
```diff
@@ -120,43 +122,52 @@ func baseCommitValidationIgnoreSignature(
...
 	}
 
-	if !bytes.Equal(proposedMsg.Message.Root[:], signedCommit.Message.Root[:]) {
+	proposedMsg, err := DecodeMessage(proposedSignedMsg.SSVMessage.Data)
+	if err != nil {
+		return err
+	}
+
+	if !bytes.Equal(proposedMsg.Root[:], msg.Root[:]) {
 		return errors.New("proposed data mistmatch")
```
nkryuchkov, line 171: Not part of this PR, but we might fix it now as we're touching this function. If you decide to fix it, please search for this word in other files too

nkryuchkov:
```suggestion
		return errors.New("proposed data mismatch")
```

qbft/round_change.go
```diff
@@ -11,11 +11,17 @@ import (
...
 ) error {
-	hasQuorumBefore := HasQuorum(i.State.Share, roundChangeMsgContainer.MessagesForRound(signedRoundChange.Message.
+
+	roundChangeMessage, err := DecodeMessage(signedRoundChange.SSVMessage.Data)
+	if err != nil {
+		return err
+	}
+
+	hasQuorumBefore := HasQuorum(i.State.Share, roundChangeMsgContainer.MessagesForRound(roundChangeMessage.
 		Round))
```
nkryuchkov, line 25: I think it doesn't have to be on the new line

nkryuchkov:
Should `ValidatorPK` be changed to `SenderID` in `p2p.Subscriber`?

```go
// Subscriber is used to abstract topic management
type Subscriber interface {
	Subscribe(vpk types.ValidatorPK) error
}
```

qbft/state.go
```diff
@@ -28,22 +24,22 @@ type IConfig interface {
...
+	GetCutOffRound() int
```
nkryuchkov, line 30: Why does it return `int` instead of `qbft.Round`?

qbft/state.go
```diff
@@ -10,12 +10,8 @@ import (
 )
 
 type signing interface {
-	// GetShareSigner returns an share signer instance
-	GetShareSigner() types.ShareSigner
 	// GetSigner returns an operator signer instance
```
nkryuchkov, line 13:
```suggestion
	// GetOperatorSigner returns an operator signer instance
```

qbft/state.go
```diff
@@ -81,14 +77,23 @@ func (c *Config) GetTimer() Timer {
...
+}
+
+// GetSignatureVerifier returns the verifier for operator's signatures
+func (c *Config) GetSignatureVerifier() types.SignatureVerifier {
+	return c.SignatureVerifier
+}
+
 type State struct {
-	Share                           *types.Share
+	Share                           *types.Operator
```
nkryuchkov, line 90: As I understand, `Share` represents  what we used to call `Share` for non-committee roles, and `Operator` represents it for committee roles. If it's correct, then why does `State` support `Share`? If not correct, what's the difference between `Share` and `Operator` and why do we need both?

ssv/committee_runner.go
```diff
@@ -0,0 +1,387 @@
...
+
+// unneeded
+func (cr CommitteeRunner) expectedPreConsensusRootsAndDomain() ([]ssz.HashRoot, phase0.DomainType, error) {
+	return nil, types.DomainError, errors.New("no pre consensus root for committee runner")
+}
+
+// This function signature returns only one domain type
+// instead we rely on expectedPostConsensusRootsAndBeaconObjects that is called later
+func (cr CommitteeRunner) expectedPostConsensusRootsAndDomain() ([]ssz.HashRoot, phase0.DomainType, error) {
+	return []ssz.HashRoot{}, types.DomainAttester, nil
```
nkryuchkov, line 263: Why does it always return `DomainAttester`? What about `DomainSyncCommittee`?

moshe-blox:
@GalRogozinski ClusterID should be rename to CommitteeID

qbft/state.go
GalRogozinski: Pretty much the former Share type held both operator and validator share metadata.
`Operator` is a pure SSV entity and `Validator` is an eth entity.
We now decoupled those 2 domains to `Operator` and `Share`.
I did add comments on these types. Maybe they were not clear enough?

The word `Share` implies the entity is part of a committee, so both `Operator` and `Share` types are called `Share` as variables. I admit that part of the reason the `Share` variable didn't change the name is to have less conflicts and changes for this huge PR. Maybe it should be renamed `OperatorShare` or some other name? Or maybe `share` is a clear short name? If you have some thoughts for better naming do share?

nkryuchkov: @GalRogozinski thanks for the explanation. I don't have any better ideas right now, I'll leave a comment if something comes to my mind

types/messages.go
GalRogozinski: done

types/messages.go
```diff
@@ -4,28 +4,25 @@ import (
...
-	roleTypeStartPos = pubKeyStartPos + pubKeySize
-
-	// SignedSSVMessage offsets
-	signatureSize    = 256
-	signatureOffset  = 0
-	operatorIDSize   = 8
-	operatorIDOffset = signatureOffset + signatureSize
-	messageOffset    = operatorIDOffset + operatorIDSize
+	roleTypeStartPos = domainStartPos + domainSize
+	senderIDSize     = 48
```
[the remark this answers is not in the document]
GalRogozinski: done

qbft/state.go
GalRogozinski: According Webster, cutoff should be one word when used as noun or adjective

qbft/state.go
GalRogozinski: changed

qbft/controller.go
```diff
@@ -17,15 +17,11 @@ type Controller struct {
...
+	Share           *types.Operator
 	config          IConfig
 }
 
-func NewController(
-	identifier []byte,
-	share *types.Share,
-	config IConfig,
-) *Controller {
+func NewController(identifier []byte, share *types.Operator, config IConfig) *Controller {
```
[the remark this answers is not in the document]
GalRogozinski: I think changing the name in Controller will ad more differences to the jsons that will make the review harder...
So we will do it in the end

qbft/state.go
nkryuchkov: @GalRogozinski Sorry, I trusted my IDE. Just aligning the comment to the function name is fine

qbft/controller.go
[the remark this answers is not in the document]
nkryuchkov: ok, I don't mind

PR #382, opened 2024-04-12
New message structures with RSA signature

MatheusFranco99:
## Overview

This PR implements https://github.com/bloxapp/SIPs/pull/45. It deprecates `qbft.SignesMessage` and `types.SignedPartialSignatureMessage`.

## Changes
- Extend `types.SignedSSVMessage` to include a buffer of signatures to encompass decided messages.
- Include `FullData` in `types.SignedSSVMessage`.
- Deprecate `qbft.SignesMessage` and `types.SignedPartialSignatureMessage`.
- Change QBFT to process `types.SignedSSVMessage`.
- Now, for PreConsensus and PostConsensus, the Runner accepts `types.PartialSignatureMessages`.
- Change PreConsensusJustification to hold `types.PartialSignatureMessages`.


qbft/messages.go
```diff
@@ -108,6 +75,12 @@ func (msg *Message) Decode(data []byte) error {
 	return msg.UnmarshalSSZ(data)
 }
 
+func GetMessageFromBytes(data []byte) (*Message, error) {
```
moshe-blox, line 78: suggest to rename to `DecodeMessage` move right below the struct definition (just to conform to Go standards)

types/messages.go
```diff
@@ -138,103 +130,206 @@ func (msg *SSVMessage) Decode(data []byte) error {
 
 // SSVMessage is the main message passed within the SSV network. It encapsulates the SSVMessage structure and a signature
 type SignedSSVMessage struct {
-	Signature  [256]byte // Created by the operator's network key
-	OperatorID OperatorID
-	Data       []byte
+	Signature  [][]byte     `ssz-max:"13,256"` // Created by the operators' key
+	OperatorID []OperatorID `ssz-max:"13"`
```
moshe-blox, line 134: since it's plural, i'd rename to Signatures and Signers or Operators

GalRogozinski: I would rename now to make this PR makes sense
when we have a PR to drop decided we can rename again...

Unless we are sure @moshe-blox that we want to drop decided messages in this fork

GalRogozinski (approved):

types/messages.go
GalRogozinski: Just to remind everyone. In the last discussion the options were: Decided messages should either become post-consensus justifications,  or be emitted upon an RC messages, or be dropped entirely.

But this has nothing to do with this PR. We can rename in the meanwhile...

qbft/message_container.go
```diff
@@ -3,82 +3,98 @@ package qbft
...
 	return &MsgContainer{
-		Msgs: map[Round][]*SignedMessage{},
+		Msgs: map[Round][]*types.SignedSSVMessage{},
 	}
 }
 
 // AllMessaged returns all messages
-func (c *MsgContainer) AllMessaged() []*SignedMessage {
-	ret := make([]*SignedMessage, 0)
+func (c *MsgContainer) AllMessaged() []*types.SignedSSVMessage {
```
moshe-blox, line 21: unrelated to this PR, but if we already changed this line we can finally fix this typo 😅 (`AllMessages`)

types/messages.go
moshe-blox: > Unless we are sure @moshe-blox that we want to drop decided messages in this fork

it's yet to be decided (🤦‍♂️)

it's your call, but i think it's a quick rename and can be quickly undone if needed

qbft/generate.go
```diff
@@ -1,4 +1,4 @@
 package qbft
 
-// rm -f ./messages_encoding.go
-// go run github.com/ferranbt/fastssz/sszgen --path messages.go --include ./types.go,../types/signer.go,../types/operator.go --exclude-objs OperatorID
+//go:generate rm -f ./messages_encoding.go
+//go:generate go run github.com/ferranbt/fastssz/sszgen --path messages.go --include ./types.go,../types/signer.go,../types/operator.go --exclude-objs OperatorID
```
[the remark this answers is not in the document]
moshe-blox: @GalRogozinski thanks for noting

@y0sher @lior-blox we need to consider how this impacts fork support

GalRogozinski (changes requested):
In signer.go can ShareSigner and then in turn KeyManager be deleted so only BeaconSigner is left?

It is fine to delete DKSSigner in the process...

types/messages.go
MatheusFranco99: I agree with the renaming and it's easy to do. Updating it

qbft/message_container.go
MatheusFranco99: Oh man, definitely

qbft/messages.go
MatheusFranco99: Yes, better. Thanks!

GalRogozinski (approved):

GalRogozinski:
@MatheusFranco99 conflicts occurred due to #381 
I am merging changes from `main` to `alan`

qbft/generate.go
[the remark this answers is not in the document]
y0sher: This is going to break all our message encoding, we'll probably need to use both versions for forking :/

qbft/prepare.go
```diff
@@ -28,7 +28,12 @@ func (i *Instance) uponPrepare(signedPrepare *SignedMessage, prepareMsgContainer
 		return nil // no quorum yet
 	}
 
-	proposedRoot := i.State.ProposalAcceptedForCurrentRound.Message.Root
+	proposalMsgAccepted, err := DecodeMessage(i.State.ProposalAcceptedForCurrentRound.SSVMessage.Data)
```
y0sher, line 31: I've noticied that we are decoding the message multiple times around the code, maybe its worth to consider some approach where we decode once and pass the decoded message as well to functions to reduce the amount of calls to decode..

types/messages.go
```diff
@@ -138,103 +130,206 @@ func (msg *SSVMessage) Decode(data []byte) error {
...
-	return msg.OperatorID
+// GetOperatorIDs returns the sender operator ID
+func (msg *SignedSSVMessage) GetOperatorIDs() []OperatorID {
+	return msg.OperatorIDs
 }
 
 // GetSignature returns the signature of the OperatorID over Data
-func (msg *SignedSSVMessage) GetSignature() [256]byte {
-	return msg.Signature
+func (msg *SignedSSVMessage) GetSignature() [][]byte {
```
y0sher, line 146: the name is now tricky imo should be called GetSignatures

types/messages.go
```diff
@@ -138,103 +130,206 @@ func (msg *SSVMessage) Decode(data []byte) error {
...
+	msg.Signatures = append(msg.Signatures, msgToAggregate.Signatures...)
+	msg.OperatorIDs = append(msg.OperatorIDs, msgToAggregate.GetOperatorIDs()...)
+
+	return nil
+}
+
+// Check if all signedMsg's signers belong to the given committee in O(n+m)
+func (msg *SignedSSVMessage) CheckSignersInCommittee(committee []*Operator) bool {
+	// Committee's operators map
+	committeeMap := make(map[uint64]struct{})
```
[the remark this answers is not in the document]
y0sher: +1

types/messages.go
```diff
@@ -138,103 +130,206 @@ func (msg *SSVMessage) Decode(data []byte) error {
...
+	}
+
+	for _, id := range msg.GetOperatorIDs() {
+		found := false
+		for _, id2 := range ids {
+			if id == id2 {
+				found = true
+			}
+		}
+
```
y0sher, line 267: could probably used slices.Contains

types/messages.go
```diff
@@ -138,103 +130,206 @@ func (msg *SSVMessage) Decode(data []byte) error {
...
+// - SSVMessage must not be nil
 func (msg *SignedSSVMessage) Validate() error {
-	if msg.OperatorID == 0 {
-		return errors.New("signer ID 0 not allowed")
+	// Validate OperatorID field
+	if len(msg.OperatorIDs) == 0 {
+		return errors.New("no signers")
+	}
+	// Check unique signers
+	signed := make(map[OperatorID]bool)
```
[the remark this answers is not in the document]
GalRogozinski: just noting that from spec perspective we just need to make it readable and we less care about optimized memory allocation.

I guess your way is also readable and since you guys tend to copy-paste from spec I changed.

qbft/spectest/tests/proposal/different_justifications.go
```diff
@@ -49,10 +48,10 @@ func DifferentJustifications() tests.SpecTest {
 	return &tests.MsgProcessingSpecTest{
 		Name:          "different proposal round change justification",
 		Pre:           pre,
-		PostRoot:      "eeee6a5783cec81c67ffe6a185f9afb4c775b80a815f22dbb32489f8ffa6adce",
+		PostRoot:      "020830fcec4b0feffbcbf86465926fe80657ce62d1dc637cc141856e4013d6ef",
 		InputMessages: msgs,
-		OutputMessages: []*qbft.SignedMessage{
-			testingutils.TestingPrepareMessageWithRound(ks10.Shares[1], types.OperatorID(1), 3),
+		OutputMessages: []*types.SignedSSVMessage{
+			testingutils.TestingPrepareMessageWithRound(ks.OperatorKeys[1], types.OperatorID(1), 3),
```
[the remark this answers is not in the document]
GalRogozinski: True!
But reading through the test it seems @MatheusFranco99 was correct with this change...
The reason it passed before is because the tests don't verify sigs, they only compare signing roots

types/messages.go
GalRogozinski: I notice we don't even call this method along another one so I am simply deleting

types/messages.go
GalRogozinski: ok @MatheusFranco99 
I think I found some bugs here:
1. Your comment talks about equality: `// MatchedSigners returns true if the provided signer ids are equal to GetOperatorIDs() without order significance`. To check equality you need to do two way containment test, not one (maybe there is a util function). But I think the bug is in the comment and not in the logic.
2. I think you wanted to do a one way containment test, but I think it should be the other way? That is because from `func (c *MsgContainer) AddFirstMsgForSignerAndRound` I understand we need to check whether all `ids` are included in `msg`?

types/messages.go
GalRogozinski: yeah... not sure if we do it now

types/messages.go
GalRogozinski: @MatheusFranco99 pointed out that this is the same logic as before

So opened #387 
Later we inspect and fix in a separate PR

PR #397, opened 2024-05-13
Fix role mapping in MapDutyToRunnerRole

nkryuchkov:
The PR fixes the case when `dutyRole` is `BNRoleAttester`. 

Currently, the result in this case is `RoleUnknown` because `BNRoleAttester | BNRoleSyncCommittee == 0 | 3 == 3`, so `RoleCommittee` is returned only for `BNRoleSyncCommittee`. This would've worked if roles were powers of two, but they are just `iota`. 

`TestMapDutyToRunnerRole` shows the change, it passes with the change in `MapDutyToRunnerRole` and fails without it.


moshe-blox (approved):

GalRogozinski (changes requested):
Good!

The spectest should test the type, not the function.
Note that (BeaconDuty.RunnerRole()) calls this function.

Do the following:
1. `types/spectest/tests` create a `duty` folder.
2. transform your test to a test on the duty type like other tests.
3. While you are touching this code it would be a good idea to change line 45 of `pre_consensus_justifications.go` to 	`if b.RunnerRoleType != data.Duty.RunnerRole() {`

if you have questions don't be shy

nkryuchkov:
@GalRogozinski Done, please review again

GalRogozinski (approved):

PR #406, opened 2024-05-20
Update go

GalRogozinski:
Update go to 1.20 
also updated github actions

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -16,7 +16,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.19
+          go-version: 1.20
 
       - name: Get Dependencies
         run: go get -v -t -d ./...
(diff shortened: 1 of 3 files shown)


.github/workflows/lint.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.19
+          go-version: 1.20
```
moshe-blox, line 19: try this instead:
```yaml
      - name: Setup Go
        uses: actions/setup-go@v4
        with:
          go-version: "1.20.x"
```

moshe-blox (changes requested):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/ssvlabs/ssv-spec
 
-go 1.19
+go 1.20
```
nkryuchkov, line 3: Why 1.20 and not 1.22?

nkryuchkov: Well, I think I can answer myself: because we are still using 1.20 in the implementation

MatheusFranco99: Yes

MatheusFranco99 (approved):

moshe-blox (approved):

PR #408, opened 2024-05-22
up go-eth2-client

GalRogozinski:
Up go-eth2-version


GalRogozinski:
@MatheusFranco99 and @AKorpusenko 
can you give me a quick explanation why we generating ssz for `consensusdata` and other types that need go-eth-2 now need ssz?

MatheusFranco99:
@GalRogozinski 
Both BeaconVote and ConsensusData are encoded to be sent as []byte to the qbft module, and we use ssz for that.

PR #409, opened 2024-05-23
Remove test cases with 500 validators

MatheusFranco99:
This PR removes test cases with 500 validators to reduce the size of the JSON tests file.


GalRogozinski:
I see in the Jsons that you also removed 100? that is necessary to get the file size down?

MatheusFranco99:
@GalRogozinski Yup

GalRogozinski (approved):

PR #410, opened 2024-05-23
Validate empty duties

GalRogozinski:
Throw an error in case Committee duty is empty.
Technically a test should be added for a filtered duty, but since we will soon remove this I decided against adding it


MatheusFranco99 (changes requested):
Good job!
- Can you add a test that starts a duty for slot X and then starts another duty for slot X-1 (with the same validators), to see that the error is also triggered that way?
- Also, can you fix the comment on `start_no_duties.go` where it has
```go
// Expects: error?
```
It can either name the expected error or even let the comment be removed since it can be seen in the test spec.

GalRogozinski:
> Can you add a test that starts a duty for slot X and then starts another duty for slot X-1 (with the same validators), to see that the error is also triggered that way?

I think you meant: Test the current `FilterCommitteeDuty` logic?
However in the SIP meeting @alonmuroch said we can remove the logic in spec but maybe keep it in impl (so I will remove soon). This means we shouldn't test this... because then we won't allow the impl to make this decision. I also wrote in the description that's why I am omitting this test

MatheusFranco99:
@GalRogozinski 
I agree. Thanks for clarifying

MatheusFranco99 (approved):

PR #412, opened 2024-05-26
Allow committee duties to run in parallel

GalRogozinski:
# Description

For committee duties we previously didn't allow validators to complete a duty if a new one came in for the same validator.
Now we no longer stop duty execution. In order to not risk slashing we now add an additional slashing check before signing in the post-consensus phase.


GalRogozinski:
@MatheusFranco99 commit is good

MatheusFranco99 (approved):

PR #413, opened 2024-05-27
Fix runner state in MessageProcessingSpecTest

MatheusFranco99:
This PR fixes the runner state for the CommitteeRunner test cases in the MessageProcessingSpecTest.


GalRogozinski:
> Before, I have also tried using the test.Runner when creating a runner for a duty. But some tests had troubles

I thought it would be simple. If I am wrong and you don't manage to fix this way simply write an explanation of why 🙏

MatheusFranco99:
@GalRogozinski 
Fixed.
The tests that had errors were due to using the same Runner for more than one duty. But these tests are now of type CommitteeSpecTest, so all solved :)

ssv/spectest/generate/state_comparison/tests_MultiMsgProcessingSpecTest/consensus_future_decided/sync committee.json
```diff
@@ -110,8 +110,7 @@
 														0,
 														1,
 														2
-												],
-												"IsStopped": false
```
GalRogozinski: what's with this change 🤔

GalRogozinski: It seems that jsons were not generated correctly in #412

PR #414, opened 2024-05-28
Fix DomainType in CommitteeRunner

MatheusFranco99:
This PR fixes the DomainType to be used when creating the MsgID in the CommitteeRunner.

Closes https://github.com/ssvlabs/ssv-spec/issues/411.


GalRogozinski (changes requested):
1.Delete the TODO in SHARE
2. Add AlanDomain that has mainnet network and a new fork. Epoch should still have a TODO
3. Make sure in other places we call the domain appropriately

MatheusFranco99:
@GalRogozinski 
1 is done.
2 was already done.
3 also sone.

types/testingutils/verifier.go
```diff
@@ -22,14 +22,11 @@ var (
...
 	if testingVerifierInstance == nil {
-		testingVerifierInstanceLock.Lock()
-		defer testingVerifierInstanceLock.Unlock()
-		if testingVerifierInstance == nil {
-			testingVerifierInstance = &testingVerifier{
-				signaturesCache: make(map[types.OperatorID]map[[32]byte][]byte),
-			}
-			return testingVerifierInstance
+		testingVerifierInstance = &testingVerifier{
+			signaturesCache: make(map[types.OperatorID]map[[32]byte][]byte),
```
GalRogozinski, line 29: Can you use `go.Once()` as recommended by wikipedia?
https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Go

if you scroll down to the java solution you see it can become tricky under certain memory models...
Before the race condition didn't always appear, I think now you solved it but let's do it the proper performant go way

GalRogozinski (changes requested):

types/testingutils/verifier.go
MatheusFranco99: Notice that the problem is not with the singleton implementation but with the race conditions between the access in the constructor and other methods.
`go.Once` won't not solve it alone. I still need to leave the lock there.
Changing.

MatheusFranco99: Changing.

types/testingutils/verifier.go
```diff
@@ -17,21 +17,19 @@ type testingVerifier struct {
...
-	if testingVerifierInstance == nil {
-		testingVerifierInstanceLock.Lock()
-		defer testingVerifierInstanceLock.Unlock()
-		if testingVerifierInstance == nil {
-			testingVerifierInstance = &testingVerifier{
-				signaturesCache: make(map[types.OperatorID]map[[32]byte][]byte),
-			}
-			return testingVerifierInstance
+	testingVerifierInstanceLock.Lock()
+	defer testingVerifierInstanceLock.Unlock()
```
GalRogozinski, line 27: no need for this lock I think now :-)

GalRogozinski: that's the whole point of the `sync.Once` primitive in go. To solve this synchronization problem correctly so you don't add locks yourself.
There are issues in the memory model in some programming languages to do efficient double locking (including go), what you initially tried to do and then fixed by making less efficient. 

`sync.Once` solves this efficiently and elegantly

MatheusFranco99: Okk, great!!

PR #415, opened 2024-05-28
Validate signing roots for CommitteeRunner

MatheusFranco99:
This PR adds a validation on the number of signing roots (and their content) for the CommitteeRunner and aligns the existing tests with the new error.

### DIFF
--- a/ssv/runner_validations.go
+++ b/ssv/runner_validations.go
@@ -19,12 +19,12 @@ func (b *BaseRunner) ValidatePreConsensusMsg(runner Runner, psigMsgs *types.Part
 		return err
 	}
 
-	roots, domain, err := runner.expectedPreConsensusRootsAndDomain()
+	roots, domains, err := runner.expectedPreConsensusRootsAndDomain()
 	if err != nil {
 		return err
 	}
 
-	return b.verifyExpectedRoot(runner, psigMsgs, roots, domain)
+	return b.verifyExpectedRoot(runner, psigMsgs, roots, domains)
 }
 
 // Verify each signature in container removing the invalid ones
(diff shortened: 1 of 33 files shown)


ssv/runner_validations.go
```diff
@@ -97,21 +81,25 @@ func (b *BaseRunner) validateDecidedConsensusData(runner Runner, val types.Encod
...
-		if err != nil {
-			return nil, errors.Wrap(err, "could not get pre consensus root domain")
-		}
 
 		ret := make([][32]byte, 0)
-		for _, rootI := range expectedRootObjs {
+		for i, rootI := range expectedRootObjs {
+			d, err := runner.GetBeaconNode().DomainData(epoch, domains[i])
+			if err != nil {
+				return nil, errors.Wrap(err, "could not get pre consensus root domain")
```
GalRogozinski, line 100: pre consensus?

ssv/spectest/tests/runner/postconsensus/invalid_then_quorum.go
```diff
@@ -26,7 +26,7 @@ func InvalidThenQuorum() tests.SpecTest {
 				),
 				Duty: testingutils.TestingAttesterDuty,
 				Messages: []*types.SignedSSVMessage{
-					testingutils.SignPartialSigSSVMessage(ks, testingutils.SSVMsgCommittee(ks, nil, testingutils.PostConsensusWrongAttestationMsg(ks.Shares[1], 1, testingutils.TestingDutySlot))),
+					testingutils.SignPartialSigSSVMessage(ks, testingutils.SSVMsgCommittee(ks, nil, testingutils.PostConsensusWrongSigAttestationMsg(ks.Shares[1], 1, testingutils.TestingDutySlot))),
```
GalRogozinski, line 29: good catch

GalRogozinski (commented):
Good!

The only thing we have to think about before merging this PR is that for committee runner the code was initially designed to work if a post-consensus was missing a root (in case there was an issue with a certain validator).

Not sure if we need to really care for this case though...
Let's talk about it tomorrow

ssv/runner_validations.go
MatheusFranco99: Well seen. Btw, we always had this error string. But I think we can't reach this part in a test.

MatheusFranco99: Fixed.

moshe-blox:
> Good!
> 
> The only thing we have to think about before merging this PR is that for committee runner the code was initially designed to work if a post-consensus was missing a root (in case there was an issue with a certain validator).
> 
> Not sure if we need to really care for this case though... Let's talk about it tomorrow

Yes, either missing or has extra roots, because currently SSV doesn't guarantee equal validator sets between operators.

If a validator registered with 1 corrupt BLS share, only that specific operator would be able to detect this, so that operator would not run this validator and there'd be 1 missing signature for the other operators, and the opposite for that operator which would see 1 extra signature.

The other reason for mismatching validator sets is contract event syncing disparity.

GalRogozinski:
So @MatheusFranco99 
It seems that we should allow subset of roots

GalRogozinski:
>> Yes, and also a superset so that you can ignore signatures for validator indices you dont recognize.

So @MatheusFranco99 maybe we need to rethink this PR...

MatheusFranco99:
@GalRogozinski Let's discuss what should be done here

GalRogozinski:
@MatheusFranco99 
Add @moshe-blox comments to the SIP, and make sure that the current code meets the criteria

MatheusFranco99:
@GalRogozinski 
Added it in https://github.com/ssvlabs/SIPs/pull/42/commits/8df6b0d118792bba3886d60d85317b5314b24da1

PR #428, opened 2024-06-24
BeaconNode: Submit multiple attestation duties

MatheusFranco99:
This PR:
- changes the BeaconNode interface to allow the submission of multiple attestation objects at once.
- aligns the CommitteeRunner to submit all attestation objects in a unique call.


ssv/committee_runner.go
```diff
@@ -249,29 +250,44 @@ func (cr CommitteeRunner) ProcessPostConsensus(signedMsg *types.PartialSignature
...
-			if role == types.BNRoleAttester {
-				att := sszObject.(*phase0.Attestation)
-				// Insert signature
-				att.Signature = specSig
-
-				if err := cr.beacon.SubmitAttestation(att); err != nil {
-					return errors.Wrap(err, "could not submit to Beacon chain reconstructed attestation")
-				}
-			} else if role == types.BNRoleSyncCommittee {
+			if role == types.BNRoleSyncCommittee {
```
GalRogozinski, line 253: why did you switch order?

MatheusFranco99: I just thought that it would be better, for the reading flow, to see the duties that are promptly submitted and after it the duties that are stored for later submission.

types/testingutils/beacon_node.go
```diff
@@ -681,9 +681,11 @@ func (bn *TestingBeaconNode) GetAttestationData(slot *phase0.Slot, committeeInde
 }
 
 // SubmitAttestation submit the attestation to the node
-func (bn *TestingBeaconNode) SubmitAttestation(attestation *phase0.Attestation) error {
-	r, _ := attestation.HashTreeRoot()
-	bn.BroadcastedRoots = append(bn.BroadcastedRoots, r)
+func (bn *TestingBeaconNode) SubmitAttestation(attestations []*phase0.Attestation) error {
```
y0sher, line 684: SubmitAttestations

ssv/types.go
```diff
@@ -32,7 +32,7 @@ type AttesterCalls interface {
 	GetAttestationData(slot *phase0.Slot, committeeIndex *phase0.CommitteeIndex) (*phase0.AttestationData,
 		spec.DataVersion, error)
 	// SubmitAttestation submit the attestation to the node
-	SubmitAttestation(attestation *phase0.Attestation) error
+	SubmitAttestation(attestations []*phase0.Attestation) error
```
y0sher, line 35: `SubmitAttestations`

ssv/types.go
```diff
@@ -32,7 +32,7 @@ type AttesterCalls interface {
 	GetAttestationData(slot *phase0.Slot, committeeIndex *phase0.CommitteeIndex) (*phase0.AttestationData,
```
y0sher, line 32: not part of this PR but since you are editing this already, I don't know why you added pointers to slot and committeeIndex, it should be like this -
`slot phase0.Slot, committeeIndex phase0.CommitteeIndex` as these are int values

ssv/types.go
y0sher, line 1: also the correct API and signature for sync committee message sending is - 
`SubmitSyncMessages(msg []*altair.SyncCommitteeMessage) error` maybe lets take care of it here as well

ssv/committee_runner.go
y0sher, line 1: we could probably send sync committee as bulk as well, beacon node enables this
`SubmitSyncMessages(msg []*altair.SyncCommitteeMessage) error`

ssv/types.go
y0sher: you are wrong, they are required 
image

please revert back to using values instead of pointers.

GalRogozinski: hmm I was looking in the `Beacon` api and not `Validator` like you did.
But I think you are correct, the `Validator` one should be used

But what happens when a committee runner asks validators to create attestations for the wrong committee index?
Are we sure that it works on all clients?
If it does, I am not sure that we have a promise that later on this won't break by some validation?

y0sher (approved):

ssv/types.go
y0sher: we are using the validator api in all clients already.
not sure about your first question though.

PR #430, opened 2024-06-27
Use PEM exclusively

GalRogozinski:
Generate tests with RSA keys in PEM format


GalRogozinski:
@MatheusFranco99 I wonder if the ssz size in the tag is correct?
Anyhow imo on a seperate task it can be removed since we only need ssz for objects transferred on wire

MatheusFranco99:
@GalRogozinski 
You are referring to the 294 size as in the below structures, right?
```
// CommitteeMember represents an SSV operator node that is part of a committee
type CommitteeMember struct {
	OperatorID        OperatorID
	CommitteeID       CommitteeID `ssz-size:"32"`
	SSVOperatorPubKey []byte      `ssz-size:"294"`
	// FaultyNodes is the number of nodes that are considered faulty or malicious in the operator's committee
	FaultyNodes uint64
	// All the members of the committee
	Committee []*Operator `ssz-max:"13"`
}

// Operator represents a node in the network that holds an ID and a public key
type Operator struct {
	OperatorID        OperatorID
	SSVOperatorPubKey []byte `ssz-size:"294"`
}
```
I believe that if you use PEM you'll get way bigger sizes. Probably, we didn't get errors from the ssz tag because, as you said, we are never really encoding it.

MatheusFranco99:
Namely, for our case:
- Size of PubKey from MarshalPKIX: 294
- Size of PubKey from MarshalPKIX then PEM size: 459

MatheusFranco99:
@GalRogozinski 
Since sizes are now different, should we update it in this PR or perhaps already remove it?

GalRogozinski:
remove it in different PR

MatheusFranco99 (approved):

y0sher (approved):

MatheusFranco99 (approved):

PR #433, opened 2024-06-27
Calculate CommitteeID when creating message IDs

GalRogozinski:
1. Add `DomainType` to `CommitteeMember`
2. Use `DomainType` and `CommitteeID` to create MessageID in CommitteeRunner


GalRogozinski:
@lior-blox 
Note that we have `Domain` in `Config` struct (old thing Alon did).
So we can use this instead

MatheusFranco99 (approved):

y0sher (changes requested):
Just to summarize, I think `DomainType` fits better in config than in committeeMember ( or share for that matter) since its used for everything under the same network and we're not going to have different domainTypes for different shares/members ever unless its a bug.

GalRogozinski:
@y0sher 
I agree with you. In my last discussion with @lior-blox he didn't agree to do the change the way you described.
Maybe we try to speak with him again?

PR #436, opened 2024-06-30
Delete preconsensus justifications

GalRogozinski:
Deleting all preconsenus justification code


GalRogozinski:
should we delete preconsensus justifications from consensus data @alonmuroch?
we can keep it there so we don't have to fork to add this

MatheusFranco99 (approved):

PR #440, opened 2024-07-01
Add ProcessingMessage structure to QBFT

MatheusFranco99:
## Overview

This PR introduces the `ProcessingMessage` structure to be used in the `qbft` module.

This structure encompasses a `SignedSSVMessage` and its associated decoded `qbft.Message`.

The `SignedSSVMessage` is needed in the `qbft` module due to QBFT justifications, while the cached `qbft.Message` in the structure allows us to remove the decoding duplication behaviour.

### Usage

We avoid using `SignedSSVMessage` in `qbft` as much as possible. The two exceptions are:
- the `qbft.Controller.ProcessMsg` function which is considered the message's entry point to the `qbft` module. This function transforms the `SignedSSVMessage` to a `ProcessingMessage` right away, which is, then, used in the rest of `qbft`.
- the `Instance.Broadcast` function since we broadcast a `SignedSSVMessage` (and not a message to be processed).

Thus, `ProcessingMessage` is used everywhere else, as in `MsgContainer`, `State.ProposalAcceptedForCurrentRound`, and in `Instance` methods.


qbft/generate.go
```diff
@@ -1,4 +1,4 @@
 package qbft
 
 //go:generate rm -f ./messages_encoding.go
-//go:generate go run github.com/ferranbt/fastssz/sszgen --path messages.go --include ./types.go,../types/signer.go,../types/operator.go --exclude-objs OperatorID
+//go:generate go run github.com/ferranbt/fastssz/sszgen --path messages.go --include ./types.go,../types/signer.go,../types/operator.go --exclude-objs OperatorID,ProcessingMessage
```
GalRogozinski, line 4: we don't pass processing message on wire..
why do we need ssz encoding for it?

qbft/messages.go
```diff
@@ -166,3 +166,56 @@ func Sign(msg *Message, operatorID types.OperatorID, operatorSigner types.Operat
...
+func (msg *ProcessingMessage) Validate() error {
+
+	// Validate
+	if err := msg.SignedMessage.Validate(); err != nil {
+		return errors.Wrap(err, "invalid SignedSSVMessage")
+	}
+
+	// Check if the decoded message is correct
+	decodedMsg := &Message{}
+	err := decodedMsg.Decode(msg.SignedMessage.SSVMessage.Data)
```
GalRogozinski, line 200: validation purpose is to validate message that comes from unknown actor on the network (we probably have duplication with p2p message validation)...

I don't think we need to decode all the time in spec

qbft/state.go
```diff
@@ -88,7 +88,7 @@ type State struct {
 	Height                          Height
 	LastPreparedRound               Round
 	LastPreparedValue               []byte
-	ProposalAcceptedForCurrentRound *types.SignedSSVMessage
+	ProposalAcceptedForCurrentRound *ProcessingMessage
```
GalRogozinski, line 91: must we have a state change? This just changes all tests...
If we must it is fine

qbft/generate.go
MatheusFranco99: I've added ProcessingMessage to the excluded objects

qbft/state.go
MatheusFranco99: I can revert to `types.SignedSSVMessage` but this will add again the decoding (only for this variable). What do you think? I have no opinion here

GalRogozinski: where do you do the decoding?
I can't find it

MatheusFranco99: I mean that if we store `ProposalAcceptedForCurrentRound` as a `types.SignedSSVMessage` then once we use it (e.g. for UponPrepare and UponCommit) we will need to decode its content again

GalRogozinski: Here is the old `UponCommit`:
```go
// UponCommit returns true if a quorum of commit messages was received.
// Assumes commit message is valid!
func (i *Instance) UponCommit(signedCommit *types.SignedSSVMessage, commitMsgContainer *MsgContainer) (bool, []byte, *types.SignedSSVMessage, error) {
	// Decode qbft message
	msg, err := DecodeMessage(signedCommit.SSVMessage.Data)
	if err != nil {
		return false, nil, nil, err
	}

	addMsg, err := commitMsgContainer.AddFirstMsgForSignerAndRound(signedCommit)
	if err != nil {
		return false, nil, nil, errors.Wrap(err, "could not add commit msg to container")
	}
	if !addMsg {
		return false, nil, nil, nil // UponCommit was already called
	}

	// calculate commit quorum and act upon it
	quorum, commitMsgs, err := commitQuorumForRoundRoot(i.State, commitMsgContainer, msg.Root, msg.Round)
	if err != nil {
		return false, nil, nil, errors.Wrap(err, "could not calculate commit quorum")
	}
	if quorum {
		fullData := i.State.ProposalAcceptedForCurrentRound.FullData /* must have value there, checked on validateCommit */

		agg, err := aggregateCommitMsgs(commitMsgs, fullData)
		if err != nil {
			return false, nil, nil, errors.Wrap(err, "could not aggregate commit msgs")
		}
		return true, fullData, agg, nil
	}
	return false, nil, nil, nil
}
```

I don't see us decoding

MatheusFranco99: In the current alan branch:
```go
func (i *Instance) uponPrepare(signedPrepare *types.SignedSSVMessage, prepareMsgContainer *MsgContainer) error {
	// ...

	if !HasQuorum(i.State.CommitteeMember, prepareMsgContainer.MessagesForRound(i.State.Round)) {
		return nil // no quorum yet
	}
        // > Here
	proposalMsgAccepted, err := DecodeMessage(i.State.ProposalAcceptedForCurrentRound.SSVMessage.Data)
	if err != nil {
		return err
	}

	proposedRoot := proposalMsgAccepted.Root
	// ...
}

func (i *Instance) BaseMsgValidation(signedMsg *types.SignedSSVMessage) error {
        // ...
        case PrepareMsgType:
		proposedSignedMsg := i.State.ProposalAcceptedForCurrentRound
		if proposedSignedMsg == nil {
			return errors.New("did not receive proposal for this round")
		}

                // > Here
		proposedMsg, err := DecodeMessage(proposedSignedMsg.SSVMessage.Data)
		if err != nil {
			return errors.Wrap(err, "proposal saved for this round is invalid")
		}

		return validSignedPrepareForHeightRoundAndRootIgnoreSignature(
			signedMsg,
			i.State.Height,
			i.State.Round,
			proposedMsg.Root,
			i.State.CommitteeMember.Committee,
		)
	case CommitMsgType:
		proposedMsg := i.State.ProposalAcceptedForCurrentRound
		if proposedMsg == nil {
			return errors.New("did not receive proposal for this round")
		}
		return validateCommit(
			signedMsg,
			i.State.Height,
			i.State.Round,
			i.State.ProposalAcceptedForCurrentRound,
			i.State.CommitteeMember.Committee,
		)
	case RoundChangeMsgType:
	// ...
}
```

PR #442, opened 2024-07-02
CommitteeRunner: allow signatures of unknown validators

MatheusFranco99:
## Issue

If, in a committee, there is, for some reason, a disparity between known validators, we can hit some edge condition that will prevent the creation of a Beacon object. Take a look at this example:

| Operator | Validators |
|----------|-----------|
| 1 |  A, B |
| 2 |  A, B |
| 3 | A, C |
| 4 |  C |

In the case where an operator fails whenever it sees a message with an unknown validator, no Beacon Object for validator A will ever be created. There could be more realistic scenarios where we lose some fault tolerance where not all operators create the beacon object.

## Fix

This PR changes the CommitteeRunner to allow signatures of unknown validators. For that, we made the following changes:
- `validatePartialSigMsgForSlot` now doesn't check if the validator index exists in the share map.
- new `validateValidatorIndexInPartialSigMsg` function called to validate ValidatorIndex for validator runners.
- In `CommitteeRunner.ProcessPostConsensus`, instead of returning an error in case of an unknown (validator, root) pair, we call "continue" to process remaining (root, validator) pairs. This is due to the edge case in which an operator may not be up to date and not have a (validator, root) pair, even though it received a quorum of signatures for such root.

### Tests:
- [x] Quorum of invalid validator index
- [x] Quorum of invalid and valid validator indexes


ssv/runner_signatures.go
```diff
@@ -43,30 +43,49 @@ func (b *BaseRunner) validatePartialSigMsgForSlot(
...
+	signerInCommittee := false
+	for _, operator := range committee {
+		if operator.Signer == msgSigner {
+			signerInCommittee = true
+			break
 		}
 	}
+	if !signerInCommittee {
+		return errors.New("unknown signer")
+	}
```
GalRogozinski, line 70: according to the flow this would stop processing... won't this defeat our purpose?

I think I said you can leave it only if it won't stop the processing flow

ssv/runner_validations.go
```diff
@@ -19,6 +19,10 @@ func (b *BaseRunner) ValidatePreConsensusMsg(runner Runner, psigMsgs *types.Part
 		return err
 	}
 
+	if err := b.validateValidatorIndexInPartialSigMsg(psigMsgs); err != nil {
+		return err
+	}
+
```
GalRogozinski, line 25: good

ssv/runner_validations.go
```diff
@@ -76,6 +80,11 @@ func (b *BaseRunner) ValidatePostConsensusMsg(runner Runner, psigMsgs *types.Par
 		if err := b.validatePartialSigMsgForSlot(psigMsgs, decidedValue.Duty.Slot); err != nil {
 			return err
 		}
+
+		if err := b.validateValidatorIndexInPartialSigMsg(psigMsgs); err != nil {
+			return err
+		}
+
```
GalRogozinski, line 87: :+1:

ssv/runner_signatures.go
GalRogozinski: even though I see tests are passing so not sure what's going on

MatheusFranco99: This checks if the operator is on the committee. This should be correct independently, no? Or are we thinking of committee changes in terms of operators?

ssv/runner_signatures.go
```diff
@@ -43,30 +43,49 @@ func (b *BaseRunner) validatePartialSigMsgForSlot(
...
-		validatorShare, ok := b.Share[msg.ValidatorIndex]
-		if !ok {
-			return errors.New("unknown validator index")
-		}
+	// Get committee (unique for runner)
+	var shareSample *types.Share
+	for _, share := range b.Share {
+		shareSample = share
+		break
+	}
```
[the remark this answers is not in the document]
MatheusFranco99: Yeah :/

ssv/committee_runner.go
```diff
@@ -236,12 +234,12 @@ func (cr CommitteeRunner) ProcessPostConsensus(signedMsg *types.PartialSignature
...
 			// Get the beacon object related to root
 			validatorObjs, exists := beaconObjects[validator]
-            if !exists {
+			if !exists {
 				anyErr = errors.Wrap(err, "could not find beacon object for validator")
 				continue
 			}
 			sszObject, rexists := validatorObjs[root]
-            if !rexists {
+			if !rexists {
```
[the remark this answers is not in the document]
MatheusFranco99: Pushed a change. Is that what you meant?

ssv/runner_signatures.go
GalRogozinski: yeah you are right

ssv/committee_runner.go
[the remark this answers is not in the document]
GalRogozinski: not exactly...
if you do it on a one line with `;` you will also limit the scope of `exists` to the `if` block...
Currently I think (not sure) you are overshadowing the previous `exists` and this may cause subtle bugs

ssv/spectest/tests/runner/postconsensus/invalid_and_valid_validator_indexes_quorum.go
```diff
@@ -0,0 +1,80 @@
...
+					testingutils.SignPartialSigSSVMessage(ks, testingutils.SSVMsgCommittee(ks, nil, testingutils.PostConsensusInvalidThenValidValidatorIndexAttestationMsg(ks.Shares[1], 1, testingutils.TestingDutySlot))),
+					testingutils.SignPartialSigSSVMessage(ks, testingutils.SSVMsgCommittee(ks, nil, testingutils.PostConsensusInvalidThenValidValidatorIndexAttestationMsg(ks.Shares[2], 2, testingutils.TestingDutySlot))),
+					testingutils.SignPartialSigSSVMessage(ks, testingutils.SSVMsgCommittee(ks, nil, testingutils.PostConsensusInvalidThenValidValidatorIndexAttestationMsg(ks.Shares[3], 3, testingutils.TestingDutySlot))),
+				},
+				OutputMessages: []*types.PartialSignatureMessages{},
+				BeaconBroadcastedRoots: []string{
+					testingutils.GetSSZRootNoError(testingutils.TestingSignedAttestation(ks)),
+				},
+				DontStartDuty: true,
+				ExpectedError: expectedError,
```
moshe-blox, line 35: Returning an error usually means that the execution failed because something unexpected happened.

(for example os.Open would only return an error if the file did not open for some unexpected reason)

I'd suggest to not return an error and instead check that the submitted attestations are only for the known validators.

moshe-blox (approved):

ssv/runner_signatures.go
```diff
@@ -43,30 +43,49 @@ func (b *BaseRunner) validatePartialSigMsgForSlot(
...
+	if shareSample == nil {
+		return errors.New("can not get committee because there is no share in runner")
+	}
```
moshe-blox, line 57: if BaseRunner requires operator ids, it should probably just require them in NewBaseRunne, and then we won't have to do this "hack"

GalRogozinski (approved):

PR #446, opened 2024-07-02
Use naming in complex nested map

MatheusFranco99:
## Overview

This PR improves the readability of the `Signatures` map field of the `PartialSigContainer` structure.

In the map definition, instead of `string`, it uses a new `SigningRoot` type, and, instead of `[]byte`, it uses the existing `types.Signature` type.

### DIFF
--- a/ssv/partial_sig_container.go
+++ b/ssv/partial_sig_container.go
@@ -12,6 +12,7 @@ import (
 type SigningRoot string
 
 type PartialSigContainer struct {
+	// Signature map: validator index -> signing root -> operator id (signer) -> signature (from the signer for the validator's signing root)
 	Signatures map[phase0.ValidatorIndex]map[SigningRoot]map[types.OperatorID]types.Signature
 	// Quorum is the number of min signatures needed for quorum
 	Quorum uint64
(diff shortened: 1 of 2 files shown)


ssv/partial_sig_container.go
```diff
@@ -9,27 +9,30 @@ import (
...
 		Quorum:     quorum,
-		Signatures: make(map[phase0.ValidatorIndex]map[string]map[types.OperatorID][]byte),
+		Signatures: make(map[phase0.ValidatorIndex]map[SigningRoot]map[types.OperatorID]types.Signature),
 	}
 }
 
 func (ps *PartialSigContainer) AddSignature(sigMsg *types.PartialSignatureMessage) {
 	if ps.Signatures[sigMsg.ValidatorIndex] == nil {
-		ps.Signatures[sigMsg.ValidatorIndex] = make(map[string]map[uint64][]byte)
+		ps.Signatures[sigMsg.ValidatorIndex] = make(map[SigningRoot]map[uint64]types.Signature)
```
GalRogozinski, line 30: uint64->operatorID

PR #447, opened 2024-07-02
Take RSA Signer and Verification out of configuration

GalRogozinski:
# Description 

The RSA Signer and Verifier are taken out of the `Config` struct.

*Note:* `TestingVerifier` was deleted and now sig verification on tests is not optimized, causing a delay.


MatheusFranco99 (approved):
Good work!

PR #461, opened 2024-07-11
Simplify comparison methods

MatheusFranco99:
## Overview

This PR fixes the issue with `CompareSignedSSVMessageOutputMessages` and simplifies comparison methods.

For the `CompareSignedSSVMessageOutputMessages` function, it implements the requirement that:
- the broadcasted and expected messages have equal length
- every broadcasted message is linked (equal) to an expected message
- two broadcasted messages are not linked to the same expected message

i.e. a bijection between the lists.

### DIFF
--- a/types/testingutils/compare.go
+++ b/types/testingutils/compare.go
@@ -88,6 +88,24 @@ func ComparePartialSignatureOutputMessagesInAsynchronousOrder(t *testing.T, expe
 	}
 }
 
+func RootCountMapForPartialSignatureMessages(msg *types.PartialSignatureMessages) map[string]int {
+	roots := make(map[string]int)
+
+	for _, partialSigMessage := range msg.Messages {
+		root, err := partialSigMessage.GetRoot()
+		if err != nil {
+			panic(err)
+		}
+		rootStr := hex.EncodeToString(root[:])
+		if _, found := roots[rootStr]; !found {
+			roots[rootStr] = 0
+		}
+		roots[rootStr] += 1
+	}
+
+	return roots
+}
+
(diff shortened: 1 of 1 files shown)


types/testingutils/compare.go
```diff
@@ -147,7 +162,7 @@ func CompareSignedSSVMessageOutputMessages(t *testing.T, expectedMessages []*typ
 	for index, msg := range broadcastedSignedMsgs {
 		r1, _ := msg.GetRoot()
 
-		msg2 := broadcastedSignedMsgs[index]
+		msg2 := expectedMessages[index]
```
GalRogozinski, line 165: why did this change?

types/testingutils/compare.go
```diff
@@ -96,33 +94,50 @@ func ComparePartialSignatureMessages(msg1 *types.PartialSignatureMessages, msg2
...
-		if _, found := roots[hex.EncodeToString(r1[:])]; !found {
-			roots[hex.EncodeToString(r1[:])] = ""
-		} else {
-			roots[hex.EncodeToString(r1[:])] = hex.EncodeToString(r1[:])
+		root := hex.EncodeToString(r2[:])
+		if _, found := roots2[root]; !found {
+			roots2[root] = 0
 		}
+		roots2[root] += 1
 	}
```
GalRogozinski, line 126: there is some copy paste of code... you can use a function

types/testingutils/compare.go
MatheusFranco99: Because `broadcastedSignedMsgs` was being compared to `broadcastedSignedMsgs`, and it's wrong, no?

types/testingutils/compare.go
MatheusFranco99: True. Thankss

GalRogozinski (approved):

PR #467, opened 2024-07-17
feature: (alan) filter committee duties per slot validator duties

y0sher:
## Description 

`Committee` struct holds all the shares for the validators of that committee, though a `CommitteeRunner` only lives for a certain slot and thereby only needs shares of validators that have duty for that slot. 

### Changes

Filter the shares of `Committee` to create a subset that has only shares of validators that have duties for this slot.


ssv/committee.go
```diff
@@ -41,7 +41,14 @@ func (c *Committee) StartDuty(duty *types.CommitteeDuty) error {
 	if _, exists := c.Runners[duty.Slot]; exists {
 		return errors.New(fmt.Sprintf("CommitteeRunner for slot %d already exists", duty.Slot))
 	}
-	c.Runners[duty.Slot] = c.CreateRunnerFn(c.Share)
+	dutyShares := make(map[spec.ValidatorIndex]*types.Share)
+	for _, bduty := range duty.ValidatorDuties {
+		if _, exists := c.Share[bduty.ValidatorIndex]; !exists {
+			return fmt.Errorf("no share for validator %d", bduty.ValidatorIndex)
```
GalRogozinski, line 47: shouldn't this be `continue` simply?
are we sure we want to fail the run just because of one miss?

MatheusFranco99: Yes, from the spec perspective, I would also say `continue` is better. On the other hand, in ssv, it's strange that the operator doesn't the share for the duty since he requested the duty for this validator.

MatheusFranco99 (approved):
@y0sher Approving, but it seems to me that `continue` makes more sense (for the above issue)

ssv/committee.go
y0sher: I agree

y0sher:
> @y0sher Approving, but it seems to me that `continue` makes more sense (for the above issue)

please change as you wish and merge when you're ready.

MatheusFranco99:
@y0sher @GalRogozinski 
Doing the change and adding tests...

ssv/spectest/tests/committee/singleduty/shares_with_fraction_of_duty_validators.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/ssvlabs/ssv-spec/qbft"
+	"github.com/ssvlabs/ssv-spec/ssv"
+	"github.com/ssvlabs/ssv-spec/ssv/spectest/tests"
+	"github.com/ssvlabs/ssv-spec/ssv/spectest/tests/committee"
+	"github.com/ssvlabs/ssv-spec/types"
+	"github.com/ssvlabs/ssv-spec/types/testingutils"
+)
+
+// SharesWithFractionOfDutyValidators performs a complete duty execution for a runner that only has a fraction of the shares for the duty's validators
+func SharesWithFractionOfDutyValidators() tests.SpecTest {
```
GalRogozinski, line 16: Rename to MissingSomeShares

ssv/spectest/tests/committee/singleduty/shares_with_fraction_of_duty_validators.go
```diff
@@ -0,0 +1,121 @@
...
+	committeeShareValidators := []int{1, 3, 5, 7, 9}
+	// KeySet and Share map for Committee
+	committeeShareKSMap := testingutils.KeySetMapForValidatorIndexList(committeeShareValidators)
+	committeeShareMap := testingutils.ShareMapFromKeySetMap(committeeShareKSMap)
+
+	// Duty's validator indexes
+	dutyValidators := []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}
+
+	multiSpecTest := &committee.MultiCommitteeSpecTest{
+		Name: "shares with fraction of duty validators",
```
GalRogozinski, line 32:
```suggestion
		Name: "start committee duty with missing shares",
```

GalRogozinski (approved):

PR #468, opened 2024-07-22
fix: check att and sync duties exist before submitting

y0sher:
### Description

We have notice in the node attempts to send empty duties. calling `SubmitSyncCommitteeMessages` with empty data, resulting in an error.

### Fix 
Check that there is any duty before calling send.


MatheusFranco99 (approved):
LGTM

GalRogozinski (approved):

PR #470, opened 2024-07-30
Update go eth2 lib

GalRogozinski:
# Description

Update go-eth2-lib to v0.21.3


moshe-blox (approved):

go.mod
```diff
@@ -3,35 +3,36 @@ module github.com/bloxapp/ssv-spec
 go 1.19
 
 require (
-	github.com/attestantio/go-eth2-client v0.19.7 // if version is changed, types/generate.go needs adjustment
+	github.com/attestantio/go-eth2-client v0.21.3 // if version is changed, types/generate.go needs adjustment
```
moshe-blox, line 6: didn't we want to upgrade to v0.21.9? @olegshmuelov @GalRogozinski

olegshmuelov: yes we want!

GalRogozinski: I prefer to use the version we tested for no forks

moshe-blox: @GalRogozinski unfortunately we cant because the timeout bug was fixed in v0.21.5: https://github.com/attestantio/go-eth2-client/commit/84ddfd928b95ea5cbd7b2e4ac0a911cc1a04dfbc

Issue #437, opened 2024-06-30
We do slashing checks on sync committees


GalRogozinski:
As part of committee consensus we perform attestation slashing checks on combined attestation/sync-committee data.

Meaning that slashable attestation data will hinder sync committee duty.

Since slashable data is rare and this should optimistically be resolved on the next QBFT round, it isn't this bad.

_Originally posted by @GalRogozinski in https://github.com/ssvlabs/ssv-spec/pull/421#discussion_r1656562403_

GalRogozinski:
This needs to be discussed

GalRogozinski:
So we have a max of 512 sync committee participants. As the network grows this number will become more and more negligible.

So perhaps it makes sense to separate runners for attestations and sync committees so we have a clean protocol that won't run into rare issues.

On the other hand, we can have an optimization instead that all non-slashable duties will use the same consensus

GalRogozinski:
This is easily solvable
In spec we need to initialize valueCheckF with only the validators that do attestation duties

GalRogozinski:
Solved by #471

PR #471, opened 2024-08-04
Value check for beacon vote fix

GalRogozinski:
# Description

The beacon vote value check is now being constructed with clear parameters


ssv/value_check.go
```diff
@@ -11,6 +11,29 @@ import (
...
+// It filters out the slashable public keys from the shares and creates a value check function
+// that checks if the beacon vote is slashable by the filtered public keys
+// It also does sanity checks on the beacon vote data
+// It returns an error if shares assigned to the duty are missing
+func BeaconVoteValueCheckF(duty *types.CommitteeDuty, signer types.BeaconSigner, beaconNetwork types.BeaconNetwork,
+	availableShares map[phase0.ValidatorIndex]*types.Share) (qbft.ProposedValueCheckF, error) {
+	duties := duty.ValidatorDuties
+	slashablePublicKeys := []types.ShareValidatorPK{}
+	for _, duty := range duties {
+		if duty.Type == types.BNRoleAttester {
```
MatheusFranco99, line 24: This is just a little suggestion. For me, it feels easier to reason about it in the following way

```suggestion
	slashablePublicKeys := []types.ShareValidatorPK{}
	for _, duty := range duties {
		if duty.Type == types.BNRoleAttester {
```

ssv/value_check.go
```diff
@@ -11,6 +11,29 @@ import (
...
+			share, ok := availableShares[duty.ValidatorIndex]
+			if !ok {
+				return nil, errors.New("assigned validator duty doesn't have a validator share")
+			}
+			slashablePublicKeys = append(slashablePublicKeys, share.SharePubKey)
+		}
+	}
```
MatheusFranco99, line 31: For sync committee duties, should we also return an error if there's no share for the duty? (Here we do it just for the attestation duty)
Or we prefer leaving this for the partial signature phase?

ssv/value_check.go
GalRogozinski: 🧐 

excuse my silliness, I can't see a difference

GalRogozinski: how is duty outside the for loop?

ssv/value_check.go
GalRogozinski: The point of this logic here is just for slashing protection
it is less relevant for sync committee

ssv/value_check.go
MatheusFranco99: Ups! Sorry, I misread it.

MatheusFranco99 (approved):

GalRogozinski:
Closes #437

ssv/value_check.go
```diff
@@ -11,6 +11,29 @@ import (
...
+// BeaconVoteValueCheckF creates a value check function for a beacon vote
+// It filters out the slashable public keys from the shares and creates a value check function
+// that checks if the beacon vote is slashable by the filtered public keys
+// It also does sanity checks on the beacon vote data
+// It returns an error if shares assigned to the duty are missing
+func BeaconVoteValueCheckF(duty *types.CommitteeDuty, signer types.BeaconSigner, beaconNetwork types.BeaconNetwork,
+	availableShares map[phase0.ValidatorIndex]*types.Share) (qbft.ProposedValueCheckF, error) {
+	duties := duty.ValidatorDuties
+	slashablePublicKeys := []types.ShareValidatorPK{}
+	for _, duty := range duties {
```
y0sher, line 23: `duty` is already defined. shadowing it with another duty is confusing and error prone. Also `duties` is only used one time so no real reason to declare a var.

```suggestion
	slashablePublicKeys := []types.ShareValidatorPK{}
	for _, vduty := range duty.ValidatorDuties {
```

y0sher (changes requested):
I think this change complicates stuff, what was wrong with leaving the previous `BeaconVoteValueCheck` and requiring the caller to pass in only slashable validators? this simple check can be done outside, its encapsulating and hiding the fact that not all `ValidatorDuties` in `CommitteeDuty` are slashable. 

I also don't think the creation of ValueCheckF  should be aware of a specific duty like here, its a dramatic change comparing to how all other value checks work.

In implementation, we are already filtering the slashable validators before passing in to the value check today, IMO spec should do the same and do the filtering in `committee.go`:`StartDuty` function (which passes validators to the value check) instead of encapsulating it here.

GalRogozinski:
@y0sher 
I remember that you did this way in impl after consulting me due to the spec being unclear.
So this is my attempt to make the spec clearer. 

Currently you may notice that besides *testing* code, nothing calls the `ValueCheckF` creation logic. So I must have some way to convey on what we do the slashing checks. I can't simply do the filtering in `committee.go:StartDuty`. The best I was able to do is moving the core logic to `createBeaconValueCheckF`, while `BeaconValueCheckF` does the filtering logic before. Also for testing purposes it is best that `BeaconValueCheckF` will do all operations.

I think the biggest issue at the moment is that before the same `ValueCheckF` function from `qbft.Config` was used throughout all instances. Now we have to reconstruct the function with the `duty` parameter. But the `Controller` object assumes a `Config` that was only created once. So the spec is still misleading.

Do note it is not good to pass `duty` as a parameter to the `valueCheckF` along side data, because the consensus layer shouldn't be aware of Ethereum or SSV logic.

I wasn't sure how to solve this issue before. I have an idea now though... In `CommitteeRunner`, `GetValCheckF` can recreate the function each time. Then the new function can be injected to the Controller's `qbft.Config`. The only bad thing about this solution is that up until now `Config` never mutated after creation. So the SSV impl should take care of concurrency issues.

P.S.
A quick reminder: `duty` is required now in the ValCheck
1. For fetching the slot, since the `BeaconVote` object doesn't carry this data.
2. Fetching the proper slashable validator pubkeys.

Another reminder:
As long as the impl ultimately performs the same logic as spec, it is fine if the design is different.

y0sher:
if you don't want to do it in `committee.go:StartDuty` I think you should just create a separate function that can filter the validators from a `duty` struct and call it outside of valuecheck then passing it anyway. call it wherever you want. 

I think its best to separate the logic for valuecheck, it should only do value check and not filtering/other stuff. or at least make it visible and noticeable in the name of the function that it's not just value check.

GalRogozinski:
@y0sher 
It isn't that I don't want to... in the current design of spec it can't be there.
If I take the filtering out of the function constructor I need to rethink tests... will look into it but not sure it is worth it

PR #478, opened 2024-08-13
Share length validation

MatheusFranco99:
# Overview

This PR adds a validation to the runners' construction function (except by `CommitteeRunner`) to assert that the `share` argument has only one entry (only one share).


ssv/aggregator.go
```diff
@@ -29,7 +29,12 @@ func NewAggregatorRunner(
 	operatorSigner *types.OperatorSigner,
 	valCheck qbft.ProposedValueCheckF,
 	highestDecidedSlot phase0.Slot,
-) Runner {
+) (Runner, error) {
+
+	if len(share) != 1 {
+		return nil, errors.New("must have only one share")
```
GalRogozinski, line 35: if we add an error we should have an accompaning test

MatheusFranco99: On it

GalRogozinski (approved):

y0sher (approved):
Approving this but with an *.

* @GalRogozinski @MatheusFranco99 
I think if this check is required maybe its time to think of separating the base for committee and other runners. Since other runners will never recv more than one share, it looks weird that their func signature can recv a map of shares.

MatheusFranco99:
@y0sher I agree.
I have opened an issue for that to be done later on

ssv/spectest/tests/runner/construction/many_shares.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/ssvlabs/ssv-spec/types"
+	"github.com/ssvlabs/ssv-spec/types/testingutils"
+)
+
+func ManyShares() tests.SpecTest {
+
+	ks := testingutils.KeySetMapForValidators(10)
+	shares := testingutils.ShareMapFromKeySetMap(ks)
+
+	// No errors since one share must be valid for all runners
```
alan-ssvlabs, line 14: Is this comment right for this test?

ssv/spectest/tests/runner/construction/no_shares.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+	"github.com/ssvlabs/ssv-spec/ssv/spectest/tests"
+	"github.com/ssvlabs/ssv-spec/types"
+)
+
+func NoShares() tests.SpecTest {
+
+	shares := map[phase0.ValidatorIndex]*types.Share{}
+
+	// No errors since one share must be valid for all runners
```
alan-ssvlabs, line 13: and here

alan-ssvlabs (approved):

ssv/spectest/tests/runner/construction/many_shares.go
MatheusFranco99: Indeed not. Updating. Thanks for the catch!

ssv/spectest/tests/runner/construction/no_shares.go
MatheusFranco99: Updating, thanks!

PR #480, opened 2024-08-14
Fix static analysis issues

MatheusFranco99:
# Overview

This PR adds fixes for static analysis issues including:
- file permission issues
- file inclusion via variable issues

It also adds a `nosec` comment for a cryptographically insecure PRNG (pseudo-random number generator) used in tests. We use this PRNG (`math/rand`) since the secure `crypto/rand` doesn't allow seeding (which is important for consistent tests).

An existing issue with implicit memory aliasing in a for loop will be automatically fixed by the go update to 1.22.

### DIFF
--- a/qbft/spectest/generate/main.go
+++ b/qbft/spectest/generate/main.go
@@ -93,7 +93,7 @@ func writeJsonStateComparison(name, testType string, post interface{}) {
 
 	file := filepath.Join(scDir, fmt.Sprintf("%s.json", name))
 	log.Printf("writing state comparison json: %s\n", file)
-	if err := os.WriteFile(file, byts, 0444); err != nil {
+	if err := os.WriteFile(file, byts, 0400); err != nil {
 		panic(err.Error())
 	}
 }
@@ -120,7 +120,7 @@ func writeJson(data []byte) {
 
 	file := filepath.Join(basedir, "tests.json")
 	log.Printf("writing spec tests json to: %s\n", file)
-	if err := os.WriteFile(file, data, 0444); err != nil {
+	if err := os.WriteFile(file, data, 0400); err != nil {
 		panic(err.Error())
 	}
 }
--- a/qbft/spectest/generate/main.go
+++ b/qbft/spectest/generate/main.go
@@ -93,7 +93,7 @@ func writeJsonStateComparison(name, testType string, post interface{}) {
 
 	file := filepath.Join(scDir, fmt.Sprintf("%s.json", name))
 	log.Printf("writing state comparison json: %s\n", file)
-	if err := os.WriteFile(file, byts, 0600); err != nil {
+	if err := os.WriteFile(file, byts, 0444); err != nil {
 		panic(err.Error())
 	}
 }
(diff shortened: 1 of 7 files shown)


qbft/spectest/generate/main.go
```diff
@@ -92,7 +93,7 @@ func writeJsonStateComparison(name, testType string, post interface{}) {
 
 	file := filepath.Join(scDir, fmt.Sprintf("%s.json", name))
 	log.Printf("writing state comparison json: %s\n", file)
-	if err := os.WriteFile(file, byts, 0644); err != nil {
+	if err := os.WriteFile(file, byts, 0600); err != nil {
```
GalRogozinski, line 96: I would update to 0444

The reason is that we don't want anyone writing to the file, to make tests pass/fail.
There's nothing confidential here so might as well let everyone read

MatheusFranco99: Yup, I agree

MatheusFranco99: Actually, `gosec` still complains about 0444:
> G306 (CWE-276): Expect WriteFile permissions to be 0600 or less (Confidence: HIGH, Severity: MEDIUM)

Can't it be 0400? Why do we need to allow "read" for group and others?

ssv/spectest/generate/main.go
```diff
@@ -112,7 +113,7 @@ func writeSingleSCJson(path string, testType string, post interface{}) {
 	}
 
 	log.Printf("writing state comparison json: %s\n", file)
-	if err := os.WriteFile(file, byts, 0644); err != nil {
+	if err := os.WriteFile(file, byts, 0444); err != nil {
```
GalRogozinski, line 116: ditto

ssv/spectest/generate/main.go
```diff
@@ -156,7 +157,7 @@ func writeJson(data []byte) {
 	}
 
 	// Write the gzipped data to a file
-	if err := os.WriteFile(file, buf.Bytes(), 0644); err != nil {
+	if err := os.WriteFile(file, buf.Bytes(), 0444); err != nil {
```
GalRogozinski, line 160: ditto

types/spectest/generate/main.go
```diff
@@ -108,7 +109,7 @@ func writeJson(data []byte) {
 	file := filepath.Join(basedir, "tests.json")
 
 	fmt.Printf("writing spec tests json to: %s\n", file)
-	if err := os.WriteFile(file, data, 0644); err != nil {
+	if err := os.WriteFile(file, data, 0444); err != nil {
```
GalRogozinski, line 112: ditto

GalRogozinski (changes requested):
just the permissions

qbft/spectest/generate/main.go
GalRogozinski: hmm 
if CI passes then it can be 0400 honestly
I just think that there is no confidential information and the SSV huts generate jsons themselves.

I just saw no sec concern so you can choose

MatheusFranco99: I also don't see security concerns, but `gosec` complains and I also don't see a reason for the group and others read permission.
Updating it to 0400

GalRogozinski (approved):

alan-ssvlabs (approved):

PR #481, opened 2024-08-14
Secure key storage

MatheusFranco99:
# Overview

This PR adjusts the `PemToPrivateKey` and `PemToPublicKey` functions to use a secure storage method, removing the deprecated functions that were being used.

### DIFF
--- a/types/encryption.go
+++ b/types/encryption.go
@@ -20,7 +20,10 @@ func GenerateKey() ([]byte, []byte, error) {
 	}
 
 	// convert to bytes
-	skPem := PrivateKeyToPem(sk)
+	skPem, err := PrivateKeyToPem(sk)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "Failed to encode private key to pem")
+	}
 	pkPem, err := GetPublicKeyPem(sk)
 	if err != nil {
 		return nil, nil, errors.Wrap(err, "Failed to marshal public key")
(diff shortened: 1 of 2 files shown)


types/encryption.go
```diff
@@ -92,13 +88,17 @@ func PemToPublicKey(pkPem []byte) (*rsa.PublicKey, error) {
 }
 
 // PrivateKeyToPem converts privateKey to pem encoded
-func PrivateKeyToPem(sk *rsa.PrivateKey) []byte {
+func PrivateKeyToPem(sk *rsa.PrivateKey) ([]byte, error) {
+	pemBytes, err := x509.MarshalPKCS8PrivateKey(sk)
```
GalRogozinski, line 92: I see in SSV they use `x509.MarshalPKCS1PrivateKey(sk)`
Why the difference?
This function isn't deprectated?

MatheusFranco99: As I understand, PKCS8 will just add a key-type tag. Follow this [link](https://stackoverflow.com/questions/48958304/pkcs1-and-pkcs8-format-for-rsa-private-key) for more

So, it's more or less the same but it seems that PKCS8 is preferable if we can use it (the audit also suggests it). Therefore, it would be good if SSV also updates it. If not, I can restore to PKCS1

GalRogozinski: The only thing is maybe node stores private keys and if we switch formats it may cause issues...
I can give an approve for now... but we should wait for @moshe-blox or @y0sher to read this comment and give their opinion

GalRogozinski (approved):

types/encryption.go
GalRogozinski: Btw, to be honest... this is an example of where we have too much implementation details in spec...
A spec probably shouldn't care about whether the key is PKCS1 or PKCS8 encoded

MatheusFranco99: Agree

types/encryption.go
```diff
@@ -49,39 +52,32 @@ func Encrypt(pk *rsa.PublicKey, plainText []byte) ([]byte, error) {
...
-		if err != nil {
-			return nil, errors.Wrap(err, "Failed to decrypt private key")
-		}
+	if block == nil {
+		return nil, errors.New("failed to decode PEM")
 	}
-	parsedSk, err := x509.ParsePKCS1PrivateKey(b)
+
+	// Parse key as PKCS8
+	parsedSk, err := x509.ParsePKCS8PrivateKey(block.Bytes)
```
y0sher, line 60: AFAIK PKCS8 isn't backward compatible with PKCS1.. wouldn't this break stuff?

y0sher (approved):
@GalRogozinski @MatheusFranco99 
(@moshe-blox, @lior-blox ).

I think I can approve this for the spec as I understand why you want this change. But I can't tell if and when we'll be able to adapt in the node. So maybe its best to wait with merging this.

types/encryption.go
MatheusFranco99: Correct
It depends on how these functions are being used. From your above comment, it seems that it would break indeed. So, I'll switch back to PKCS1

MatheusFranco99:
@y0sher 
Sure
Adding back the PKCS1 due to the break issues

y0sher:
@MatheusFranco99 I would create an issue about switching to PKCS8 in the future if its better in any terms, we just need to to this carefully and think about it from a product perspective.

MatheusFranco99:
@y0sher Agree 100%. Thanks

alan-ssvlabs (approved):
Agree with the discussions above

GalRogozinski:
@y0sher 
there is no hinderance now that we switched back to PKCS1 and this can be merged from your perspective?

y0sher:
Other than that looks good to me.

GalRogozinski:
Btw @MatheusFranco99 and @alan-ssvlabs @y0sher, I don't see a reason to open an issue with PKCS8 unless someone can tell me a clear advantage of it that is relevant to us.
I understood the tags will be helpful if we ever change our encryption algo, when we do that the encoding can change.

PR #484, opened 2024-08-15
Sort signers in decided message

MatheusFranco99:
# Overview

This PR adds the feature of sorting the signers in the decided message.

It adds a "sorted decided" test to test the sorting upon receiving commit messages in an unsorted order.

Closes https://github.com/ssvlabs/ssv-spec/issues/472


ssv/runner_validations.go
```diff
@@ -50,7 +50,7 @@ func (b *BaseRunner) ValidatePostConsensusMsg(runner Runner, psigMsgs *types.Par
 	}
 
 	// TODO https://github.com/ssvlabs/ssv-spec/issues/142 need to fix with this issue solution instead.
-	if b.State.DecidedValue == nil || len(b.State.DecidedValue) == 0 {
```
GalRogozinski, line 53: that's from another PR no?
you can resolve when solving conflicts

MatheusFranco99: Due to the linter update, I have been fixing this lint issue in all the recent PRs

qbft/spectest/tests/commit/sorted_decided.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"crypto/rsa"
+
+	"github.com/ssvlabs/ssv-spec/qbft/spectest/tests"
+	"github.com/ssvlabs/ssv-spec/types"
+	"github.com/ssvlabs/ssv-spec/types/testingutils"
+)
+
+// SortedDecided tests the creation of the decided message that should have sorted signers
+func SortedDecided() tests.SpecTest {
```
GalRogozinski, line 12: no output messages?

GalRogozinski (changes requested):

qbft/spectest/tests/commit/sorted_decided.go
MatheusFranco99: Notice that `ControllerSpecTest.OutputMessages` is unused. For this test, we actually use `ControllerSpecTest.RunInstanceData.ExpectedDecidedState.BroadcastedDecided` to test the broadcasted decided message.

MatheusFranco99: Notice that not only does none of the implemented tests define it but the spec test itself doesn't use it.

As I see it, the reason is that the controller spec test wants to test decided values and the decided messages for a list of instances (and not the instance's broadcasted messages).

I think that we should either drop this field or implement it in the tests if wanted.

GalRogozinski (approved):

y0sher (approved):
lgtm

alan-ssvlabs (approved):

PR #487, opened 2024-08-19
Stop processing after decided

MatheusFranco99:
# Overview

This PR makes the `qbft.Controller` and `ssv.Runner` stop processing consensus messages after the consensus instance is decided.

Closes https://github.com/ssvlabs/ssv-spec/issues/473


qbft/controller.go
```diff
@@ -105,16 +105,16 @@ func (c *Controller) UponExistingInstanceMsg(msg *ProcessingMessage) (*types.Sig
 
 	prevDecided, _ := inst.IsDecided()
 
+	// if previously decided, we don't process more messages
+	if prevDecided {
```
y0sher, line 109: does this include only consensus messages? we want to keep processing post-consensus message.s.

MatheusFranco99: Yup, includes only consensus messages because this is already at the `qbft.Controller` level, so no partial sig msg should reach here

alan-ssvlabs (approved):

y0sher (approved):

PR #490, opened 2024-08-21
remove ssz encoding for CommitteeMember and Share

### LINKED ISSUE #419: We encode Operator and Share even though we don't pass them on the wire
Could it be because of we use the encoding to register validators?
If so this should be documented

alan-ssvlabs:
Fixes #419 

ssz encoding for CommitteeMember and Share are defined but not used, therefore this PR removes the encoding functions.

For CommitteeMember encoding no tests are designed, therefore only the functions are removed. 
For Share encoding, there are tests defined for this function. The tests are also removed as the function itself is actually unused.


types/spectest/run_test.go
```diff
@@ -79,12 +79,12 @@ func TestJson(t *testing.T) {
...
-				require.NoError(t, err)
-				typedTest := &share.EncodingTest{}
-				require.NoError(t, json.Unmarshal(byts, &typedTest))
-				typedTest.Run(t)
+			// case reflect.TypeOf(&share.EncodingTest{}).String():
+			// 	byts, err := json.Marshal(test)
+			// 	require.NoError(t, err)
+			// 	typedTest := &share.EncodingTest{}
+			// 	require.NoError(t, json.Unmarshal(byts, &typedTest))
+			// 	typedTest.Run(t)
```
MatheusFranco99, line 87: Since we are removing this test type, you can actually delete these lines :)

MatheusFranco99 (changes requested):
Good job! I left just a little request

types/spectest/run_test.go
alan-ssvlabs: You are right, removed

alan-ssvlabs (commented):
remove commented code

MatheusFranco99 (approved):
Good job

GalRogozinski (approved):

y0sher (approved):
We are encoding shares for the database using SSZ in the implementation, but I guess we'll have to implement it in the implementation since its not used here in the spec.

Issue #439, opened 2024-06-30
`DomainType` fits better in config than in committeeMember ( or share for that matter) since its used for everything under the same network and we're not going to have different domainTypes for different shares/members ever unless its a bug.


GalRogozinski:
`DomainType` fits better in config than in committeeMember ( or share for that matter) since its used for everything under the same network and we're not going to have different domainTypes for different shares/members ever unless its a bug.

_Originally posted by @y0sher in https://github.com/ssvlabs/ssv-spec/pull/433#pullrequestreview-2147352256_

GalRogozinski:
One solution is to add another `config` for SSV. And then have duplicates in SSV and QBFT.
That's fine because they are different packages.

PR #494, opened 2024-08-28
Move domainType from CommitteeMember and Share to Network

alan-ssvlabs:
domainType was stored in CommitteeMember and Share, this PR creates a config for the ssv package and stores domainType in config. Config is stored in BaseRunner, as all runners should have the same domainType.

Json files are regenerated and modified as a result of the change of CommitteeMember, Share, and BaseRunner structures.


qbft/state.go
```diff
@@ -5,6 +5,8 @@ import (
 )
 
 type IConfig interface {
+	// GetSignatureDomainType returns the Domain type used for signatures
+	GetSignatureDomainType() types.DomainType
```
MatheusFranco99, line 9: Why this function has been added in `IConfig`? I see that the `Config` structure also has it but it seems unused.

If that's the case we can actually drop it all. Maybe in another PR since it's not the scope of this PR. If you agree, pls create an issue so we can solve this later :)

types/testingutils/runner.go
```diff
@@ -156,6 +156,7 @@ var ConstructBaseRunnerWithShareMap = func(role types.RunnerRole, shareMap map[p
 			km,
 			opSigner,
 			valCheck,
+			&ssv.Config{Domain: TestingSSVDomainType},
```
MatheusFranco99, line 159: I think we could use a variable here since it seems that all other configs are equal, right? *Similar as you did in the next function
Or is there a purpose for it to be created for every runner?

MatheusFranco99 (commented):
Good work! Just some questions

GalRogozinski (commented):
It seems that moving this to `Config` is somewhat incorrect. This is because `Config` in our case is a `qbft` config, and shouldn't be used as part of SSV.

Imo it is better to move this to be something returned by the `Network` interface (in `types.go`). This is because `Network` is an interface that is used by both the `ssv` and `qbft` modules. 

@moshe-blox and @MatheusFranco99... wdyt?

I also take this opportunity to note that changing `DomainType` dynamically (due to forks or network switch) isn't covered by the spec. The spec should always assumes a node running on a specific network for a specific fork.

MatheusFranco99:
@GalRogozinski 
It seems fine for me to put DomainType into Network.

qbft/state.go
alan-ssvlabs: You are right, dropping this

alan-ssvlabs:
@GalRogozinski @MatheusFranco99 Now domainType is in Network instead of config, I think it is indeed a better place for it. Please check, thank you!

types/testingutils/network.go
```diff
@@ -10,6 +10,7 @@ type TestingNetwork struct {
 	BroadcastedMsgs []*types.SignedSSVMessage
 	OperatorID      types.OperatorID
 	OperatorSK      *rsa.PrivateKey
+	Domain          types.DomainType
```
MatheusFranco99, line 13: In the `NewTestingNetwork` function, this field is unassigned. I believe you could initialise it as a TestingSSVDomaintType. However, it's strange to me that the tests are still passing even though this is unassigned.

MatheusFranco99 (approved):

GalRogozinski:
Fixes #439

GalRogozinski (approved):

GalRogozinski (approved):

y0sher (approved):
lgtm, it's about time :)

PR #503, opened 2024-09-17
Fix GetCommitteeId

MatheusFranco99:
# Overview

This PR fixes a possible overflow in the `GetCommitteeID` function.

Closes https://github.com/ssvlabs/ssv-spec/issues/501 .


GalRogozinski (approved):
Note that the underlying issue stems for `OperatorID` being too large (64 bits)...
Since it is transferred on wire it can be 32 bits or even lower.

Having said that, we can always pad OperatorID in the future for committee hash calculations. There's no performance downside to my understanding since the padded input is still smaller than SHA's block size. It will just look a little funny and raise eyebrows.

AKorpusenko (approved):

MatheusFranco99:
This PR is being closed since the conversion won't overflow (since OperatorID << MaxInt32 and it's incremented by contract), and it would cause a network fork.

PR #505, opened 2024-09-23
Add GoSec action and fix issues

MatheusFranco99:
# Overview

This PR adds the GoSec GitHub action and fixes issues raised by GoSec.

Note that there is still one issue raised by GoSec. This is already solved by #503 . To avoid double work and review efforts, we didn't add it to this PR.


types/beacon_types.go
```diff
@@ -231,16 +231,44 @@ func (n BeaconNetwork) EstimatedCurrentSlot() spec.Slot {
...
+	// Get delta time
+	genesis := n.MinGenesisTime()
+	deltaTime := uint64(time) - genesis
+
+	// Get slot duration
+	slotDurationInSeconds := int64(n.SlotDurationSec().Seconds())
+	// Sanitize slot duration
+	if slotDurationInSeconds <= 0 {
+		return spec.Slot(math.MaxUint64)
+	}
```
GalRogozinski, line 248: I am wondering whether we need all of this in spec, especially since Slot Duration should be a constant...
GoSec yells if it isn't there?

types/beacon_types.go
```diff
@@ -231,16 +231,44 @@ func (n BeaconNetwork) EstimatedCurrentSlot() spec.Slot {
...
+
+	// Get genesis time
+	minGenesisTime := n.MinGenesisTime()
+
+	// Sanitize variables
+	if minGenesisTime > uint64(math.MaxInt64) || d > uint64(math.MaxInt64) {
+		return int64(math.MaxInt64)
+	}
+
+	return int64(minGenesisTime) + int64(d)
```
GalRogozinski, line 271: converting from uint64 to int64 is considered safe by gosec?
A big number can turn negative I think (no that it will happen but you know)

types/testingutils/share_map.go
```diff
@@ -17,20 +17,31 @@ func ValidatorIndexList(limit int) []int {
 	return ret
 }
 
-func KeySetMapForValidatorIndexList(valIndexes []int) map[phase0.ValidatorIndex]*TestKeySet {
+func KeySetMapForValidatorIndexList(valIndexes []phase0.ValidatorIndex) map[phase0.ValidatorIndex]*TestKeySet {
```
GalRogozinski, line 20: nice!

types/testingutils/share_map.go
```diff
@@ -17,20 +17,31 @@ func ValidatorIndexList(limit int) []int {
...
+	}
+	validators := make([]phase0.ValidatorIndex, limit)
+	for i := 0; i < limit; i++ {
+		validatorIndex := (i + 1)
+		if validatorIndex < 0 {
+			panic("Invalid validator index")
+		}
+		validators[i] = phase0.ValidatorIndex(validatorIndex)
+	}
+	return KeySetMapForValidatorIndexList(validators)
```
GalRogozinski, line 44: why not change `ValidatorIndexList`?

GalRogozinski (commented):
Hey,
jsons changed only because of #503?
It is best to have all changes that change spec jsons in seperate PR...

GalRogozinski:
Hmm technically it will be best to simply use correct types in the first place that will require less casting?
It is just that I think this PR made spec a bit less readable... nevertheless I think we should merge the changes and maybe later fix types... since this will put a strain on the engineering team maybe?

MatheusFranco99:
@GalRogozinski 
> jsons changed only because of https://github.com/ssvlabs/ssv-spec/pull/503?
It is best to have all changes that change spec jsons in seperate PR...

I believe so. Merging 503 first would be great. Since I wanted this PR to pass tests, I had to also change here. If 503 is merged, these changes will disappear.

types/beacon_types.go
MatheusFranco99: Notice that, before the conversion, sanity checks ensure it'll be converted properly. GoSec sees this. If I remove the checks, it will raise the error.

types/testingutils/share_map.go
MatheusFranco99: It would incur changes through many tests. I can definitely do it here if you want. But, since it's not the goal of this PR, I decided to leave this for a future PR.

types/beacon_types.go
MatheusFranco99: I agree. But, yes, GoSec yells (and with reason tbh 😅)

MatheusFranco99:
> mm technically it will be best to simply use correct types in the first place that will require less casting?

I definitely agree. Some return values should clearly be `uint`, but they are `int` and keep requiring casting due to others correctly defined `uint` types. I decided not to change the functions' outputs because I believe this would incur several other changes on the ssv repo. If this is wanted, and they have time to align with it, I can add it to this PR. But, personally, I agree that we should merge this and later fix these types when they will have time to align.

MatheusFranco99:
@GalRogozinski Removed the change on GetCommitteeID as you asked for

types/testingutils/beacon_node.go
```diff
@@ -419,6 +419,9 @@ var TestingProposerDutyFirstSlot = types.ValidatorDuty{
 }
 
 func getValPubKeyByValIdx(valIdx int) phase0.BLSPubKey {
+	if valIdx < 0 {
```
alan-ssvlabs, line 422: Can we just use uint for valIdx?

alan-ssvlabs (commented):
We add quite a few (maybe practically meaningless) check for whether ints are negative. Would the better fix be updating them to uints?

types/testingutils/beacon_node.go
MatheusFranco99: `phase0.ValidatorIndex` is uint64 so I believe I will need to do the conversion either way. But I think your question extends easily to why not use uint64 here as well, right?

Since the goal of this PR was just to solve GoSec issues, I limited myself to this scope. But I agree that in a later PR, we should refactor and use the correct typing where possible, including in these test functions. But this would require much more work now (and I guess there's some priority to fix this and not incur more changes for the `ssv` repo).

alan-ssvlabs (approved):

types/beacon_types.go
GalRogozinski: I think... that if the sum of `int64(minGenesisTime) + int64(d)` will be larger than `math.MaxInt64` there will be an overflow... weird that go-sec didn't shout about that...

Anyhow, using some safe arithmetics here is an overkill since we both know that the above scenario will never happen and we don't want to have too many impl details in spec...
So if they shout I will try to talk to them about this

MatheusFranco99: By "they shout" you mean the implementation team?

GalRogozinski:
@alan-ssvlabs 
see https://github.com/ssvlabs/ssv-spec/pull/505#issuecomment-2372512928

GalRogozinski (approved):

PR #511, opened 2024-11-16
Electra: Attestation, Aggregation and Proposal Alignment

MatheusFranco99:
## Overview

This PR aligns the construction of attestations, aggregations and proposals to the latest changes by the [electra fork](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md).

### Proposal

Since a **versioned** data was already fetched and submitted to the beacon node, it was only necessary to accept the Electra version in the `ValidatorConsensusData` validation.

### Attestation

The changes follow the [ethereum electra spec for constructing attestations](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/validator.md#attesting).

- Starts using the  [`spec.VersionedAttestation`](https://github.com/attestantio/go-eth2-client/blob/electra/spec/versionedattestation.go) type as the output of the attestation duty. For `DataVersionElectra`, it represents a [`electra.Attestation`](https://github.com/attestantio/go-eth2-client/blob/electra/spec/electra/attestation.go) object. For previous versions, it represents a [`phase0.Attestation`](https://github.com/attestantio/go-eth2-client/blob/electra/spec/phase0/attestation.go).
- For creating the input for consensus, the `CommitteeRunner` still fetches a [`phase0.AttestationData`](https://github.com/attestantio/go-eth2-client/blob/electra/spec/phase0/attestationdata.go) type since it remains unchanged and is sufficient for constructing the `BeaconVote`. However, when being constructed after electra, the `Index` field that identifies the committee is set to 0 ([EIP-7549](https://eips.ethereum.org/EIPS/eip-7549)).

> [!NOTE]
>  The Ethereum specification states that, under electra, a [`electra.SingleAttestation`](https://github.com/attestantio/go-eth2-client/blob/electra/spec/electra/singleattestation.go) object should be returned. However, the `CommitteeRunner` outputs a [`spec.VersionedAttestation`](https://github.com/attestantio/go-eth2-client/blob/electra/spec/versionedattestation.go) to the `BeaconNode` as this is the type specified in the [`go-eth2-client`](https://github.com/attestantio/go-eth2-client) library. Internally, the library [transforms](https://github.com/attestantio/go-eth2-client/blob/electra/spec/electra/attestation.go#L176) the versioned object into a `electra.SingleAttestation` later on.

> [!NOTE]
>  The `CommitteeRunner` will keep running consensus over a `BeaconVote`. Even though it doesn't include a `DataVersion`, operators are assumed to compute the correct `DataVersion` given the duty's slot.

### Aggregation

The changes follow the [ethereum electra spec for constructing aggregations](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/validator.md#attestation-aggregation).

- Starts using the [`VersionedSignedAggregateAndProof`](https://github.com/attestantio/go-eth2-client/blob/electra/spec/versionedaggregateandproof.go) type from the `go-eth2-client` library. For `DataVersionElectra`, it represents a [`electra.SignedAggregateAndProof`](https://github.com/attestantio/go-eth2-client/blob/electra/spec/electra/aggregateandproof.go) object. For previous versions, it represents a [`phase0.SignedAggregateAndProof`](https://github.com/attestantio/go-eth2-client/blob/electra/spec/phase0/aggregateandproof.go).

### Beacon Node

- `SubmitAttestation` now accepts `spec.VersionedAttestation` rather than `phase0.Attestation`.
- `SubmitSignedAggregateSelectionProof` now accepts `spec.VersionedSignedAggregateAndProof` rather than `phase0.SignedAggregateAndProof`.
- `SubmitAggregateSelectionProof` now should return either a `phase0.SignedAggregateAndProof` or a `electra.SignedAggregateAndProof`, depending on the version.
- A new method, `DataVersion(epoch phase0.Epoch)`, is added to return the `DataVersion` given the epoch. Note that, for this frequent API call, it's recommended for the implementation to use a cache.

### Tests

All tests for the Aggregator duty and Committee duty (that encompasses the Attester duty) became versioned with a test for version `DataVersionPhase0` and another for `DataVersionElectra`. Tests for the Proposal duty now includes the Electra version.

### Testing Utils

Major refactoring for better separating specific duties utils, and for versioning utils for the aggregator and committee duties.


qbft/round_robin_proposer.go
```diff
@@ -7,16 +7,11 @@ import "github.com/ssvlabs/ssv-spec/types"
 // Each new height has a different first round proposer which is +1 from the previous height.
 // First height starts with index 0
 func RoundRobinProposer(state *State, round Round) types.OperatorID {
```
oleg-ssvlabs, line 9: Wdyt about changing the function parameters to something like this:

`RoundRobinProposer(height Height, round Round, committee []types.OperatorID)`

I think it’s better because the function explicitly defines what it works with, rather than taking a large object that contains other values unused by this function. Also, I believe [this](https://github.com/ssvlabs/ssv/blob/main/message/validation/consensus_validation.go#L443) was a workaround, because the `State` object is not accessible in the place where this function is supposed to be called. Changing the signature as suggested would allow us to eliminate that workaround and use this function instead

MatheusFranco99: Hey @oleg-ssvlabs , thanks for the suggestion!
I think it's a good change. Indeed, this signature has caused some unnecessary work before (e.g. in msg validation).
@GalRogozinski can I add a PR for it?

ssv/committee_runner.go
```diff
@@ -401,12 +408,11 @@ func (cr *CommitteeRunner) expectedPostConsensusRootsAndBeaconObjects() (
 		case types.BNRoleAttester:
 
 			// Attestation object
-			attestationData := constructAttestationData(beaconVote, validatorDuty)
-			aggregationBitfield := bitfield.NewBitlist(validatorDuty.CommitteeLength)
-			aggregationBitfield.SetBitAt(validatorDuty.ValidatorCommitteeIndex, true)
-			unSignedAtt := &phase0.Attestation{
-				Data:            attestationData,
-				AggregationBits: aggregationBitfield,
+			dataVersion := cr.beacon.DataVersion(cr.beacon.GetBeaconNetwork().EstimatedEpochAtSlot(validatorDuty.Slot))
```
y0sher, line 411: all the duties will be in the same epoch so we could do this call one time outside of the loop. 
we could potentially even avoid this call at all since now in Electra `DataVersion` is returned from AttestationData call.

MatheusFranco99: We discussed this possibility. Indeed, we get a version with the AttestationData, but to use it later on, we would need to store it in the state, which we didn't want to change
Yes, I can optimize it by calling outside the loop
Thanks!

MatheusFranco99: Added. Thx again

MatheusFranco99: Added. Thx again

GalRogozinski (commented):
You have long switch case statements that go over all block versions... but we only need two in practice: Deneb and Electra.

so maybe reduce the code?

Also optionally (not a blocker) it will be nice to have all the Ethereum specific logic in another file, separate from committee runner. This is because those are specifications that don't relate to our protocol

ssv/committee_runner.go
```diff
@@ -483,12 +494,100 @@ func (cr CommitteeRunner) GetOperatorSigner() *types.OperatorSigner {
...
+	case spec.DataVersionAltair:
+		return &spec.VersionedAttestation{Version: dataVersion, Altair: ConstructPhase0AttestationWithoutSignature(attestationData, validatorDuty)}, nil
+	case spec.DataVersionBellatrix:
+		return &spec.VersionedAttestation{Version: dataVersion, Bellatrix: ConstructPhase0AttestationWithoutSignature(attestationData, validatorDuty)}, nil
+	case spec.DataVersionCapella:
+		return &spec.VersionedAttestation{Version: dataVersion, Capella: ConstructPhase0AttestationWithoutSignature(attestationData, validatorDuty)}, nil
+	case spec.DataVersionDeneb:
+		return &spec.VersionedAttestation{Version: dataVersion, Deneb: ConstructPhase0AttestationWithoutSignature(attestationData, validatorDuty)}, nil
+	case spec.DataVersionElectra:
+		return &spec.VersionedAttestation{Version: dataVersion, Electra: ConstructElectraAttestationWithoutSignature(attestationData, validatorDuty)}, nil
```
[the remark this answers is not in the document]
MatheusFranco99: Added

y0sher (approved):

GalRogozinski (approved):

types/testingutils/beacon_node_consts.go
```diff
@@ -1,5 +1,103 @@
 package testingutils
 
+// TO-DO: use json string from a test in api/v1/electra of go-eth2-client
+// Beacon block taken from https://dora.pectra-devnet-6.ethpandaops.io/slot/58430
+// To fill the electra.BlockContents, just get the KZGProofs and Blobs from the deneb block
```
GalRogozinski, line 5: don't just write TODOs, open an issue

GalRogozinski (commented):
Good job!
After you resolve conflicts and go over my comments Alan must review. This needs to be approved on Friday.

types/consensus_data_encoding.go
```diff
@@ -1,5 +1,5 @@
 // Code generated by fastssz. DO NOT EDIT.
-// Hash: da7f54b805a60d108e45282c8eafc508bfd48cbce5a494a521cabc5c72f19bc4
+// Hash: 5cba5a2ddfd055def50b036d4df3d79bfcd63163b70897c4938b39e571e1e761
```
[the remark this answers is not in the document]
MatheusFranco99: Correct, I've changed the `consensus_data.go` file

ssv/aggregator.go
```diff
@@ -318,3 +321,47 @@ func (r *AggregatorRunner) GetSigner() types.BeaconSigner {
...
+	case spec.DataVersionBellatrix:
+		ret.Bellatrix = &phase0.SignedAggregateAndProof{
+			Message:   aggregateAndProof.Bellatrix,
+			Signature: signature,
+		}
+	case spec.DataVersionCapella:
+		ret.Capella = &phase0.SignedAggregateAndProof{
+			Message:   aggregateAndProof.Capella,
+			Signature: signature,
+		}
```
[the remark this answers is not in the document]
MatheusFranco99: For this structure, I'm the same pattern as in go-eth2-client.

types/consensus_data.go
```diff
@@ -203,6 +205,12 @@ func (ci *ValidatorConsensusData) GetBlockData() (blk *api.VersionedProposal, si
 			return nil, nil, errors.Wrap(err, "could not unmarshal ssz")
 		}
 		return &api.VersionedProposal{Deneb: ret, Version: ci.Version}, ret.Block, nil
+	case spec.DataVersionElectra:
```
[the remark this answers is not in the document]
MatheusFranco99: I think it's worth keeping it for testnets, even for simulation/private ones, that could use an old version

types/consensus_data.go
```diff
@@ -223,17 +231,71 @@ func (ci *ValidatorConsensusData) GetBlindedBlockData() (*api.VersionedBlindedPr
...
+		}
+
+		return &spec.VersionedAggregateAndProof{Version: ci.Version, Bellatrix: ret}, ret, nil
+
+	case spec.DataVersionCapella:
+		ret := &phase0.AggregateAndProof{}
+		if err := ret.UnmarshalSSZ(ci.DataSSZ); err != nil {
+			return nil, nil, errors.Wrap(err, "could not unmarshal ssz")
+		}
+
```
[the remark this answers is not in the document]
MatheusFranco99: I see, but, for this structure, I'm doing the same pattern as in go-eth2-client.

types/testingutils/beacon_node_consts.go
MatheusFranco99: Indeed, thanks!

types/messages_encoding.go
```diff
@@ -25,7 +25,6 @@ func (s *SSVMessage) MarshalSSZTo(buf []byte) (dst []byte, err error) {
 
 	// Offset (2) 'Data'
 	dst = ssz.WriteOffset(dst, offset)
-	offset += len(s.Data)
```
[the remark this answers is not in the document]
GalRogozinski: I thought if the encoding file changes then the hash should change.
But hte hash is for the original file the encodign is derived from

GalRogozinski (approved):
Good Job!

PR #491, opened 2024-08-22
Message Validation

MatheusFranco99:
# Overview

This PR updates the message validation module.

It adds a `MessageValidator` structure that implements the `ValidatorForTopic` function, as defined by the libp2p framework.

Its main validation function starts a chain of validation checks for the pub-sub, syntax, semantics, QBFT and duty logic rules.

## Documentation and message rate estimation

The PR also updates the message validation and scoring documentation in `p2p/SPEC.md` and `p2p/SCORING.md`. The go code for the message rate estimation is added (along with tests) since this estimation is referenced by the scoring documentation.

## Tests

This PR also adds tests for each implemented rule, as can be seen in the following list.

### PubSub
- [x] PubSub message with empty data
- [x] PubSub message with a data bigger than the allowed size
- [x] PubSub message that can't decode a SignedSSVMessage

### Syntax validation
- [x] No signers
- [x] No signatures
- [x] Wrong RSA signature size
- [x] Signers not sorted
- [x] Zero signer
- [x] Duplicated signer
- [x] Signers and signatures with different lengths
- [ ] Nil SSVMessage. This test can't be implemented due to the internals of `ssz` decoding
- [x] Empty data
- [ ] SSVMessage with data above the allowed limit. This test can't be implemented due to `ssz` encoding errors.
- [x] Undecodable SSVMessage.Data for consensus message
- [x] Undecodable SSVMessage.Data for partial signature message
- [x] Undecodable PrepareJustificaiton
- [x] Undecodable RoundChangeJustification

### General semantics validation
- [x] Signer not in committee
- [x] Wrong domain
- [x] Invalid role
- [x] Unknown validator
- [x] Validator not attesting
- [x] Validator liquidated
- [x] Non-xistent CommitteeID
- [x] Incorrect topic
- [x] DKG Message
- [x] Unknown MessageType

### Consensus semantics validation
- [x] Non-decided with multiple signers
- [x] Decided with not enough signers
- [x] Prepare with FullData
- [x] Commit with FullData
- [x] Invalid root
- [x] Unknown QBFT MessageType
- [x] Zero round
- [x] Mismatched identifier

### Consensus QBFT logic validation
- [x] Signer not leader in proposal message
- [x] Decided with the same signers as previous message
- [x] Duplicated proposal with different FullData
- [x] Duplicated consensus message
- [x] Round not allowed considering time spread
- [x] Round already advanced
- [x] Unexpected RoundChangeJustification in Prepare message
[...]

PR #513, opened 2024-11-28
Validators per operator cap: 500 to 1000

MatheusFranco99:
## Overview

This PR implements the spec changes associated with the update of the number of validators per operator from 500 to 1000.

Namely, it changes the number of signatures allowed in the `PartialSignatureMessages` structure.


GalRogozinski (approved):
Just in the description. link to the SIP

GalRogozinski:
@MatheusFranco99 
btw, please add a PR on top of #491 
don't change 491 since it will be a fork

alan-ssvlabs (approved):

MatheusFranco99:
@GalRogozinski 
My implementation is already adjusted to min(2*validators, validators + sync_committee_size). So I think we don't need a PR on top of it

moshe-blox (approved):
LGTM 🔥

PR #518, opened 2025-01-30
Stateless proposer function

MatheusFranco99:
## Overview

This PR makes the `RoundRobinProposer` function stateless by passing as arguments the height and list of operators instead of a state.


alan-ssvlabs (approved):
LGTM

PR #522, opened 2025-02-18
fix json generation

MatusKysel:
After recent change in spec, we need to run `go generate` to get tests.json file used for testing in ssv. But because of `runtime.Caller(0)` generating json outside the repo is impossible. Whit this proposed change we can run it in any directory, where the binary used for generating code is located

### DIFF
--- a/types/spectest/generate/main.go
+++ b/types/spectest/generate/main.go
@@ -103,7 +103,7 @@ func writeJsonStateComparison(name, testType string, post interface{}) {
 
 	file := filepath.Join(scDir, fmt.Sprintf("%s.json", name))
 	log.Printf("writing state comparison json: %s\n", file)
-	if err := os.WriteFile(file, byts, 0400); err != nil {
+	if err := os.WriteFile(file, byts, 0664); err != nil {
 		panic(err.Error())
 	}
 }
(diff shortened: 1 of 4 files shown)


alan-ssvlabs (approved):

GalRogozinski (changes requested):
I think you should use `filepath tools`
I think they are not system dependent..

I think currently it is possible to work on spec tests from Windows (never tried it)

MatheusFranco99:
@MatusKysel, you've made the change for the `qbft` and `ssv` modules, but not for the `types` one. Is it not tested on your side? I think that, even so, we may adjust it in a similar manner for consistency.

MatheusFranco99 (approved):

types/spectest/generate/main.go
```diff
@@ -126,26 +109,11 @@ func writeJsonStateComparison(name, testType string, post interface{}) {
...
-	if !ok {
-		panic("no caller info")
-	}
-	basedir = strings.TrimSuffix(basedir, "main.go")
-
-	// try to create directory if it doesn't exist
-	_ = os.Mkdir(basedir, os.ModeDir)
-
-	file := filepath.Join(basedir, name+".json")
+	file := name + ".json"
```
GalRogozinski, line 116: here you didn't change

GalRogozinski (changes requested):
I know this isn't on you but can you change permissions to everytime we write file to (0664). This should solve issues when we run generate several times

GalRogozinski (approved):

PR #523, opened 2025-02-20
Fix test entropy

MatheusFranco99:
## Overview

This PR fixes the test entropy, created in the Electra change, that uses randomly sorted validator indexes in messages.


GalRogozinski (changes requested):
When we do the tests itself it is arguably fine to copy paste
but we shouldn't do it for writing utils code

maybe create a function called SortedMapKeys that uses [generics](https://go.dev/blog/intro-generics), and put it in a utils file:
```
// SortMapByKey sorts a map by its keys and returns a sorted slice of key-value pairs.
func SortedMapKeys[K constraints.Ordered, V any](m map[K]V) []struct {
    Key   K
    Value V
} {
    // Extract and sort keys
    keys := make([]K, 0, len(m))
    for k := range m {
        keys = append(keys, k)
    }
    sort.Slice(keys, func(i, j int) bool { return keys[i] < keys[j] })

    // Create sorted key-value pairs
    pairs := make([]struct {
        Key   K
        Value V
    }, len(keys))

    for i, k := range keys {
        pairs[i] = struct {
            Key   K
            Value V
        }{k, m[k]}
    }

    return pairs
}
```

MatheusFranco99:
@GalRogozinski added the generic function. Pls let me know if it's ok

GalRogozinski (approved):
awesome

alan-ssvlabs (approved):
LGTM

PR #530, opened 2025-03-17
feat: add support of hoodi testnet

MatusKysel:
based on https://github.com/eth-clients/hoodi/blob/main/metadata/config.yaml


types/beacon_types.go
```diff
@@ -204,6 +211,8 @@ func (n BeaconNetwork) MinGenesisTime() uint64 {
 		return 1606824023
 	case HoleskyNetwork:
 		return 1695902400
+	case HoodiNetwork:
+		return 1742212800 + 600
```
y0sher, line 215: why not use the final number?

y0sher (approved):

PR #548, opened 2025-06-05
Stop sending decided message

alan-ssvlabs:


types/testingutils/keymanager.go
```diff
@@ -43,10 +42,10 @@ var (
 )
 
 func NewTestingKeyManager() *TestingKeyManager {
-	return NewTestingKeyManagerWithSlashableSlots(map[string][]phase0.Slot{})
+	return NewTestingKeyManagerWithSlashableSlots(map[string][]spec.Slot{})
```
GalRogozinski, line 45: didn't we say we use phase0?

alan-ssvlabs: Yes, before we said so I decided to use spec when doing this PR. Updated now

PR #554, opened 2025-06-23
Fix round number in qbft tests

### LINKED ISSUE #551: Round number still set to 0 in some tests
Hey all, 

We've run into more problems with round validation. 

It seems this [PR](https://github.com/ssvlabs/ssv-spec/pull/537) hasn't addressed all the tests. 

Here are two examples:
https://github.com/ssvlabs/ssv-spec/blob/main/qbft/spectest/generate/tests/tests.CreateMsgSpecTest_qbft_create_message_create_round_change.json#L38

https://github.com/ssvlabs/ssv-spec/blob/main/qbft/spectest/generate/tests/tests.CreateMsgSpecTest_qbft_create_message_create_round_change_no_justification_quorum.json#L38

Is there a way we can check validation of types for most of the tests to try and catch all of these?

alan-ssvlabs:
closes #551 

In QBFT create message spec tests, some tests do not specify round number, as a result the round is set to 0 (qbft.noRound), which is not quite valid. This PR assigns round number 1 to these tests.

### DIFF
--- a/qbft/spectest/generate/state_comparison/tests_CreateMsgSpecTest/qbft create message create round change no justification quorum.json
+++ b/qbft/spectest/generate/state_comparison/tests_CreateMsgSpecTest/qbft create message create round change no justification quorum.json
@@ -35,7 +35,7 @@
 				0
 		],
 		"StateValue": "AQIDBAUGBwgJAQIDBAUGBwgJAQIDBAUGBwgJ",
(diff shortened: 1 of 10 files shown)


MatheusFranco99 (approved):
LGTM

AgeManning (approved):
LGTM. Thank you!

PR #559, opened 2025-07-01
Add private keys to tests

### LINKED ISSUE #542: Add private keys to tests
## Description

In many tests we have the `Signatures` field. It is best to print the private key that created the signature. We can do it inside the json itself or have some external file that maps public keys to private keys.

alan-ssvlabs:
closes #542


types/testingutils/comparable/helpers.go
```diff
@@ -114,6 +114,10 @@ func CompareWithJson(t *testing.T, test any, testName string, testType string) {
 	expectedTestMap, err := GetExpectedStateFromScFile(testName, testType)
 	require.NoError(t, err)
 
+	// Remove PrivateKeys field state comparison
+	delete(testMap, "PrivateKeys")
+	delete(expectedTestMap, "PrivateKeys")
+
```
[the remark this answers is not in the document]
alan-ssvlabs: This is used when running the SC tests; the code in main.go is used when generating the SC tests.
When generating the tests, the type is not a map so can't use the same code as here.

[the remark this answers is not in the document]
alan-ssvlabs: I was being stupid. There is a simpler way to exclude private keys when generating SC, from the `GetPostState()` function in each test type. So we don't need to change main.go indeed

qbft/spectest/tests/messages_spectest.go
```diff
@@ -59,5 +65,31 @@ func (test *MsgSpecTest) TestName() string {
...
 
 func (test *MsgSpecTest) GetPostState() (interface{}, error) {
+	// remove private keys
+	test.PrivateKeys = nil
+
 	return test, nil
 }
+
+// SetPrivateKeys populates the PrivateKeys field with keys from the given TestKeySet
+func (test *MsgSpecTest) SetPrivateKeys(ks *testingutils.TestKeySet) {
```
Zacholme7, line 75: Not too strong in go so not sure about how visibility works, but seems like this function is copied and pasted a few times. Can this just be written once and imported?

qbft/spectest/tests/messages_spectest.go
```diff
@@ -59,5 +65,31 @@ func (test *MsgSpecTest) TestName() string {
...
+	for operatorID, shareSK := range ks.Shares {
+		privateKeyInfo.Shares[operatorID] = hex.EncodeToString(shareSK.Serialize())
+	}
+
+	// Add operator keys (RSA private keys used for signing)
+	for operatorID, operatorKey := range ks.OperatorKeys {
+		// For RSA keys, we'll include the modulus and exponent
+		privateKeyInfo.OperatorKeys[operatorID] = fmt.Sprintf("N:%s,E:%d",
+			operatorKey.N.String(), operatorKey.E)
+	}
```
Zacholme7, line 92: What is the reason behind just including modulus and exponent? Im not sure if it is even possible to rebuild the key from these two parts, it is? Also makes parsing more tricky as we have to search for the cutoffs. 

Can we just write out the full hex encoded keys here?

Zacholme7 (commented):
2 small questions, this will help a lot though!

Zacholme7:
Follow up - Can we just write out the entire TestKeySet to the json file instead?

qbft/spectest/tests/messages_spectest.go
alan-ssvlabs: Thanks for this suggestion! Now the logic of this function is moved to a single file.

qbft/spectest/tests/messages_spectest.go
alan-ssvlabs: Getting full hex encoded keys is kind of difficult because private keys are not designed to be exported in this way and there isn't such a built-in function for this. However you are right that we need more than N and E, there is also D which is the most important part. I added this now.

Zacholme7:
Thx for including the variables. 

As for writing out the keys, can you just write out its PEM? Or it could go from DER into hex

Zacholme7 (approved):

dknopik (approved):

PR #571, opened 2025-08-02
Tests - Change empty arrays to null

### LINKED ISSUE #568: Optional fields are inconsistently represented
There is a lot of mixing and matching between the way in which null fields are represented. 

For example [here](https://github.com/ssvlabs/ssv-spec/blob/96f9be2ac2970f0ac07a3beb7dc458615d283a89/ssv/spectest/generate/tests/tests.MultiMsgProcessingSpecTest_nil_SSVMessage.json#L4532) `BeaconBroadcastedRoots` is `null` but [here](https://github.com/ssvlabs/ssv-spec/blob/96f9be2ac2970f0ac07a3beb7dc458615d283a89/ssv/spectest/generate/tests/tests.MultiMsgProcessingSpecTest_post_consensus_before_decided.json#L597) it is `[]` and this makes for some messy parsing. 

Is there a semantic difference here?

alan-ssvlabs:
closes #568


ssv/spectest/run_test.go
```diff
@@ -203,11 +203,13 @@ func newRunnerDutySpecTestFromMap(t *testing.T, m map[string]interface{}) *newdu
 	}
 
 	outputMsgs := make([]*types.PartialSignatureMessages, 0)
-	for _, msg := range m["OutputMessages"].([]interface{}) {
-		byts, _ := json.Marshal(msg)
-		typedMsg := &types.PartialSignatureMessages{}
-		require.NoError(t, json.Unmarshal(byts, typedMsg))
-		outputMsgs = append(outputMsgs, typedMsg)
+	if m["OutputMessages"] != nil {
```
GalRogozinski, line 206: won't we be getting errors because of this change?

GalRogozinski: it is ok

GalRogozinski (approved):

PR #572, opened 2025-08-04
Aggregator Committee

olegshmuelov:


GalRogozinski:
So far looks good @olegshmuelov

types/consensus_data.go
```diff
@@ -251,74 +245,180 @@ func (cd *ValidatorConsensusData) GetBlockData() (blk *api.VersionedProposal, si
...
+
+		default:
+			return nil, nil, WrapError(UnknownBlockVersionErrorCode, fmt.Errorf("unsupported version %s", a.Version.String()))
 		}
 
-		return &spec.VersionedAggregateAndProof{Version: cd.Version, Fulu: ret}, ret, nil
-	default:
-		return nil, nil, fmt.Errorf("unknown aggregate and proof version %d", cd.Version)
+		proofs = append(proofs, aggregateAndProof)
 	}
```
GalRogozinski: this is really confusing
`proofs` are a VersionedAggregateAndProofs
and `hashRoots` are the same in SSZ form?

this is really confusing. The names of the variables are bad. Also you can also extract the `hashroot` easily from the `proof`.

Maybe just return `aggregateAndProofs` list and that's it?

GalRogozinski (changes requested):

ssv/aggregator_committee.go
```diff
@@ -0,0 +1,1104 @@
...
+	}
+
+	// Early exit if no aggregators selected (conditioned to no error)
+	if !hasAnyAggregator && anyErr == nil {
+		r.BaseRunner.State.Finished = true
+		if anyErr != nil {
+			return anyErr
+		}
+		return nil
+	}
```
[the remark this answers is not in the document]
MatheusFranco99: Fixed thanks!

MatheusFranco99:
On the following TODO:
> // TODO(Aleg) why this sort? why not root sort?

about why sorting the validators to be processed once a root is selected.

No major reason to choose validator index specifically against sorting the root.
Pretty much, it ensures that every node processes the validator metadata entries in ascending validator-index order instead of the random order that map iteration would yield, thus controlling the order aggregators/contributors get appended to aggregatorData

types/runner_role.go
```diff
@@ -4,29 +4,21 @@ package types
 type RunnerRole int32
 
 const (
-	RoleCommittee RunnerRole = iota
-	RoleAggregator
```
nkryuchkov: I don't think we can remove it now and shift the values, because the implementation depends on spec and will still use Alan code before the fork. 

@GalRogozinski @MatheusFranco99 I'm going to revert it for now, because it broke the implementation. We can continue discussing this until we decide on a workaround for this

y0sher:
@GalRogozinski @nkryuchkov @MatheusFranco99 WDYT about shipping the Boole fork version with a spec dependency that has new and old code so we can do both in the impl. and removing the old from the spec in the next version?

ssv/committee.go
```diff
@@ -10,10 +10,10 @@ import (
 	"github.com/ssvlabs/ssv-spec/types"
 )
 
-type CreateRunnerFn func(shareMap map[phase0.ValidatorIndex]*types.Share) *CommitteeRunner
+type CreateRunnerFn func(shareMap map[phase0.ValidatorIndex]*types.Share) Runner
```
y0sher: since we have 2 types of runners now, shouldn't this function have the ability to differentiate between them, better pass something to identify the type of runner needed.

ssv/runner_validations.go
```diff
@@ -16,20 +16,25 @@ func (b *BaseRunner) ValidatePreConsensusMsg(runner Runner, psigMsgs *types.Part
 		return types.NewError(types.NoRunningDutyErrorCode, "no running duty")
 	}
 
-	if err := b.validatePartialSigMsgForSlot(psigMsgs, b.State.StartingDuty.DutySlot()); err != nil {
-		return err
-	}
+	switch runner.(type) {
+	case *AggregatorCommitteeRunner:
+		return b.validatePartialSigMsgForSlot(psigMsgs, b.State.StartingDuty.DutySlot())
+	default:
```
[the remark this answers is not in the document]
y0sher: also since there's only one case and default here (like it used to be in post-consensus), it might be nicer to do it with an if instead:
```go
if runner.(*AggregatorCommitteeRunner) {
   return ...
}

// rest of the code
```

ssv/aggregator_committee.go
```diff
@@ -0,0 +1,1105 @@
...
+			switch role {
+			case types.BNRoleAggregator:
+				aggregateAndProof := beaconObj.(*spec.VersionedAggregateAndProof)
+				signedAgg, err := r.constructSignedAggregateAndProof(aggregateAndProof, blsSig)
+				if err != nil {
+					anyErr = errors.Wrap(err, "failed to construct signed aggregate and proof")
+					continue
+				}
+
+				if err := r.beacon.SubmitSignedAggregateSelectionProof(signedAgg); err != nil {
```
y0sher: we should be able to send all of them at once like we do with attestations. the go-eth2-client api `SubmitAggregateAttestations` which SubmitSignedAggregateSelectionProof uses allows passing a slice of msgs.

GalRogozinski:
Ask @liorrutenberg about it

GalRogozinski:
@liorrutenberg decided to remove pre fork logic

ssv/aggregator_committee.go
```diff
@@ -637,8 +636,6 @@ func (r *AggregatorCommitteeRunner) processAggregatorSelectionProof(
 		return false, nil
 	}
 
-	// TODO: waitToSlotTwoThirds(vDuty.Slot)
```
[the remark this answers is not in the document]
MatheusFranco99: This has never been in spec, only in implementation.
Maybe we should have it in spec, but it would be outside the scope of this PR

types/runner_role.go
MatheusFranco99: Made it explicit (without iota). Now, RoleAggregatorCommittee maps to a new `6` index.
Thanks!

MatheusFranco99:
@nkryuchkov 
Made the values of `PartialSigMsgType` persistent through the fork. Thanks! 

cc @GalRogozinski  @shane-moore

types/runner_role.go
GalRogozinski: @liorrutenberg 
wdyt?

nkryuchkov: > I was wrong in thinking it would be quick to do. If I add the deleted runner roles and partial signature types, I cannot override methods on them like `String()`, so I have to create a new type, which breaks a lot of places with interaction with spec. Overall, I think it's doable, but it will require copying a lot of things from spec.

I've found a workaround for this without copying a lot of things, so this is not an issue now. But still, maybe keeping those consts would be good for clarity. Although, I don't mind deleting them

shane-moore (approved):
have spent significant time thinking through and implementing the spec so far without issues except what we've already discussed.  I approve the spec, and if I find anything else later that would be an enhancement or potential fix, I'll raise an issue to discuss

ssv/aggregator_committee.go
```diff
@@ -0,0 +1,1105 @@
...
+				r.RecordSubmission(types.BNRoleAggregator, validator, root)
+
+			case types.BNRoleSyncCommitteeContribution:
+				contribAndProof := beaconObj.(*altair.ContributionAndProof)
+				signedContrib := &altair.SignedContributionAndProof{
+					Message:   contribAndProof,
+					Signature: blsSig,
+				}
+
+				if err := r.beacon.SubmitSignedContributionAndProof(signedContrib); err != nil {
```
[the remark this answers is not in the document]
MatheusFranco99: Pls, resolve if fine :)

ssv/aggregator_committee.go
MatheusFranco99: Pls, resolve if fine :)

ssv/runner_validations.go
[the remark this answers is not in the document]
MatheusFranco99: Pls, resolve if fine :)

ssv/committee.go
MatheusFranco99: Pls, resolve if fine :)

types/runner_role.go
MatheusFranco99: Nice. Pls, resolve this comment if it's fine now :)

PR #576, opened 2025-08-20
spectests: align with robust message replay

iurii-ssv:
This PR intends to update/align the code in `ssv-spec` repo with the changes in https://github.com/ssvlabs/ssv/pull/2445 so that we can have **spec-tests** passing in `ssv` repo


qbft/spectest/tests/controller/decided/no_quorum.go
```diff
@@ -30,8 +30,7 @@ func NoQuorum() tests.SpecTest {
 			},
 		},
 		nil,
-		// TODO: before merge ask engineering how often they see such message in production
-		"could not process msg: invalid signed message: did not receive proposal for this round",
+		"no proposal for round: invalid signed message: did not receive proposal for this round",
 		nil,
 		ks,
 	)
```
iurii-ssv, line 36: We do see quite a bit of these, here is an example from our production cluster:
image

So, with https://github.com/ssvlabs/ssv/pull/2445 we should be able to buffer messages in this scenario and "replay" it later (vs just dropping the message as if it never came, as is done currently)

ssv/spectest/tests/runner/postconsensus/invalid_msg_slot.go
```diff
@@ -116,32 +115,46 @@ func InvalidMessageSlot() tests.SpecTest {
 				},
 				ExpectedError: "no post consensus phase for voluntary exit",
 			},
+			{
+				Name: fmt.Sprintf("aggregator (%s)", spec.DataVersionPhase0.String()),
```
GalRogozinski, line 119: @iurii-ssv 
can you explain the reason for this change?

iurii-ssv: Yes, so in https://github.com/ssvlabs/ssv/pull/2445 I had to introduce a new error to split `invalid partial sig slot` in 2:
- `invalid partial sig slot`
- `ErrFuturePartialSigMsg`(this is a newly-defined error current spec-tests don't know about and this change fixes that)

the reason to split it like that is because `ErrFuturePartialSigMsg` is a retryable error (we can buffer the message and replay it ~100ms later successfully)

PR #585, opened 2025-10-08
Error codes for spec-tests

iurii-ssv:
Resolves https://github.com/ssvlabs/ssv-spec/issues/514 to implement error-code-based checks in tests (to replace error-message-based ones),

the general approach I took is:
- go over all the tests that assert on the expected error message
- "turn that error-message into an error-code" and assert on the error-code instead (the implementation returns an error that has both code and message, in spec-tests only the error-code is asserted on to allow for updating error-messages without tripping spec-tests - this is mostly useful for upgrading consensus code in https://github.com/ssvlabs/ssv without having to update the mirroring error-messages in this repo)
- there are some errors that are not checked for by any of the tests, and hence that means those errors are left unchanged (I did not create a corresponding error-code for those) ... I think it's OK for now (if somebody adds a test that needs such an error to have an error-code it would be easy to add following the approach I'm using in this PR)
- all error-codes are listed in `types/error.go` file
- besides the changes to spec-tests (and their corresponding generated `.json` files), the only other notable change this PR introduces is to split the `invalid partial sig slot` error (in `validatePartialSigMsgForSlot` func) into 2 distinct errors `invalid partial sig slot` and `future partial sig msg` to be able to differentiate between messages we want to retry vs messages we want to drop (as is implemented in https://github.com/ssvlabs/ssv/pull/2445)


types/error_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestError(t *testing.T) {
```
GalRogozinski: this file can be removed imo

iurii-ssv: It just ensures that we properly implement the methods that allow to wrap/unwrap our `Error` type (it might seem trivial but there are lots of gotchas in making it work correctly actually), 

any particular reason to remove it ?

PR #586, opened 2025-10-23
Re-add missing tests and fix error code handling

GalRogozinski:
## Description

Some tests were commented out mistakingly before.
I re-added them and fixed an issue with the new error codes that made them fail.


ssv/spectest/tests/partialsigcontainer/test.go
```diff
@@ -48,10 +48,6 @@ func (test *PartialSigContainerTest) Run(t *testing.T) {
 		result, err := ps.ReconstructSignature(root, test.ValidatorPubKey, validatorIndex)
 		// Check the result and error
 		tests.AssertErrorCode(t, test.ExpectedErrorCode, err)
-		if err != nil {
-			require.Contains(t, err.Error(), test.ExpectedErrorCode)
-			return
-		}
```
iurii-ssv: Note: I did not remove this `if err != nil` check here because it changes the behavior - namely, the `require.EqualValues(t, test.ExpectedResult, result)` wouldn't execute if we got an error

now `require.EqualValues(t, test.ExpectedResult, result)` would execute - I'm not 100% sure what is the intended behavior, but I've tried to preserve the behavior as it was in my [error-codes PR](https://github.com/ssvlabs/ssv-spec/pull/585)

GalRogozinski: So now I changed behavior so result is simply `nil`
Test has changed as a result

PR #589, opened 2025-11-26
Relax Majority Protection Rules

GalRogozinski:
## Description

We should relax majority protection rules to only checks the checkpoints epochs of a `BeaconVote` (attestation) against the node's own local view of the chain.

This will ensure that the DV will remain live in case of a fork. It will only fail the value check in the case the resulting `SignedAttestation` will make choosing the other fork a slashable offense.

### DIFF
--- a/ssv/value_check.go
+++ b/ssv/value_check.go
@@ -57,7 +57,7 @@ func BeaconVoteValueCheckF(
 		if bv.Source.Epoch != expectedSource {
 			return types.NewError(types.CheckpointMismatch,
 				fmt.Sprintf("attestation data source checkpoint %d does not match expected %d",
-					bv.Source, expectedSource))
+					bv.Source.Epoch, expectedSource))
 		}
 
 		if bv.Target.Epoch != expectedTarget {


diegomrsantos:
I think it would help if the description explicitly stated the scope of this change (and the corresponding SIP).

With this PR, “majority fork protection” is effectively slashing protection only in majority-fork scenarios:
	•	BeaconVoteValueCheckF now checks only source_epoch / target_epoch against the local BN view; roots are no longer part of the value check.
	•	The full AttestationData (incl. roots) is used only by the per-share slashing logic, so we still avoid creating slashable attestations, but we no longer differentiate forks beyond “slashable vs non-slashable”.  ￼

Concretely, in a Holesky-style fork:
	•	Some operators can be on a minority / canonical fork with different source/target roots but the same epochs as the majority fork.
	•	Under the new rule, they will still accept and sign an attestation built on the majority fork as long as the epochs match and it’s not slashable for their share.
	•	This means the DV can continue to attest on a non-canonical majority fork, even when some operators’ CLs are following the canonical minority fork.

I’d suggest adding a short “Scope and limitations” section making it explicit that:
	•	the mechanism guarantees slashing safety for the DV (no slashable attestations), but
	•	it does not guarantee staying on the canonical/minority fork in majority-client-bug events.

diegomrsantos:
@GalRogozinski,

I believe "Majority-Fork Slashing Protection" would be a clearer name for the SIP.

PR #591, opened 2025-12-13
Round Robin Proposer with Epochs

MatheusFranco99:
## Overview

This PR adds epochs into the Round Robin proposer computation in order to generate variance through epochs.

More precisely, the previous round robin function rotated only through slots and rounds.
Because there are 32 slots in an epoch and 32%4=0, for committees with 4 operators every epoch had the same profile of proposers.


qbft/round_robin_proposer_test.go
```diff
@@ -1 +0,0 @@
-package qbft_test
```
GalRogozinski: how did this exist? 😛

GalRogozinski (approved):
We just need to make a [SIP](https://github.com/ssvlabs/SIPs/issues/65#issuecomment-3691220019)

PR #594, opened 2025-12-21
Agg comm improvements

MatheusFranco99:
## Overview

This PR adds improvements to the Aggregator Commmittee Runner:
- duty validation
- partial signature message sorting
- improve consensus data to remove duplicated redundant data (for the same subnets)
- add explicit logic about pre-consensus finalization (namely, whenever at least a quorum for one validator is reached, proceed to consensus)


ssv/spectest/tests/aggregatorcommittee/multipleduty/duty_with_different_slots.go
```diff
@@ -0,0 +1,53 @@
...
+					BeaconBroadcastedRoots: []string{},
+					ExpectedErrorCode:      types.InvalidAggregatorCommitteeDutyErrorCode,
+				},
+			}...)
+		}
+	}
+
+	multiSpecTest := committee.NewMultiCommitteeSpecTest(
+		"aggregator committee runner duty with different slots",
+		"Testing aggregator committee runner with duties that have different slots",
```
GalRogozinski: I don't get this test..
docs should explain that different number of validators are running.
Not sure why we need so many iterations

ssv/spectest/tests/committee/test.go
```diff
@@ -35,6 +35,9 @@ type CommitteeSpecTest struct {
 	BeaconBroadcastedRoots []string
 	ExpectedErrorCode      int
 	PrivateKeys            *testingutils.PrivateKeyInfo `json:"PrivateKeys,omitempty"`
+	// StrictMessageOrder, when true, compares output messages with strict ordering
+	// (both message order and partial signature order within each message must match exactly)
+	StrictMessageOrder bool `json:"StrictMessageOrder,omitempty"`
```
GalRogozinski: can you wirte a word why this has become neccessary now

types/consensus_data.go
```diff
@@ -265,41 +265,79 @@ type AggregatorCommitteeConsensusData struct {
...
 
 	// Aggregator duties
-	Aggregators                 []AssignedAggregator `ssz-max:"3000"`
-	AggregatorsCommitteeIndexes []uint64             `ssz-max:"3000"`
-	Attestations                [][]byte             `ssz-max:"3000,1048576"`
+	Aggregators []AssignedAggregator `ssz-max:"3000"` // For a maximum of 3k validators per committee
+	// AggregatorsCommitteeIndexes is a list of committee indexes used by the above aggregators
+	AggregatorsCommitteeIndexes []uint64 `ssz-max:"64"`
+	// AggregatedAttestations is a list of aggregated attestations (SSZ bytes), one for each committee above
+	AggregatedAttestations [][]byte `ssz-max:"64,1048576"`
```
GalRogozinski: 64 is better now
still not sure about 3000 and 1048576

types/consensus_data.go
```diff
@@ -265,41 +265,79 @@ type AggregatorCommitteeConsensusData struct {
...
+		if _, ok := allowedAggCommittees[agg.CommitteeIndex]; !ok {
+			return NewError(AggCommCommIdxMismatchErrorCode, "aggregator committee index not listed in AggregatorsCommitteeIndexes")
+		}
+		// Mark as used
+		usedAggCommittees[agg.CommitteeIndex] = struct{}{}
+	}
+	// Ensure no committee index was left unused (no more than necessary)
+	if len(usedAggCommittees) != len(allowedAggCommittees) {
+		return NewError(AggCommUnusedCommIdxErrorCode, "leftover aggregator committee index not usedAggCommittees by any aggregator")
+	}
```
GalRogozinski: are there tests for the new errors?
will help me understand how they are all triggered

ssv/spectest/tests/aggregatorcommittee/multipleduty/duty_with_different_slots.go
MatheusFranco99: Sorry, in this case, a `AggregatorCommitteeDuty` has several `ValidatorDuty`s with different slots, which is invalid (invalid duty) and that's why the test checks that the error is raised.
Makes sense?

types/error.go
```diff
@@ -79,12 +79,18 @@ const (
 	TimeoutInstanceErrorCode
 	AggCommAggAttCntMismatchErrorCode
```
GalRogozinski: not used

ssv/aggregator_committee.go
```diff
@@ -677,13 +704,20 @@ func (r *AggregatorCommitteeRunner) processSyncCommitteeSelectionProof(
...
+		if contrib.SubcommitteeIndex == subnetID {
+			// If so, just add to contributors and return
+			aggregatorData.Contributors = append(aggregatorData.Contributors, types.AssignedAggregator{
+				ValidatorIndex: vDuty.ValidatorIndex,
+				SelectionProof: selectionProof,
+				CommitteeIndex: syncCommitteeIndex,
+			})
 			return true, nil
 		}
 	}
```
GalRogozinski: add a spec test that hits this path please!

GalRogozinski (changes requested):
Add a test that checks that a preconsensus quorum was completed but not aggregator was found, but then later comes another message that completes the missing validator that may be an aggregator

types/consensus_data.go
```diff
@@ -265,41 +265,79 @@ type AggregatorCommitteeConsensusData struct {
...
 
 	// Aggregator duties
-	Aggregators                 []AssignedAggregator `ssz-max:"3000"`
-	AggregatorsCommitteeIndexes []uint64             `ssz-max:"3000"`
-	Attestations                [][]byte             `ssz-max:"3000,1048576"`
+	Aggregators []AssignedAggregator `ssz-max:"3000"` // For a maximum of 3k validators per committee
+	// AggregatorsCommitteeIndexes is a list of committee indexes used by the above aggregators
+	AggregatorsCommitteeIndexes []uint64 `ssz-max:"64"`
+	// AggregatedAttestations is a list of aggregated attestations (SSZ bytes), one for each committee above
+	AggregatedAttestations [][]byte `ssz-max:"64,131308"`
```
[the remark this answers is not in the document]
MatheusFranco99: You are correct in your estimation, but that's for a phase0 Attestation, which has a maximum size of 2276.
For electra attestations, the aggbits alone may have up to 131072 bytes.

ssv/aggregator_committee.go
```diff
@@ -709,9 +743,9 @@ func (r *AggregatorCommitteeRunner) processSyncCommitteeSelectionProof(
 		aggregatorData.Contributors = append(aggregatorData.Contributors, types.AssignedAggregator{
 			ValidatorIndex: vDuty.ValidatorIndex,
 			SelectionProof: selectionProof,
+			CommitteeIndex: subnetID,
```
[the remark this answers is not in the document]
MatheusFranco99: Took a deeper look.
The correct was indeed to use `SubnetID`.
The other is the validator index in the sync committee.
Actually we got no errors before because our testing Beacon node implements:
```go
func (bn *TestingBeaconNode) SyncCommitteeSubnetID(index phase0.CommitteeIndex) uint64 {
	// Real calculation:
	// Each subnet has syncCommitteeSize / subnetCount validators
	// subnetCount is 4 for mainnet
	// const subnetCount = 4
	// const syncCommitteeSize = 512
	// const subnetSize = syncCommitteeSize / subnetCount
	// return index / subnetSize

	// For testing, we use a simplified approach where subnet ID equals the index
	// This means validator at index 0 is in subnet 0, index 1 in subnet 1, etc.
	// In production, multiple validators share the same subnet (128 validators per subnet)
	return uint64(index)
}
```
I'm thinking if this should be tested and if so, will add a test

ssv/aggregator_committee.go
MatheusFranco99: Added

MatheusFranco99:
@GalRogozinski added the tests. Let me know if you agree with it

GalRogozinski (commented):
LGTM
just quickly answer unresolved comments

PR #596, opened 2025-12-26
Rename ValidatorConsensusData to ProposerConsensusData

MatheusFranco99:
## Overview

The `ValidatorConsensusData` type is currently only used for the block proposer duty.
Thus, it's best to rename it and align to its usage.
This PR renames it to `ProposerConsensusData`, changes its validation, and align tests.


types/error.go
```diff
@@ -14,8 +14,6 @@ const (
 	WrongMessageHeightErrorCode
 	SignerIsNotInCommitteeErrorCode
 	NonUniqueSignerErrorCode
-	ValidatorRegistrationNoConsensusDataErrorCode
-	ValidatorExitNoConsensusDataErrorCode
```
GalRogozinski: I think I asked before but why was this changed?

GalRogozinski: ok you changed Validate

PR #597, opened 2026-01-02
Aggregator Committee - Fix committee runners management

MatheusFranco99:
## Overview

This PR fixes the `Committee` class to correctly manage `CommitteeRunner`s and `AggregatorCommitteeRunner`s.

### DIFF
--- a/ssv/spectest/tests/committee/wrong_role.go
+++ b/ssv/spectest/tests/committee/wrong_role.go
@@ -93,10 +93,10 @@ func WrongRole() tests.SpecTest {
 			Input: []interface{}{
 				aggDuty,
 
-				invalidateRunnerRole(testingutils.SignPartialSigSSVMessage(ks, testingutils.SSVMsgAggregatorCommittee(ks, nil, testingutils.PreConsensusAggregatorCommitteeMsgForDuty(aggDuty, ksMap, 1, version)))),
+				invalidateRunnerRole(testingutils.SignPartialSigSSVMessage(ks, testingutils.SSVMsgAggregatorCommittee(ks, nil, testingutils.PreConsensusAggregatorCommitteeMsgForDuty(aggDuty, ksMap, 1)))),
 			},
 			OutputMessages: []*types.PartialSignatureMessages{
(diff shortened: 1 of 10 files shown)


ssv/json_testutils.go
```diff
@@ -162,39 +165,24 @@ func (c *Committee) UnmarshalJSON(data []byte) error {
...
-	for slot, runnerData := range raw.Runners {
-		// Try to detect the runner type by looking for type-specific fields
-		var typeDetector struct {
-			BaseRunner struct {
-				RunnerRoleType types.RunnerRole `json:"RunnerRoleType"`
-			} `json:"BaseRunner"`
-		}
-
-		if err := json.Unmarshal(runnerData, &typeDetector); err != nil {
+	for slot, runnerData := range raw.CommitteeRunners {
```
GalRogozinski: so all the previous code was unneeded?

ssv/spectest/tests/committee/wrong_role.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/ssvlabs/ssv-spec/ssv/spectest/tests"
+	"github.com/ssvlabs/ssv-spec/types"
+	"github.com/ssvlabs/ssv-spec/types/testingutils"
+)
+
+// WrongRole ensures that a message with wrong runner role (nor CommitteeRuner or AggregatorCommitteeRunner) in the MessageID field raises an error.
+func WrongRole() tests.SpecTest {
+
+	tests := []*CommitteeSpecTest{}
+	expectedError := types.CommitteeWrongRoleErrorCode
```
GalRogozinski: We also have `UnknownDutyRoleDataErrorCode` today
Do we need another error?
Should there be a difference in tests between committee and non-committee roles?

MatheusFranco99: The committee allows concurrent duties in different slots (e.g. `S+1` can start even though `S` hasn't finished yet).
The validator class runs sequentially.
Is the error you mentioned reachable here?

PR #599, opened 2026-01-13
Round-Robin: Ensure sorting

MatheusFranco99:
## Overview

This PR makes the round-robin function ensure that the operators list is sorted before using it.


julienh-ssv (dismissed):

GalRogozinski (dismissed):
Looks good, but if we have time we can see if in-place sorting is fine.

dknopik (approved):

PR #608, opened 2026-03-20
always blind blocks

nkryuchkov:
**Summary**

This PR updates `ssv-spec` proposer flow so QBFT always agrees on the blinded block form.

**What Changed**

- Updated proposer pre-consensus flow to blind fetched beacon blocks before encoding `ValidatorConsensusData`
- Regenerated affected proposer spectest JSON and state-comparison fixtures

**Why**

The spec should define the consensus algorithm, and the important invariant here is that proposer QBFT values are blinded. This reduces the mismatch between `ssv-spec` and `ssv` while keeping the spec focused on protocol behavior rather than node-specific mechanics.


y0sher (approved):
lgtm code wise,but @GalRogozinski should probably approve if he wants this change in the spec.

olegshmuelov (approved):

GalRogozinski (changes requested):
denied until discussion concludes:
https://github.com/ssvlabs/SIPs/discussions/78#discussioncomment-16054804

PR #609, opened 2026-03-20
MessageID - separate constructor for Committee and Validator messages

oleg-ssvlabs:
## Summary
  Split `MessageID` construction into dedicated validator and committee helpers so the executor identifier type is
  explicit at each call site.

  ## Why
  `MessageID` stores the duty executor in a 48-byte slot, but we were previously building it through a single
  `NewMsgID(..., []byte, ...)` helper for both:
  - validator pubkeys (48 bytes)
  - committee IDs (32 bytes, right-aligned in the 48-byte slot)

  That made the encoding rule implicit and easy to misuse.

  ## What changed
  - added `NewValidatorMsgID(...)` for validator-based runners
  - added `NewCommitteeMsgID(...)` for committee-based runners, with explicit right-alignment of the 32-byte committee
  ID in the executor slot
  - updated production runner call sites to use the correct constructor
  - updated testing helpers and spec tests to use the new constructors
  - regenerated `types/messages_encoding.go`
  
  Closes https://github.com/ssvlabs/ssv-node-board/issues/864


olegshmuelov (dismissed):
LGTM.

nit: The redundant `types.RunnerRole(...)` cast in `validator_registration.go:154` could be removed since `RunnerRoleType` is already `RunnerRole`. Pre-existing, not introduced by this PR.

iurii-ssv (dismissed):
LGTM, perhaps worth checking out those Greptile comments above.

olegshmuelov (dismissed):

ljuba-ssv (dismissed):

ljuba-ssv (approved):

iurii-ssv (approved):
LGTM, still probably want spec-team to have a look

MatheusFranco99 (approved):
LGTM

PR #509, opened 2024-10-08
Commit small test json files for each test instead of the large ones

alan-ssvlabs:
Fixes #482 

This PR creates one json file for each test. Large tests.json files are still generated but no longer committed.

The tests still use large tests.json files to run, the small json files are to keep track of changes easily.

PR #610, opened 2026-03-23
ci: add spec-tests sync workflows

vaclav-ssvlabs:
## Summary

Adds GitHub Actions workflows to automatically sync generated spec test JSON files to the dedicated `spec-tests` repository.

- **`generate-spec-tests` composite action**: sets up Go, runs generators for `ssv`/`qbft`/`types` spectests, copies output to `../spec-tests`
- **`sync-spec-tests-pr`**: on PR open/synchronize, pushes generated files to a matching branch in spec-tests repo and creates/updates a mirrored PR
- **`sync-spec-tests-merge`**: on push to `main`, regenerates files and merges the corresponding spec-tests PR
- **`test.yaml`**: bumped to `ubuntu-24.04`, node24-compatible action versions

## Motivation

1. The specs are a dependency for ssv-node implementation. The jsons blow up the size of the dependency to an unacceptable level. Compilation errors may occur.
2. Small changes in spec make huge changes in jsons. This is problematic for modern tools and humans.


GalRogozinski:
Hey @vaclav-ssvlabs after the known issues, there may be more.

 I recommend the following:
1. On a `ci-stage` branch we should merge this and #602. This way we have the complete feature and we don't need the 3 reviews.
2. For tests, We need to tweak the environment so that the automation will work w.o main being merged. After we are confident we can collect the 3 reviews and merge to main.

wdyt?

I also opened #619 to fix the no changes edge case. Codex claims the greptile bug is a non-issue, but we should verify.
After you review my PR DM me and let's see if we can properly test this.

momosh-ssv (commented):
Thanks for putting this together — moving the JSONs to a dedicated repo is a big quality-of-life win. Spent some time walking through the workflows and the path-resolution helpers to make sure I understood the trust and failure-mode story. A few questions and suggestions below — nothing blocking, mostly things we'd love to talk through.

types/testingutils/comparable/helpers.go
```diff
@@ -101,6 +107,30 @@ func GetSCDir(basedir string, testType string) string {
...
+		return "", err
+	}
+	return filepath.Join(root, "..", "spec-tests", module), nil
+}
+
+func moduleRootFrom(start string) (string, string, error) {
+	dir := filepath.Clean(start)
+	for {
+		base := filepath.Base(dir)
+		if base == "qbft" || base == "ssv" || base == "types" {
```
momosh-ssv: Could this lead to a wrong-resolve if the repo is cloned under a directory literally named `types`, `ssv`, or `qbft`? The walk returns on the first ancestor whose basename matches, so e.g. `~/types/ssv-spec` would resolve to `~/spec-tests/types` silently — wrong module, wrong root. What do you think about anchoring on the `go.mod` of `github.com/ssvlabs/ssv-spec` instead? Walking up to find the module file and deriving the suffix from the cwd's relative path to it removes the basename guessing entirely.

.github/workflows/sync-spec-tests-pr.yaml
```diff
@@ -0,0 +1,128 @@
...
+        uses: ./.github/actions/generate-spec-tests
+
+      - name: Extract spec-tests repo name
+        id: repo-name
+        env:
+          SPEC_TESTS_REPO: ${{ vars.SPEC_TESTS_REPO }}
+        run: echo "name=${SPEC_TESTS_REPO##*/}" >> "$GITHUB_OUTPUT"
+
+      - name: Get GitHub App token
+        uses: actions/create-github-app-token@v3.0.0
```
momosh-ssv: Trying to understand the trust boundary here. `go generate` (from the composite action above) runs PR-controlled code in the same runner where the App token is exported a couple of steps later. Forks don't see the secret, but same-repo PR branches do, so a collaborator-pushed branch with a malicious generator could background a process that scrapes `${{ steps.app-token.outputs.token }}` once it's in the env. What do you think about splitting this into two jobs — one untrusted job that just generates artifacts and uploads them, and one trusted job that downloads and pushes with the App token? That way the App token never coexists with code from the PR.

.github/workflows/sync-spec-tests-pr.yaml
```diff
@@ -0,0 +1,128 @@
...
+
+          if [ -z "${SPEC_TESTS_REPO}" ]; then
+            echo "SPEC_TESTS_REPO variable is not set (e.g. ssvlabs/spec-tests)" >&2
+            exit 1
+          fi
+
+          BRANCH_NAME="${HEAD_REF}"
+
+          # Refuse to push to protected branch names in spec-tests
+          if [ "${BRANCH_NAME}" = "main" ] || [ "${BRANCH_NAME}" = "master" ]; then
```
momosh-ssv, line 134: Worth considering a stricter ref-shape check here? The literal `main`/`master` guard catches the obvious case, but `HEAD_REF` still flows into `git checkout -B`, `git push --force`, and `gh pr create --head`. A branch name starting with `-`, containing `..`, or named `HEAD` could be set on a fork PR. Something like `[[ "${BRANCH_NAME}" =~ ^[A-Za-z0-9._/-]+$ ]] && [[ "${BRANCH_NAME}" != -* ]] && [[ "${BRANCH_NAME}" != *..* ]] && [[ "${BRANCH_NAME}" != HEAD ]]` early in the step would close the gap.

qbft/spectest/generate/main.go
```diff
@@ -18,8 +18,9 @@ import (
 
 //go:generate go run main.go
 
-var testsDir = "tests"
-var stateComparisonDir = "state_comparison"
+var specTestsDir = filepath.Join("..", "..", "..", "..", "spec-tests", "qbft")
```
momosh-ssv: What happens if someone runs the generator from a different cwd than `qbft/spectest/generate`? The four `..` jumps are resolved against `os.Getwd()` at runtime, so `go run ./qbft/spectest/generate/main.go` from the repo root would land three levels above the repo's parent — and `clearStateComparisonFolder` then `os.RemoveAll`s that path. Same shape in `ssv/spectest/generate/main.go` and `types/spectest/generate/main.go`. Could we anchor the path off a `go.mod` walk (the same one `SpecTestsDirFrom` could use), or at least assert the resolved abs path matches `*/spec-tests/<module>` before deleting?

.github/workflows/sync-spec-tests-merge.yaml
```diff
@@ -0,0 +1,167 @@
...
+            --jq '.[0] | {number: (.number // empty), head_ref: (.head.ref // empty), base_ref: (.base.ref // empty)}' \
+            2>/dev/null || echo "{}")
+          PR_NUMBER=$(echo "${PR_INFO}" | jq -r '.number // empty')
+          PR_HEAD_REF=$(echo "${PR_INFO}" | jq -r '.head_ref // empty')
+          PR_BASE_REF=$(echo "${PR_INFO}" | jq -r '.base_ref // empty')
+
+          if [ -z "${PR_NUMBER}" ] || [ -z "${PR_HEAD_REF}" ]; then
+            echo "No PR found for commit ${GITHUB_SHA} in ${{ github.repository }}." >&2
+            echo "Every push to main must come from a merged PR. If this is a GitHub API timing issue, re-run the workflow." >&2
+            exit 1
```
momosh-ssv, line 79: What do you think about not failing the whole sync when a push to `main` has no associated PR? Right now an admin direct-push (or any push where the GitHub commit→PR association lags or breaks) exits 1 and blocks every subsequent push from syncing until a normal PR-merge happens. A `warn + exit 0` would be more forgiving. Also worth considering filtering the `jq` to PRs with `merged_at != null` — `.[0]` just picks the first associated PR, which on a merge commit could include unrelated ones.

.github/workflows/sync-spec-tests-merge.yaml
```diff
@@ -0,0 +1,167 @@
...
+          fi
+
+          git clone "https://x-access-token:${GH_TOKEN}@github.com/${SPEC_TESTS_REPO}.git" "${CLONE_DIR}"
+          cd "${CLONE_DIR}"
+          git config user.name "${BOT_NAME}"
+          git config user.email "${BOT_EMAIL}"
+          git fetch origin
+
+          RESOLVED_BASE="${PR_BASE_REF:-main}"
```
momosh-ssv, line 150: Could this drift if the spec-tests PR was opened against a different base than the ssv-spec PR's current base ref? `PR_BASE_REF` is whatever the ssv-spec PR's base is at merge time; the spec-tests PR was created by the PR-sync workflow against whatever the base was at PR-open time, and the two can diverge. Might be worth looking up the spec-tests PR's actual base via `gh pr view --json baseRefName` and using that here, so the merge workflow lines up with what the PR actually targets.

.github/workflows/sync-spec-tests-pr.yaml
```diff
@@ -0,0 +1,128 @@
...
+            echo "No changes to sync"
+            exit 0
+          fi
+
+          git commit -m "sync: ssv-spec PR #${PR_NUMBER}"
+          # Force push to overwrite previous syncs for this PR
+          git push --force origin "${BRANCH_NAME}"
+
+          # Create PR if missing, otherwise update title/body
+          if gh pr view --repo "${SPEC_TESTS_REPO}" "${BRANCH_NAME}" >/dev/null 2>&1; then
```
momosh-ssv: Any reason we're not scoping `gh pr view` to open PRs only? If a previous PR from this same branch was closed without merge, `gh pr view <branch>` can return the closed one — in which case the `gh pr edit` branch edits a closed PR (and the `gh pr create` else branch would have failed with 'a PR already exists for this branch' anyway). Something like `gh pr list --repo "${SPEC_TESTS_REPO}" --head "${BRANCH_NAME}" --state open --json number --jq '.[0].number'` would make the open-vs-not check explicit. Same shape in `sync-spec-tests-merge.yaml` at line 156.

.github/workflows/sync-spec-tests-pr.yaml
```diff
@@ -0,0 +1,128 @@
...
+          app_slug="${{ steps.app-token.outputs.app-slug }}"
+          user_id=$(gh api "/users/${app_slug}[bot]" --jq '.id')
+          echo "name=${app_slug}[bot]" >> "$GITHUB_OUTPUT"
+          echo "email=${user_id}+${app_slug}[bot]@users.noreply.github.com" >> "$GITHUB_OUTPUT"
+
+      - name: Sync to spec-tests PR branch
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          PR_TITLE: ${{ github.event.pull_request.title }}
```
momosh-ssv, line 106: Worth a one-line comment pinning the invariant that `PR_TITLE` must stay in `env:` and never be inlined as `${{ github.event.pull_request.title }}` directly in the shell script? It's safe today because the env passthrough quotes it, but a future refactor that 'simplifies' it back to inline `${{ }}` interpolation would reopen a script-injection vector via PR titles.

.github/workflows/sync-spec-tests-pr.yaml
```diff
@@ -0,0 +1,128 @@
...
+          GENERATED_DIR="${GITHUB_WORKSPACE}/../spec-tests"
+          CLONE_DIR=$(mktemp -d)
+
+          # Validate generated output exists
+          if [ -z "$(ls -A "${GENERATED_DIR}" 2>/dev/null)" ]; then
+            echo "Generated spec-tests directory is empty, aborting sync" >&2
+            exit 1
+          fi
+
+          git clone "https://x-access-token:${GH_TOKEN}@github.com/${SPEC_TESTS_REPO}.git" "${CLONE_DIR}"
```
momosh-ssv, line 148: Trying to understand the App token's lifetime here. It lands in `.git/config` of the clone via the URL. The clone is in `$(mktemp -d)` so it dies with the runner — fine today. As defense-in-depth, would it be worth either stripping the token from `.git/config` after push (`git remote set-url origin https://github.com/...` once the push is done) or asserting that `${CLONE_DIR}` is never uploaded as an artifact in a future step? Same shape at line 104 in `sync-spec-tests-merge.yaml`.

.github/workflows/sync-spec-tests-pr.yaml
```diff
@@ -0,0 +1,128 @@
...
+          git add -A
+
+          if git diff --cached --quiet; then
+            echo "No changes to sync"
+            exit 0
+          fi
+
+          git commit -m "sync: ssv-spec PR #${PR_NUMBER}"
+          # Force push to overwrite previous syncs for this PR
+          git push --force origin "${BRANCH_NAME}"
```
momosh-ssv, line 176: Might be worth namespacing the spec-tests branch with the PR number (e.g. `pr-${PR_NUMBER}/${HEAD_REF}`) to avoid cross-PR force-push collisions? If two ssv-spec PRs ever share a head branch name — e.g. someone re-uses a branch after a previous merge/close — they'd target the same spec-tests branch and clobber each other's generated content. The namespacing would make that impossible by construction.

.github/workflows/sync-spec-tests-merge.yaml
```diff
@@ -0,0 +1,167 @@
...
+          else
+            git commit -m "sync: final update from ssv-spec ${GITHUB_SHA}"
+            git push --force origin "${PR_HEAD_REF}"
+          fi
+
+          # Merge the spec-tests PR now that the ssv-spec PR has been merged.
+          # Look up the PR number by branch name at merge time to avoid stale output data.
+          SPEC_PR_NUMBER=$(gh pr view --repo "${SPEC_TESTS_REPO}" "${PR_HEAD_REF}" --json number --jq '.number' 2>/dev/null || echo "")
+          if [ -z "${SPEC_PR_NUMBER}" ]; then
+            if git diff --quiet "origin/${RESOLVED_BASE}...HEAD" -- "${GENERATED_PATHS[@]}"; then
```
momosh-ssv, line 201: Do we need to guard against `GENERATED_PATHS` being empty here? If the array somehow ends up empty (e.g. the earlier 'directory empty' check is bypassed by a future refactor), `git diff -- ` with no path args diffs everything, which changes the meaning of this check. A `${#GENERATED_PATHS[@]} -gt 0` assertion right before this line would be cheap insurance.

.github/workflows/sync-spec-tests-merge.yaml
vaclav-ssvlabs: There should be no admin direct push to main; the whole GitHub org settings is done this way. We should never do a commits to main. even Admin can do simple PR a merge the PR

and `blocks every subsequent push from syncing` is wrong the workflow runs on each commit

.github/workflows/sync-spec-tests-merge.yaml
vaclav-ssvlabs: The `spec-tests `should not be opened or changed manually; the whole idea is that this is automatic and not manual, so the PR should never be opened against a different base.

The fix could be applied, but IMHO this is very low risk.

.github/workflows/sync-spec-tests-merge.yaml
vaclav-ssvlabs: possibly yes, but again very low risk

.github/workflows/sync-spec-tests-pr.yaml
vaclav-ssvlabs: The majority of our CI are kinda build in a way that we trust our collaborators. If someone has write permissions, they could, in their PR, write any workflow file they want and potentially get this token.

Your suggestion does improve it, but by a small margin. A very malicious attacker could still get the token, but the token is only scoped to the spec-tests repo only.

.github/workflows/sync-spec-tests-pr.yaml
vaclav-ssvlabs: yes

.github/workflows/sync-spec-tests-pr.yaml
vaclav-ssvlabs: potentially yes

.github/workflows/sync-spec-tests-pr.yaml
vaclav-ssvlabs: all the tokens  are short lvied as they are generated on the run via the github app  credentials, but sure it could be cleaned out after the push

.github/workflows/sync-spec-tests-pr.yaml
vaclav-ssvlabs: AFAIK, you cannot have 2 branch  having the same name at ssv-spec at same time - which could create such a collision, if someone creates name and then close the PR, and then someone creates a new branch with same name on ssv-spec, then  there cannot be work collision, as the previous work was already deleted on the ssv-spec

Anyway, feel fee to pursue if you thnik this a actually a real scenario that can happen

.github/workflows/sync-spec-tests-pr.yaml
vaclav-ssvlabs: make sense

vaclav-ssvlabs:
I reacted to all comments regarding workflow, IMHO, the majority of them as very minor/low-risk things. There are a few of them that are worth pursuing, but since this is my last day, I don't have time to follow up and test all the things. So I will leave the solutions for you guys. Good luck!

qbft/spectest/generate/main.go
GalRogozinski: fixed

types/testingutils/comparable/helpers.go
GalRogozinski: please confirm if this is still an issue

momosh-ssv: Confirmed, resolved. Anchoring on `go.mod` and deriving the module from the relative path's first segment removes the basename guessing entirely — the `~/types/ssv-spec` case now errors out instead of silently resolving to the wrong module. Thanks for the fix 👍

qbft/spectest/generate/main.go
momosh-ssv: Verified. Paths are go.mod-anchored now, and `EnsureSpecTestsSubdir` asserts the target sits under `*/spec-tests/<module>` before `RemoveAll` — exactly the guard I was hoping for. Resolved.

.github/workflows/sync-spec-tests-pr.yaml
momosh-ssv: Fair that the org model trusts write-access collaborators and the token's scoped to spec-tests. 

Still, might be worth splitting this into an untrusted job that just runs `go generate` and uploads the artifacts, and a trusted job that downloads them and pushes with the App token? 

This is the one place a secret shares an environment with PR-controlled code, so a malicious same-repo branch could background a scraper for `steps.app-token.outputs.token`. The extra job removes that overlap by construction — I think it's worth it here.

.github/workflows/sync-spec-tests-pr.yaml
momosh-ssv: Might be worth a strict ref-shape guard early in the step? `HEAD_REF` still flows into `git checkout -B`, `push --force`, and `gh pr create --head`, so a name starting with `-`, containing `..`, or equal to `HEAD` could slip through. A one-line regex check closes it cheaply.

.github/workflows/sync-spec-tests-pr.yaml
momosh-ssv: Should we scope `gh pr view` to open PRs here? 

If a previous PR from this branch was closed without merge, `gh pr view <branch>` can return the closed one and we'd end up editing a dead PR. 

`gh pr list --head "${BRANCH_NAME}" --state open --json number` makes the open-vs-not check explicit. Same shape at `merge.yaml:156`.

.github/workflows/sync-spec-tests-pr.yaml
momosh-ssv: Worth a one-line comment pinning that `PR_TITLE` must stay in `env:` and never be inlined as `${{ github.event.pull_request.title }}`? 

I fear that possible future "simplification" back to inline interpolation would reopen the title→shell injection vector.

.github/workflows/sync-spec-tests-merge.yaml
momosh-ssv: Good points — branch protection prevents direct pushes to `main`, so the no-PR path effectively can't trigger, and you're right that the per-commit runs mean one failure wouldn't block subsequent syncs. Dropping this.

.github/workflows/sync-spec-tests-merge.yaml
momosh-ssv: Agreed — since the sync workflow controls both sides and spec-tests is never edited manually, the bases can't diverge. Dropping.

.github/workflows/sync-spec-tests-pr.yaml
momosh-ssv: Agreed it's low risk — the clone lives in a `mktemp -d` that dies with the runner and the token's short-lived. Dropping for now.

.github/workflows/sync-spec-tests-pr.yaml
momosh-ssv: Makes sense — branch names are unique at any one time on ssv-spec, so there's no live collision. Dropping.

.github/workflows/sync-spec-tests-merge.yaml
momosh-ssv: Agreed, low risk — the empty-array case is only reachable via a future refactor. Dropping for now.

.github/workflows/sync-spec-tests-pr.yaml
GalRogozinski: Done in ada2cb75d — split into two jobs.

`generate` checks out ssv-spec, runs `go generate`, tars the output and uploads it as an artifact. It references no secrets. `sync` does **not** check out ssv-spec at all: it downloads the tarball, unpacks it, mints the App token and pushes. So the token no longer coexists with anything the PR can execute; the generated tree only ever crosses the boundary as data.

One honest caveat, since you already raised the trust model: for `pull_request` the workflow file itself comes from the PR ref, so a collaborator can still rewrite this file and get the token. The split closes the accidental/generator path, not that one — the real control there is branch protection plus the token being scoped to spec-tests only.

Tarball rather than raw files because the generated tree is tens of thousands of small JSONs and `upload-artifact` is very slow file-by-file. Extraction is `tar -xzf ... -C generated`, and tar rejects absolute and `..` members unless `-P` is passed, so a generator-controlled filename cannot escape the directory.

.github/workflows/sync-spec-tests-pr.yaml
GalRogozinski: Done in ada2cb75d — guard added at the top of the step, before `BRANCH_NAME` reaches any git/gh call:

```bash
if ! [[ "${BRANCH_NAME}" =~ ^[A-Za-z0-9._/-]+$ ]] \
  || [[ "${BRANCH_NAME}" == -* ]] \
  || [[ "${BRANCH_NAME}" == *..* ]] \
  || [[ "${BRANCH_NAME}" == "HEAD" ]]; then
  echo "Refusing to sync: unsafe branch name '${BRANCH_NAME}'" >&2
  exit 1
fi
```

Checked against `--upload-pack=evil`, `-x`, `a..b`, `HEAD`, `a b`, `a$(id)`, `a;id`, empty, and non-ASCII — all rejected; `feature/foo_bar.v2` and friends still pass. The existing `main`/`master` check stays right after it.

.github/workflows/sync-spec-tests-pr.yaml
GalRogozinski: Done in ada2cb75d — both places now use an explicit open-PR lookup:

```bash
SPEC_PR_NUMBER=$(gh pr list --repo "${SPEC_TESTS_REPO}" \
  --head "${BRANCH_NAME}" --state open --json number --jq '.[0].number // empty')
```

`pr.yaml` branches on `-n "${SPEC_PR_NUMBER}"` and edits by number instead of by branch; `merge.yaml:156` gets the same call. Side benefit: the old `merge.yaml` line swallowed gh failures with `2>/dev/null || echo ""` and treated them as “no PR”, which it no longer does.

.github/workflows/sync-spec-tests-pr.yaml
GalRogozinski: Done in ada2cb75d — comment sits directly above the `env:` block so it is unmissable to anyone editing it:

```yaml
      - name: Sync to spec-tests PR branch
        # Attacker-controlled values (PR_TITLE, HEAD_REF) MUST stay in `env:` and
        # be referenced as shell variables. Inlining them as `${{ ... }}` in the
        # script below would splice them into the shell source: script injection.
        env:
```

GalRogozinski:
Picked up the outstanding review threads from @momosh-ssv now that @vaclav-ssvlabs has moved on. Addressed in ada2cb75d:

| Thread | Action |
|---|---|
| App token shares a runner with `go generate` | **Fixed** — `sync-spec-tests-pr` is now two jobs: `generate` (checkout + `go generate` + upload tarball, no secrets) and `sync` (download + push with App token, no ssv-spec checkout). |
| `HEAD_REF` ref-shape validation | **Fixed** — strict regex guard before it reaches `git checkout -B` / `git push` / `gh pr create --head`. |
| `gh pr view <branch>` can return a closed PR | **Fixed** — `gh pr list --head ... --state open` in both `pr.yaml` and `merge.yaml`. |
| `PR_TITLE` must stay in `env:` | **Fixed** — invariant documented above the `env:` block. |
| Token left in `.git/config`, PR-number branch namespacing, empty `GENERATED_PATHS`, no-PR-on-push-to-main, base-ref drift | Dropped by the reviewer as low risk — no change. |
| go.mod-anchored generator paths | Already fixed in cc74c0248 and confirmed by the reviewer. |

Also swapped the sync commit message from `#${PR_NUMBER}` to the full PR URL, same auto-link issue Greptile flagged for the PR body.

Note the split adds an artifact round-trip (one tarball, since tens of thousands of small JSONs would be far slower uploaded individually) plus one extra runner start per PR. Worth it to keep the token off the same runner as PR-controlled code.

Verified end to end on `ada2cb75d`: both jobs green, the artifact round-tripped, and the `sync` job cloned spec-tests, force-pushed `spec-test-ci`, and the new `gh pr list --state open` lookup correctly found and updated the existing ssvlabs/spec-tests#5. (Correcting myself — an earlier version of this comment said the repo variables weren't configured yet. They are; the workflow has been running for real.)

---

Unrelated defect the verified run exposed, not raised by any reviewer, **not** fixed here since it changes what this PR ships:

The spec-tests output currently contains every generated file **twice**, under two different path shapes:

```
types/tests/....json                          <- written directly by the anchored generator
types/spectest/generate/tests/....json        <- copied from the repo by the composite action
```

Cause: the merge from main in 5354bf5a7 resurrected 269 JSON files under `ssv/spectest/generate/{tests,state_comparison}` and `types/spectest/generate/{tests,state_comparison}` that `860caf4e2` had deleted (main had regenerated them via #601). The composite action's "Copy generated files to spec-tests output directory" step then rakes those stale files in and mirrors them under the old path.

Since cc74c0248, every generator writes straight to `../spec-tests/<module>/...` via `SpecTestsDirForModule`, and every reader resolves through `SpecTestsDirFrom` — I grepped, nothing reads the in-repo `*/spectest/generate/{tests,state_comparison}` paths any more. So the fix is to delete those 269 dead files and drop the now-redundant copy step from the composite action. Worth confirming main won't just resurrect them again on the next merge.

.github/workflows/sync-spec-tests-merge.yaml
```diff
@@ -0,0 +1,170 @@
...
+              echo "Branch '${PR_HEAD_REF}' not found in ${SPEC_TESTS_REPO}; verified no generated diff against '${RESOLVED_BASE}'."
+              exit 0
+            fi
+
+            echo "Branch '${PR_HEAD_REF}' not found in ${SPEC_TESTS_REPO}, but generated spec-tests differ from '${RESOLVED_BASE}'." >&2
+            echo "The PR sync workflow should have created the mirror branch." >&2
+            exit 1
+          fi
+
+          git checkout -B "${PR_HEAD_REF}" "origin/${PR_HEAD_REF}"
```
MatheusFranco99, line 184: **Blocker — this can force-push `main` in `spec-tests`.**

`sync-spec-tests-pr.yaml` validates `HEAD_REF` before using it (shape regex, no leading `-`, no `..`, not `HEAD`) and explicitly refuses `main`/`master`. None of that is applied to `PR_HEAD_REF` here, even though it flows into `git checkout -B`, `git push --force` and `gh pr list --head`.

Reachable path: a fork PR whose head branch is named `main` — very common, since contributors often work on their fork's default branch. The PR workflow correctly refuses to create a mirror branch for it. The PR gets merged anyway. This workflow then resolves `PR_HEAD_REF=main`, `git show-ref origin/main` succeeds at line 130, so it falls through to line 144, checks out `main` from `origin/main`, and line 151 runs `git push --force origin main` — writing generated fixtures straight to `spec-tests` `main`, bypassing the mirror PR entirely and clobbering anything that landed there during the run. It then `exit 1`s at line 168 for "no open spec-tests PR", i.e. *after* the push has happened.

`master` is reachable the same way from a same-repo branch.

Fix: apply the identical validation + protected-name refusal here. Better, see my note on branch naming in `sync-spec-tests-pr.yaml` — deriving the mirror branch from the PR number removes this whole class.

.github/workflows/sync-spec-tests-merge.yaml
```diff
@@ -0,0 +1,170 @@
...
+            RESOLVED_BASE="main"
+          fi
+
+          mapfile -d '' GENERATED_ROOTS < <(find "${GENERATED_DIR}" -maxdepth 1 -mindepth 1 -print0)
+          GENERATED_PATHS=()
+          for entry in "${GENERATED_ROOTS[@]}"; do
+            GENERATED_PATHS+=("$(basename "${entry}")")
+          done
+
+          sync_generated_paths() {
```
MatheusFranco99, line 161: **Extract this — the sync logic is duplicated across both workflows and has already diverged.**

`sync_generated_paths` + clone + reset + commit + push is ~60 lines copy-pasted between this file and `sync-spec-tests-pr.yaml`, and the copies are no longer equivalent: the branch-name validation exists there and not here, which *is* the bug I flagged at line 144. That divergence will recur.

Suggest `.github/scripts/sync-spec-tests.sh` (or a composite action) taking `branch`, `generated_dir`, `base_ref`, `mode=pr|merge`, called from both. It also shrinks the diff and gives the validation a single home.

.github/workflows/sync-spec-tests-pr.yaml
```diff
@@ -0,0 +1,182 @@
...
+          PR_URL: ${{ github.server_url }}/${{ github.repository }}/pull/${{ github.event.pull_request.number }}
+        run: |
+          set -euo pipefail
+
+          if [ -z "${SPEC_TESTS_REPO}" ]; then
+            echo "SPEC_TESTS_REPO variable is not set (e.g. ssvlabs/spec-tests)" >&2
+            exit 1
+          fi
+
+          BRANCH_NAME="${HEAD_REF}"
```
MatheusFranco99, line 120: **Mirror branch names collide — `head.ref` is not unique across forks.**

`github.head_ref` is a bare branch name with no fork owner, so two open PRs from different forks both on `patch-1` (or both on `main`) map to the same `spec-tests` branch and the same mirror PR, silently overwriting each other's fixtures.

Suggest namespacing on the PR number instead:

```
BRANCH_NAME="ssv-spec-pr-${{ github.event.pull_request.number }}"
```

Three things fall out for free: collisions become impossible, the branch can never be named `main`/`master` (which kills the force-push-to-main path in `sync-spec-tests-merge.yaml`), and the shape validation above stops being load-bearing since the value is no longer attacker-controlled. The merge workflow already resolves the PR number, so it can construct the same name rather than trusting `head.ref`.

.github/workflows/sync-spec-tests-pr.yaml
```diff
@@ -0,0 +1,182 @@
...
+        uses: actions/upload-artifact@v7.0.1
+        with:
+          name: spec-tests
+          path: spec-tests.tar.gz
+          retention-days: 1
+          if-no-files-found: error
+
+  # Holds the spec-tests App token. Runs no code from the PR: ssv-spec is never
+  # checked out here, the generated tree is only ever treated as data.
+  sync:
```
MatheusFranco99, line 53: **Fork PRs can't work — this job has no secrets.**

The trigger is `pull_request`, so for a PR from a fork `secrets.SPEC_TESTS_APP_PRIVATE_KEY` is empty and `create-github-app-token` fails. Every external contribution gets a failed sync job and no mirror branch, which then feeds directly into the merge-time problem I flagged in `sync-spec-tests-merge.yaml`.

Raising this as a question rather than prescribing a fix, because it's a policy call: **do we accept fork PRs on ssv-spec?**

- If no (internal branches only): simplest thing is to make it skip cleanly instead of failing — `if: github.event.pull_request.head.repo.full_name == github.repository` — and say so in the workflows README.
- If yes: the job split you've built here is already the right shape, it just needs the trusted half on a trusted trigger. Move `sync` to its own workflow on `workflow_run` (gated on `conclusion == 'success'`, downloading by `run-id`). It'd still never check out PR code. The wrinkle is that `workflow_run` gives no PR context, so the PR number/title have to travel inside the artifact — and they'd remain attacker-authored strings, so the `env:`-only discipline you already document below still applies to them.

.github/workflows/sync-spec-tests-pr.yaml
```diff
@@ -0,0 +1,182 @@
+name: Sync Spec Tests (PR)
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
```
MatheusFranco99, line 5: **No `concurrency` group.** Two pushes in quick succession to the same PR start two sync jobs racing on the same force-push and the same `gh pr list` / `gh pr create` — the classic window for a duplicate mirror PR or a lost sync.

```yaml
concurrency:
  group: sync-spec-tests-${{ github.event.pull_request.number }}
  cancel-in-progress: true
```

.github/workflows/sync-spec-tests-merge.yaml
```diff
@@ -0,0 +1,170 @@
+name: Sync Spec Tests (Merge)
+
+on:
+  push:
+    branches: [ "main" ]
```
MatheusFranco99, line 5: Same `concurrency` point as on the PR workflow, but note the setting differs here: this job pushes *and then merges*, so it must not be interrupted midway.

```yaml
concurrency:
  group: sync-spec-tests-merge
  cancel-in-progress: false
```

.github/workflows/sync-spec-tests-merge.yaml
```diff
@@ -0,0 +1,170 @@
...
+      - uses: actions/checkout@v6.0.2
+
+      - name: Validate required variables
+        run: |
+          if [ -z "${{ vars.SPEC_TESTS_REPO }}" ]; then
+            echo "SPEC_TESTS_REPO variable is not set (e.g. ssvlabs/spec-tests)" >&2
+            exit 1
+          fi
+
+      - name: Generate spec tests
```
MatheusFranco99, line 26: This job runs `go generate` (and `go get -v -t -d ./...`) in the same job that later holds `steps.app-token.outputs.token` — an installation token with contents+PR write on `spec-tests`. Anything executing in the job can read that token off the runner, including a malicious `//go:generate` directive or a compromised transitive dependency resolved at generate time.

`sync-spec-tests-pr.yaml` deliberately refuses that arrangement and documents why: `generate` holds no credentials, `sync` holds the token and runs no repo code. The justification for dropping the invariant here is presumably "post-merge code is trusted" — true of the repo content, but not of the dependency graph, which is the same reasoning the PR workflow declined to accept.

Since the artifact handoff already exists, mirroring the split costs a `needs:` plus an upload/download pair. If you'd rather keep one job, please add a comment stating the trust assumption explicitly, so the asymmetry between the two workflows reads as a decision rather than an oversight.

.github/workflows/sync-spec-tests-pr.yaml
```diff
@@ -0,0 +1,182 @@
...
+      - name: Generate spec tests
+        uses: ./.github/actions/generate-spec-tests
+
+      # One tarball rather than raw files: the generated tree is tens of
+      # thousands of small JSONs, which upload-artifact handles poorly.
+      - name: Pack generated spec tests
+        run: tar -czf spec-tests.tar.gz -C "${GITHUB_WORKSPACE}/.." spec-tests
+
+      - name: Upload generated spec tests
+        uses: actions/upload-artifact@v7.0.1
```
MatheusFranco99: **Pin third-party actions by commit SHA, not tag** — at minimum in the jobs that touch the App private key (here, `create-github-app-token`, `download-artifact`, and `checkout` above).

`@v7.0.1` resolves a git *tag*, which is mutable: whoever controls the action repo can retag it to different code and your next run executes that code in a job holding the token. This isn't hypothetical — in the `tj-actions/changed-files` compromise every version tag was moved to a commit that dumped runner memory (secrets included) into public workflow logs.

```yaml
uses: actions/upload-artifact@<full-40-char-sha> # v7.0.1
```

Dependabot understands that form and bumps both the SHA and the trailing comment, so nothing is lost on upgrades.

.github/workflows/lint.yaml
```diff
@@ -21,8 +21,8 @@ jobs:
       - name: Get Dependencies
         run: go get -v -t -d ./...
 
-      - name: Prepare
-        run: make lint-prepare
-
       - name: Lint
-        run: make lint
+        uses: golangci/golangci-lint-action@v9
```
MatheusFranco99: Good change, but it leaves `make lint` and CI disagreeing. The Makefile's `lint-prepare` still installs `latest` into `./bin` while CI pins v2.12.2 via the action, and there's no `.golangci.yml` in the repo — so local and CI run different versions against different defaults, and golangci-lint v2 changed its default linter set relative to v1.

Suggest pinning the Makefile to the same version (or deleting `lint-prepare`/`lint` now that they're unused by CI), and ideally committing a v2-format `.golangci.yml` so both sides read one source of truth.

.github/workflows/test.yaml
```diff
@@ -8,21 +8,15 @@ on:
 
 jobs:
 
-  build:
-    runs-on: ubuntu-latest
+  test:
```
MatheusFranco99, line 11: Heads-up, operational rather than code: the job id changed from `build` to `test`. If `build` is configured as a required status check on `main`, PRs will block forever waiting for a check that no longer reports. Either keep the old job id or update the branch-protection rule as part of merging this.

.github/workflows/README.md
```diff
@@ -0,0 +1,112 @@
...
+| `test.yaml` | PR, push to `main` | Build, generate, run tests |
+| `sync-spec-tests-pr.yaml` | PR opened / updated / reopened | Sync generated files to spec-tests, create/update PR |
+| `sync-spec-tests-merge.yaml` | Push to `main` | Push final files to spec-tests branch, merge spec-tests PR |
+| `../.github/actions/generate-spec-tests/action.yaml` | (composite, called by all above) | Set up Go, generate JSON fixtures |
+
+---
+
+## Workflows in Detail
+
+### `test.yaml`
```
MatheusFranco99, line 58: This documents the CI pipeline well, but nothing tells a *developer* how to run the tests locally now, and the local workflow has changed materially:

- fixtures resolve to `<parent-of-ssv-spec>/spec-tests/<module>`, so a sibling directory is required;
- generation must precede `make test` — a fresh clone that runs `make test` straight away now fails, where before the committed JSONs made it work;
- that sibling path isn't namespaced per checkout, so two worktrees under one parent share a single `spec-tests` and will overwrite each other's generated output.

Worth a short section here and a pointer from the root README/CONTRIBUTING, since it's the first thing every maintainer will hit.

MatheusFranco99 (changes requested):
Reviewed with a focus on correctness and the operational surface this adds. The direction is right and the security reasoning in `sync-spec-tests-pr.yaml` is genuinely above average for CI work — the `generate`/`sync` job split, untrusted values passed via `env:` rather than spliced into shell, the branch-name shape check, the `--state open` scoping. Those comments explaining *why* are valuable; please keep them.

Two themes in the findings:

1. **`sync-spec-tests-merge.yaml` didn't get the same hardening as the PR workflow.** The missing branch-name validation there is reachable and can force-push `spec-tests` `main`, bypassing the mirror PR. The root cause is that ~60 lines of sync logic are copy-pasted between the two files and have drifted.
2. **Identity of the mirror branch.** Keying it on `head.ref` makes it non-unique across forks and lets it take a protected name; keying it on the PR number removes both problems and most of the validation burden.

Requesting changes on the first of those. Everything else is either a small fix or a question.

Blocking:
- force-push-to-`main` path in `sync-spec-tests-merge.yaml`
- extract the shared sync script (root cause of the above)
- branch naming on `head.ref`

Non-blocking but worth doing now: `concurrency` groups on both workflows, SHA-pinning the actions in the token-holding jobs, the `make lint` / CI version divergence, and the local-development docs.

Open question rather than a finding: whether fork PRs are in scope at all — see my comment on the `sync` job. As written they fail on missing secrets, and that failure is what makes the merge-time path above dangerous.

One nit on the PR description itself: it says the composite action "copies output to `../spec-tests`", but the action's own comment correctly notes the generators write there directly and there's nothing to copy. Worth updating so the description doesn't send the next reader looking for a copy step.

Not raised as findings, for the record: the `os.RemoveAll` on a path outside the repo, the absence of an override for the hardcoded `../spec-tests`, and assorted style nits — discussed and deliberately deferred.

.github/workflows/README.md
```diff
@@ -0,0 +1,112 @@
...
+| `SPEC_TESTS_REPO` | `ssvlabs/spec-tests` |
+| `SPEC_TESTS_APP_ID` | `12345` |
+
+**Secrets:**
+
+| Name | Value |
+|---|---|
+| `SPEC_TESTS_APP_PRIVATE_KEY` | Contents of the `.pem` private key file for the GitHub App |
+
+See the [GitHub App setup guide](https://github.com/ssvlabs/ssv-spec/blob/main/.github/workflows/GITHUB_APP_SETUP.md) for step-by-step instructions on creating the app.
```
iurii-ssv: Broken link: this points to `.github/workflows/GITHUB_APP_SETUP.md`, but that file isn't added anywhere in this PR and doesn't exist on `main`, so the "GitHub App setup guide" 404s. Since this is the doc an operator needs to create the App and populate `SPEC_TESTS_APP_ID` / `SPEC_TESTS_APP_PRIVATE_KEY`, consider adding the guide in this PR or dropping the link until it exists.

.github/workflows/README.md
```diff
@@ -0,0 +1,112 @@
...
+---
+
+## Files
+
+| File | Trigger | Purpose |
+|---|---|---|
+| `test.yaml` | PR, push to `main` | Build, generate, run tests |
+| `sync-spec-tests-pr.yaml` | PR opened / updated / reopened | Sync generated files to spec-tests, create/update PR |
+| `sync-spec-tests-merge.yaml` | Push to `main` | Push final files to spec-tests branch, merge spec-tests PR |
+| `../.github/actions/generate-spec-tests/action.yaml` | (composite, called by all above) | Set up Go, generate JSON fixtures |
```
iurii-ssv: Minor: this path looks off by one. The other rows in this table use bare filenames (relative to this README's own dir, `.github/workflows/`), so for consistency the action should read `../actions/generate-spec-tests/action.yaml` — `../.github/actions/...` resolves to `.github/.github/actions/...`. It's plain text rather than a link, so nothing breaks; just slightly misleading.

qbft/spectest/run_test.go
```diff
@@ -28,7 +29,11 @@ func TestAll(t *testing.T) {
 
 func TestJson(t *testing.T) {
 	basedir, _ := os.Getwd()
-	path := filepath.Join(basedir, "generate", "tests.json")
+	specTestsDir, err := typescomparable.SpecTestsDirFrom(basedir)
+	if err != nil {
+		panic(err.Error())
```
iurii-ssv: Minor consistency nit (partly subjective): the newly-added `SpecTestsDirFrom` failure uses `panic(err.Error())`, whereas the read/unmarshal handlers immediately below it use `t.Fatalf` — as does the equivalent resolver check in [`ssv/spectest/run_test.go`](https://github.com/ssvlabs/ssv-spec/blob/spec-test-ci/ssv/spectest/run_test.go#L58) (`t.Fatalf("Failed to resolve spec-tests dir: %v", err)`). [`types/spectest/run_test.go`](https://github.com/ssvlabs/ssv-spec/blob/spec-test-ci/types/spectest/run_test.go#L45) has the same `panic`. Suggest matching `t.Fatalf` in qbft and types so a resolution failure surfaces as a clean test failure. Not a bug — and note qbft's `default:` switch case (line 136) also panics, so this is about local consistency of the new code, not a hard rule.

iurii-ssv (commented):
Couple minor things.

.github/workflows/sync-spec-tests-merge.yaml
GalRogozinski: @zktaiga I think the token is very limited (only controls the ssv-spec repo) can you double check and confirm the severity.

.github/workflows/sync-spec-tests-merge.yaml
GalRogozinski: Fixed in [`25abb89c8`](https://github.com/ssvlabs/ssv-spec/commit/25abb89c8) — same shape check and `main`/`master` refusal as `sync-spec-tests-pr.yaml:114–126`, applied where `PR_HEAD_REF` is resolved so it fails before any `git` or `gh` call.

Confirmed reachable, for the record: `spec-tests` `main` has no branch protection, and fork PRs are merged here regularly (#583, #581, #564, #523, #509), so `PR_HEAD_REF=main` is a live input — and `origin/main` always exists, so the line-130 check can't catch it.

Not taking the PR-number branch naming: it changes the mirror-branch contract both workflows depend on, and the fork policy in this branch closes the collision class another way. Happy to revisit as a follow-up.

.github/workflows/sync-spec-tests-merge.yaml
GalRogozinski: Not extracting this — most of the divergence is deliberate, and the premise behind it was settled earlier in this PR.

The PR workflow resets the mirror branch to the base while the merge workflow continues the existing one and treats a missing branch as an error; no-diff exits 0 there and continues to the merge here; the tails are `gh pr create`/`edit` versus `gh pr merge --delete-branch`; and the PR workflow splits `generate` from `sync` so the codegen job holds no token. A shared script with `mode=pr|merge` would fold those apart-on-purpose paths back into one file and make each future change touch both.

`spec-tests` being 100% machine-generated and never manually edited is intentional ([r2974727423](https://github.com/ssvlabs/ssv-spec/pull/610#discussion_r2974727423), [r3323711855](https://github.com/ssvlabs/ssv-spec/pull/610#discussion_r3323711855)).

The one part that genuinely was a divergence rather than a design choice — the missing branch-name validation — is fixed in [`25abb89c8`](https://github.com/ssvlabs/ssv-spec/commit/25abb89c8).

.github/workflows/sync-spec-tests-pr.yaml
GalRogozinski: Not renaming the mirror branches — keeping `head_ref`.

Fork PRs stay allowed but are now skipped by the pre-merge sync and discouraged in the workflows README ([`9d81dc743`](https://github.com/ssvlabs/ssv-spec/commit/9d81dc743)), so in practice every mirror branch comes from an in-repo branch — and ssv-spec can't hold two branches of the same name at once, per [r3323774669](https://github.com/ssvlabs/ssv-spec/pull/610#discussion_r3323774669). Fork PRs should be rare here.

The residual is real and we're accepting it: nothing structural stops a fork branch name from colliding if the pre-merge sync is ever opened to forks. It takes a deliberate policy change to become reachable, and the force-push-to-`main` variant is already closed by [`25abb89c8`](https://github.com/ssvlabs/ssv-spec/commit/25abb89c8).

`ssv-spec-pr-<number>` would remove the class by construction, but it changes a branch-name contract both workflows depend on and loses the readable link between a mirror branch and its PR. Not worth restructuring the design at this point — happy to revisit as a follow-up if fork contributions become common.

.github/workflows/sync-spec-tests-pr.yaml
GalRogozinski: Answering the policy question: fork PRs are supported but not the recommended path. Taking your "if no" option for the pre-merge half, and closing the post-merge gap separately. Commits [`9d81dc743`](https://github.com/ssvlabs/ssv-spec/commit/9d81dc743) → [`750163aec`](https://github.com/ssvlabs/ssv-spec/commit/750163aec).

`sync-spec-tests-pr.yaml` now skips fork PRs via `if: github.event.pull_request.head.repo.full_name == github.repository` on both jobs, instead of failing at `create-github-app-token`. `sync-spec-tests-merge.yaml` derives `is-fork` from the PR lookup it already does and skips its push step for fork-originated merges.

Not taking the `workflow_run` restructure. It would let any GitHub user cause App-token writes to `spec-tests` — arbitrary mirror branches and PRs, on every push, with no review — which is a new class of external write access. Post-merge sync stays inside the trust boundary already accepted here: the merge workflow runs merged code next to the token regardless, and a human reviewed it first.

Honest limitation, now documented in the workflows README: a merged fork PR's fixtures reach `spec-tests` with the next in-repo merge **that has a mirror branch**. An in-repo PR whose own fixtures already match `spec-tests` never pushes one, so if such a PR merges next, its merge run finds no mirror branch, sees the fork's fixture diff, and fails loudly rather than pushing; the sync completes on the following in-repo merge that does have one. Pushing to that PR's branch (or requiring branches be up to date) also clears it. `spec-tests` is only ever behind, never wrong.

What fork contributors lose is the mirror-PR preview while their PR is open. `test.yaml` needs no secrets, so their fixtures are still generated and tested.

.github/workflows/sync-spec-tests-pr.yaml
GalRogozinski: Done in [`286a41558`](https://github.com/ssvlabs/ssv-spec/commit/286a41558) — workflow-level group keyed on `pull_request.number`, `cancel-in-progress: true`.

One correction: a duplicate mirror PR isn't reachable. GitHub rejects a second PR for the same head/base (422), so that race fails the job rather than duplicating. The lost-sync half is the real one: two runs force-push the same mirror branch (line 165) with no ordering, so the older commit's fixtures can land last.

Accepting two limits. The group gives mutual exclusion, not ordering, so a manual re-run of an older commit can still push stale fixtures. And a cancel between the push at 165 and `gh pr create` at 177 leaves a branch with no PR — the next push repairs it, and the merge workflow otherwise fails loudly rather than syncing something wrong.

.github/workflows/sync-spec-tests-merge.yaml
GalRogozinski: Not adding it here, and the reason is the static group rather than `cancel-in-progress: false`.

GitHub keeps only one *pending* run per concurrency group and cancels the previously pending one, so three merges to `main` in quick succession would drop the middle run entirely. Each run finalizes a different PR's mirror branch, so a dropped one leaves that spec-tests PR open with nothing to retry it — worse than the race it prevents.

The runs also share no mutable state: each clones into its own `mktemp -d` and force-pushes its own `${PR_HEAD_REF}` (line 151). And `cancel-in-progress: false` doesn't make the push→merge window atomic anyway — manual cancel, job timeout and runner loss all still land inside it; re-running the push event is what recovers.

If you want protection here, the shape that doesn't evict queued merges is a per-commit group:

```yaml
concurrency:
  group: sync-spec-tests-merge-${{ github.sha }}
  cancel-in-progress: false
```

Happy to add that if you'd rather have it explicit than absent.

.github/workflows/lint.yaml
GalRogozinski: Done in [`fb533f0a6`](https://github.com/ssvlabs/ssv-spec/commit/fb533f0a6) and [`a2505dd9d`](https://github.com/ssvlabs/ssv-spec/commit/a2505dd9d) — `lint-prepare` now installs `$(GOLANGCI_LINT_VERSION)`, defaulted to v2.12.2 with a comment pointing at `lint.yaml`.

The pin alone wasn't enough: the `install.sh` served from the golangci-lint `master` branch fails sha256 verification for v2 releases, so `make lint-prepare` installed nothing.

```
hash_sha256_verify checksum for golangci-lint-2.12.2-darwin-arm64.tar.gz did not verify
```

[`a2505dd9d`](https://github.com/ssvlabs/ssv-spec/commit/a2505dd9d) switches to `https://golangci-lint.run/install.sh`, which installs the pinned version — same change as #629, kept together with the pin so local and CI stay on one version.

Leaving `.golangci.yml` out for now: with versions matched, both sides read the same v2 defaults. Say if you'd rather the linter set be explicit.

.github/workflows/test.yaml
GalRogozinski: non-blocking: verified on this PR — the rollup is green with `Test / test`, `Lint / build` and both sync jobs, so nothing is waiting on a missing check.

Worth knowing why: `lint.yaml`'s job is also called `build` (line 11), so a required check named `build` still reports, just from Lint rather than Test. The rename is currently harmless, but that safety is incidental — renaming Lint's job later would break it.

`main` is covered by an org-level ruleset rather than classic branch protection, and its rules aren't readable without org admin. Flagging for whoever merges to confirm no required check is pinned to `Test / build`; no code change.

.github/workflows/README.md
GalRogozinski: Done in [`75cbb177e`](https://github.com/ssvlabs/ssv-spec/commit/75cbb177e) (branch `pr-610-fixes`, #635) — added a **Running Locally** section here covering the sibling `<parent-of-ssv-spec>/spec-tests/<module>` layout, generate-before-`make test` (`make test` has no `generate-jsons` dependency), and the shared-directory caveat for multiple checkouts, plus a pointer from the root README.

The multi-checkout collision is documented rather than fixed: the path is `<repo-root>/../spec-tests/<module>` (`types/testingutils/comparable/helpers.go:207`) with no override. Say if you'd rather it be namespaced per checkout.

.github/workflows/README.md
GalRogozinski: Done in [`9cb764328`](https://github.com/ssvlabs/ssv-spec/commit/9cb764328) — dropped the dead link rather than adding the guide in this PR.

Replaced with a pointer to GitHub's own [Registering a GitHub App](https://docs.github.com/en/apps/creating-github-apps/registering-a-github-app/registering-a-github-app) docs, plus a line naming where the App ID and `.pem` go (`SPEC_TESTS_APP_ID` / `SPEC_TESTS_APP_PRIVATE_KEY`). The two required permissions were already in the table above it, so an operator has everything without a second file to keep in sync.

.github/workflows/README.md
GalRogozinski: Done in [`9cb764328`](https://github.com/ssvlabs/ssv-spec/commit/9cb764328) — reads `../actions/generate-spec-tests/action.yaml` now.

qbft/spectest/run_test.go
GalRogozinski: Done in [`9cb764328`](https://github.com/ssvlabs/ssv-spec/commit/9cb764328) — `t.Fatalf("Failed to resolve spec-tests dir: %v", err)` in both `qbft/spectest/run_test.go` and `types/spectest/run_test.go`, matching `ssv/spectest/run_test.go:58`.

Left the `default:` case panic in qbft alone — that one is an unreachable-by-construction guard rather than a runtime failure a test run can hit.

.github/workflows/sync-spec-tests-merge.yaml
GalRogozinski: Not splitting this one — the asymmetry with the PR workflow is deliberate.

That workflow triggers on `pull_request`, so its `go generate` runs code from an unmerged branch; that's why it got the two-job split in [`ada2cb75d`](https://github.com/ssvlabs/ssv-spec/commit/ada2cb75d). This one triggers on `push: main`, and every push to `main` comes from a merged PR — direct pushes are blocked by org settings ([r3323697738](https://github.com/ssvlabs/ssv-spec/pull/610#discussion_r3323697738)). The generator running beside the token here has already been reviewed, so there's no PR-controlled code in this job to isolate.

Same blast radius as before: the token is scoped to `spec-tests` only, and anyone with write access can already author a workflow that reads it ([r3323736448](https://github.com/ssvlabs/ssv-spec/pull/610#discussion_r3323736448)).

iurii-ssv (approved):
LGTM

PR #613, opened 2026-03-26
chore(ci) - Update actions base image to node:24

vaclav-ssvlabs:
## Summary
GitHub is enforcing Node.js 24 for GitHub Actions runtime and will block workflows that rely on older Node runtimes.

This PR updates CI workflows to align with the requirement.

## Changes
- Update reusable action references (`uses:`) to latest released versions.
- Update workflow Node container images (`node:<version>`) to `node:24` where they were below 24.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6.3.0
         with:
           go-version: 1.22.x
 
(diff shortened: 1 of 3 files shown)


iurii-ssv (commented):
Seems like we'd need to upgrade Golang to `>= 1.25.0` in order to resolve the failing pipeline due to:
- upstream actions/setup-go maintainers introduced the GOTOOLCHAIN=local behavior in v6 which prevents the pipeline from automatically auto-switching to Go 1.25+
- this PR switched from actions/setup-go@v5 to actions/setup-go@v6, which is what activated that behavior in this pipeline

PR #614, opened 2026-03-30
fix(round change) - advance round only after successful broadcast

oleg-ssvlabs:
## Summary

  Fix QBFT round advancement so an instance only moves to the next round after a round change message is successfully broadcast.

  ## Problem

  The timeout path advanced `State.Round` and reset the timer via `defer`, which meant the local instance could move
  to the next round even if:
  - `CreateRoundChange()` failed
  - `Broadcast()` failed

  The same behavior existed in the partial-quorum round change path.

  This could leave a node ahead of the committee locally without ever sending its round change message, which is a
  liveness / desynchronization risk.

  ## Changes

  - added a shared `advanceRound()` helper for local round state updates
  - updated `UponRoundTimeout()` to advance round only after successful round change broadcast
  - updated `uponChangeRoundPartialQuorum()` to do the same
  - preserved current behavior for successful round changes

  ## Why

  This keeps local round state, timer progression, and network-visible round changes consistent:
  - if round change creation or broadcast fails, the instance stays in the current round
  - the node does not reject current-round messages prematurely
  - timer state is not advanced unless the round transition was actually emitted

  ## Tests

  Added targeted QBFT tests covering broadcast failure cases:
  - timeout path does not advance round on failed round change broadcast
  - partial-quorum path does not advance round on failed round change broadcast

  ## Verification

  - `go test ./qbft ./types/testingutils`
  
  Closes https://github.com/ssvlabs/ssv-node-board/issues/844

### DIFF
--- a/qbft/round_change.go
+++ b/qbft/round_change.go
@@ -85,9 +85,6 @@ func (i *Instance) uponRoundChange(
 }
 
 func (i *Instance) uponChangeRoundPartialQuorum(newRound Round, instanceStartValue []byte) error {
-	i.State.Round = newRound
-	i.State.ProposalAcceptedForCurrentRound = nil
-	i.config.GetTimer().TimeoutForRound(i.State.Round)
 	roundChange, err := CreateRoundChange(i.State, i.signer, newRound, instanceStartValue)
 	if err != nil {
 		return errors.Wrap(err, "failed to create round change message")
@@ -95,6 +92,7 @@ func (i *Instance) uponChangeRoundPartialQuorum(newRound Round, instanceStartVal
 	if err := i.Broadcast(roundChange); err != nil {
 		return errors.Wrap(err, "failed to broadcast round change message")
 	}
+	i.advanceRound(newRound)
 	return nil
 }
 
--- a/qbft/timeout.go
+++ b/qbft/timeout.go
@@ -18,17 +18,18 @@ var (
 	CutoffRound = 12 // stop processing attestations after 8*2+120*3 = 6.2 min (~ 1 epoch)
 )
 
+func (i *Instance) advanceRound(newRound Round) {
+	i.State.Round = newRound
+	i.State.ProposalAcceptedForCurrentRound = nil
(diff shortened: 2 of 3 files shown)


nkryuchkov (approved):
The code looks good to me, but the behavioral change needs to be approved by the spec team

GalRogozinski (changes requested):
I am rejecting this.
Basically once F+1 round changes have been made the node should go to the next round. This is because at this point no quorum ($Q$) for consensus should be achievable on the current round for $N$ nodes.  $N-(F+1) = (3F+1) - (F+1) = 2F < Q = 2F + 1$

Implementation can have a broadcast retry mechanism that respects message validation.
Please @oleg-ssvlabs close the issue if you agree.

oleg-ssvlabs:
> I am rejecting this. Basically once F+1 round changes have been made the node should go to the next round. This is because at this point no quorum ( Q ) for consensus should be achievable on the current round for N nodes. N − ( F + 1 ) = ( 3 F + 1 ) − ( F + 1 ) = 2 F < Q = 2 F + 1
> 
> Implementation can have a broadcast retry mechanism that respects message validation. Please @oleg-ssvlabs close the issue if you agree.

After analysis, I agree this change is harmful and the original behavior (advance round regardless of broadcast success) is correct.                                                                                                
                            
  The PR moves `advanceRound` to after a successful broadcast in two paths:                                             
                                                                                                                      
  1. _Timeout path_ (`UponRoundTimeout`): Each node's timer fires independently. If this node fails to broadcast but other nodes broadcast successfully, the network moves to the next round. With this PR, the node stays on the old round with no timer restart, permanently stuck with no recovery path.                                                    
                            
  2. _Partial quorum path_ (`uponChangeRoundPartialQuorum`): The node already received F+1 round change messages, proving other nodes have moved on. Even if this node fails to broadcast its own round change, those other nodes still broadcasted theirs. The leader may still reach quorum without us. With this PR, the node refuses to advance despite having evidence the network is moving on.
  
Adding small diagram explaining the problem:

image

  
Closing this PR

PR #620, opened 2026-04-05
Standardize sentinel pattern (F-ssv-spec-027)

julienharbulot:
Fix for:

Inconsistent Enum Sentinel Values
ssv-spec · Inconsistency · Assigned: dev4 · Effort: —
`BNRoleUnknown = math.MaxUint64` (BeaconRole) vs `RoleUnknown = -1` (RunnerRole as int32). The RunnerRole sentinel `-1` can't be properly represented as uint32 (line 50-52 in messages.go converts to uint32), potentially matching corrupted data.

Impact: Type confusion between sentinel values. Corrupted message data could match `RoleUnknown` unintentionally.

Recommendation: Standardize sentinel pattern across all enums.

### DIFF
--- a/types/runner_role.go
+++ b/types/runner_role.go
@@ -1,5 +1,7 @@
 package types
 
+import "math"
+
 // https://github.com/ssvlabs/ssv-spec/issues/423 - this can be taken down to one byte on the wire
 type RunnerRole int32
 
@@ -11,6 +13,30 @@ const (
 	RoleAggregatorCommittee   = RunnerRole(6) // Combines aggregator and sync committee contribution duties
 	RoleUnknown               = RunnerRole(-1)
 )
+
+const runnerRoleUnknownWireUint32 = uint32(math.MaxUint32)
+
+// WireUint32 returns the canonical on-wire representation for RunnerRole.
+//
+// Note: RunnerRole is encoded as a 4-byte little-endian uint32 in MessageID.
+// RoleUnknown uses the all-ones sentinel (MaxUint32), consistent with other enums
+// (e.g. BeaconRole uses MaxUint64).
+func (r RunnerRole) WireUint32() uint32 {
+	if r == RoleUnknown || r < 0 {
+		return runnerRoleUnknownWireUint32
+	}
+	return uint32(r)
+}
+
+// RunnerRoleFromWireUint32 converts a 4-byte uint32 MessageID role field into RunnerRole.
+// Values outside the int32 range are treated as unknown to avoid surprising negative roles.
+func RunnerRoleFromWireUint32(v uint32) RunnerRole {
+	// Any value outside the int32 range (including the RoleUnknown sentinel) is treated as unknown.
+	if v > uint32(math.MaxInt32) {
+		return RoleUnknown
+	}
+	return RunnerRole(int32(v))
+}
 
 // String returns the name of the runner role
 func (r RunnerRole) String() string {
--- a/types/runner_role_test.go
+++ b/types/runner_role_test.go
@@ -1,0 +1,49 @@
+package types
+
+import (
+	"encoding/binary"
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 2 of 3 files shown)


types/runner_role_test.go
```diff
@@ -0,0 +1,54 @@
+package types
+
+import (
+	"encoding/binary"
+	"math"
+	"testing"
+)
+
+func TestRunnerRoleWireMapping(t *testing.T) {
```
olegshmuelov, line 11: The test covers `RoleUnknown` sentinel and the `0x80000000` boundary, but doesn't exercise the `r < 0` guard in `WireUint32()` for non-sentinel negative values. Since `WireUint32` has an explicit `r < 0` branch (defense-in-depth beyond just `r == RoleUnknown`), a single test case would document that contract.

Suggestion:
```go
// Arbitrary negative role (not RoleUnknown) should also map to sentinel.
if got := RunnerRole(-42).WireUint32(); got != uint32(math.MaxUint32) {
    t.Fatalf("RunnerRole(-42).WireUint32() = %d, want %d", got, uint32(math.MaxUint32))
}
```

julienh-ssv: Good point — added a negative (non-sentinel) role case to document the `r < 0` contract. Fixed in aa3ead15.

olegshmuelov (dismissed):

types/runner_role.go
```diff
@@ -12,6 +14,30 @@ const (
...
+		return runnerRoleUnknownWireUint32
+	}
+	return uint32(r)
+}
+
+// RunnerRoleFromWireUint32 converts a 4-byte uint32 MessageID role field into RunnerRole.
+// Values outside the int32 range are treated as unknown to avoid surprising negative roles.
+func RunnerRoleFromWireUint32(v uint32) RunnerRole {
+	// Any value outside the int32 range (including the RoleUnknown sentinel) is treated as unknown.
+	if v > uint32(math.MaxInt32) {
```
nkryuchkov, line 35: `uint32(math.MaxUint32)` is also an unknown role, right? 
```suggestion
	if v >= uint32(math.MaxInt32) {
```

types/runner_role_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	for _, role := range []RunnerRole{
+		RoleCommittee,
+		RoleProposer,
+		RoleValidatorRegistration,
+		RoleVoluntaryExit,
+		RoleAggregatorCommittee,
+	} {
+		if got := RunnerRoleFromWireUint32(role.WireUint32()); got != role {
+			t.Fatalf("round-trip(%v) = %v, want %v", role, got, role)
```
nkryuchkov: can we use `testify` in this file?

types/runner_role.go
julienh-ssv: No here we only exclude values out of the Int32 range to avoid negative values as explained in the function's header comment. `math.MaxInt32` is still a potentially valid RunnerRole. It's a different layer of the code that will decide it.

nkryuchkov: sorry, I misread it as `math.MaxUint32` and I thought `runnerRoleUnknownWireUint32` covers it

nkryuchkov (dismissed):

ljuba-ssv (dismissed):

MatheusFranco99 (commented):
Good change! Looks good to me! :)
However, I didn't get why the .json file tests got modified (since the edge case was already the same 0xff... , and some tests don't seem related to the issue)
Can you give me a help on that?

julienharbulot:
@MatheusFranco99 I had regenerated the fixtures after merge but you are right it wasn't necessary. I have cleaned up the git history and removed the unnecessary changes to allow for a clean merge on top of main

olegshmuelov (approved):

nkryuchkov (approved):

julienharbulot:
Fixes https://github.com/ssvlabs/ssv-node-board/issues/866 
@MatheusFranco99 ready for your review

PR #621, opened 2026-04-06
Add validation for core protocol types (F-ssv-spec-026)

julienharbulot:
F-ssv-spec-026 P1 Triaged
Missing Validate() Methods on Core Types
ssv-spec · Spec Gap · Assigned: dev4 · Effort: —

Multiple protocol-critical types lack `Validate()` methods:
- `SSVMessage` — no validation of MsgType range, Data size, MsgID structure
- `Share` — no validation of Graffiti (≤32 bytes), Committee size, DomainType
- `Operator` — no validation of SSVOperatorPubKey length (should be 459 bytes)
- `CommitteeMember` — no validation of Committee non-empty, FaultyNodes bounds
- `BeaconVote` — no validation that Source.Epoch < Target.Epoch
- `ValidatorDuty` — no Validate() method at all

Impact: Implementers must guess valid ranges. Different implementations may accept different inputs, causing consensus failures.

Recommendation: Add `Validate()` to all types that cross trust boundaries.


types/messages.go
```diff
@@ -123,6 +125,36 @@ func (msg *SSVMessage) Decode(data []byte) error {
...
+	case SSVConsensusMsgType, SSVPartialSignatureMsgType, DKGMsgType:
+		// ok
+	default:
+		return NewError(MessageTypeInvalidErrorCode, "invalid SSV message type")
+	}
+
+	// Role is used by higher-level components (committee/runner selection) and not all code paths
+	// enforce the same constraints. Here we only reject negative role values (including RoleUnknown),
+	// to avoid accepting wrapped/invalid encodings.
+	if msg.MsgID.GetRoleType() < 0 {
```
olegshmuelov: Doc comment says "MsgID must encode a known RunnerRole" but the implementation only rejects negative values (including RoleUnknown=-1). Positive undefined roles (1, 3, 7, 8...) pass validation.

The inline comment explains this is intentional for forward compatibility, which is reasonable for a spec. But the doc comment is misleading — it overpromises what the check actually enforces. Implementers reading the doc will expect stricter validation than what exists.

Suggestion: align the doc with the implementation.
```go
// - MsgID must encode a non-negative RunnerRole (unknown positive roles are
//   accepted for forward compatibility; routing rejects them downstream)
```

types/consensus_data.go
```diff
@@ -122,6 +122,19 @@ func (b *BeaconVote) Decode(data []byte) error {
 	return b.UnmarshalSSZ(data)
 }
 
+// Validate checks the following rules:
+// - Source and Target checkpoints must be non-nil
+// - Source.Epoch must be strictly less than Target.Epoch
+func (b *BeaconVote) Validate() error {
+	if b == nil || b.Source == nil || b.Target == nil {
```
olegshmuelov: When `b` itself is nil, the error message says "nil source or target checkpoint" which is slightly misleading — the caller gets no signal that the entire BeaconVote was nil vs. just a checkpoint field. Consider splitting:

```go
if b == nil {
    return NewError(BeaconVoteNilCheckpointErrorCode, "nil beacon vote")
}
if b.Source == nil || b.Target == nil {
    return NewError(BeaconVoteNilCheckpointErrorCode, "nil source or target checkpoint")
}
```

types/validation_test.go
olegshmuelov, line 1: The test file covers SSVMessage, SignedSSVMessage delegation, and BeaconVote validation. But for a spec repo, the following types have new `Validate()` methods with **no dedicated unit tests**:

- `Operator.Validate()` — pubkey length, zero ID
- `CommitteeMember.Validate()` — faulty nodes formula, duplicate operators, self-membership, CommitteeID hash, domain type
- `Share.Validate()` — SharePubKey length, committee bounds, duplicate signers, graffiti, domain
- `ValidatorDuty.Validate()` — duty type mapping, zero pubkey, committee index bounds, sync indices limit
- `DomainType.IsKnown()` — known vs unknown domains

These are implicitly covered by spec tests (JSON fixtures exercising the full flow), but since this is a **spec repo** defining normative validation rules, explicit unit tests for each Validate() method would strengthen implementer confidence. A table-driven test per type with valid/invalid cases would be ideal.

types/share.go
```diff
@@ -19,6 +19,56 @@ type ShareMember struct {
...
+		if len(member.SharePubKey) != 48 {
+			return NewError(InvalidShareErrorCode, "invalid committee member share public key length")
+		}
+		if _, exists := seenSigners[member.Signer]; exists {
+			return NewError(InvalidShareErrorCode, "duplicate committee member signer")
+		}
+		seenSigners[member.Signer] = struct{}{}
+	}
+
+	if len(share.Graffiti) > 32 {
```
olegshmuelov: Graffiti is `ssz-size:"32"` (fixed, not max), and the generated SSZ encoder already enforces `size != 32` (`share_encoding.go:44`). But `Validate()` only checks `> 32`, so a nil or short Graffiti passes validation but fails encoding.

Should match the SSZ tag and the pattern used for `SharePubKey` (`!= 48`) two lines above:

```go
if len(share.Graffiti) != 32 {
    return NewError(InvalidShareErrorCode, "invalid graffiti length")
}
```

The doc comment on L26 also says "Graffiti must be <= 32 bytes" — should be "Graffiti must be exactly 32 bytes."

ssv/committee.go
```diff
@@ -39,8 +43,25 @@ func NewCommittee(
 	return c
 }
 
+func (c *Committee) validateInvariants() error {
+	c.validateOnce.Do(func() {
+		if err := (&c.CommitteeMember).Validate(); err != nil {
+			c.validateErr = errors.Wrap(err, "invalid committee member")
```
olegshmuelov, line 49: The repo's CLAUDE.md recommends `fmt.Errorf("...: %w", err)` for error wrapping. The new code uses `errors.Wrap` from `pkg/errors` which matches the existing codebase pattern. This is fine for consistency, but if there's an ongoing migration effort, these could be the new style.

types/operator.go
```diff
@@ -21,6 +21,85 @@ type Operator struct {
...
+	}
+	if len(cm.Committee) > 13 {
+		return NewError(InvalidCommitteeMemberErrorCode, "committee too large")
+	}
+
+	committeeSize := uint64(len(cm.Committee))
+	// Enforce the QBFT committee requirement n >= 3f+1 without risking uint64 overflow.
+	// Equivalent to: f <= floor((n-1)/3)
+	if cm.FaultyNodes > (committeeSize-1)/3 {
+		return NewError(InvalidCommitteeMemberErrorCode, "invalid faulty nodes bound for committee size")
```
olegshmuelov, line 57: The validation enforces `f <= floor((n-1)/3)`, which is the mathematical BFT minimum. But the quorum formula `GetQuorum() = 2f+1` only guarantees safety when `n = 3f+1` exactly. For in-between sizes like n=5,f=1, two quorums of 3 overlap by only 1 node — if that node is faulty, conflicting values can be committed.

In practice SSV uses only {4,7,10,13}-member committees (n=3f+1 for f=1..4), so this is not exploitable in production. But the spec now normatively accepts sizes that the quorum math doesn't cover.

This is a pre-existing design property (the quorum formula is unchanged), not a regression from this PR. Worth discussing with @MatheusFranco99 @GalRogozinski  as a spec tightening: either enforce `n == 3f+1` in Validate(), or use a supported-sizes allowlist `{4,7,10,13}`.

types/operator.go
```diff
@@ -21,6 +21,85 @@ type Operator struct {
...
+		return NewError(InvalidCommitteeMemberErrorCode, "operator ID 0 not allowed")
+	}
+	if len(cm.SSVOperatorPubKey) != 459 {
+		return NewError(InvalidCommitteeMemberErrorCode, "invalid operator public key length")
+	}
+
+	if len(cm.Committee) == 0 {
+		return NewError(InvalidCommitteeMemberErrorCode, "empty committee")
+	}
+	if len(cm.Committee) > 13 {
```
olegshmuelov: The committee max `13` appears as a literal in 6+ validation checks across `operator.go:61`, `share.go:40`, `beacon_types.go:167`, plus all SSZ tags and generated encoders.

SSZ tags can't reference constants, but the validation code can. A shared constant would make the bound self-documenting and prevent drift if it ever changes:

```go
// MaxCommitteeSize is the maximum number of operators in a committee.
const MaxCommitteeSize = 13
```

Then validation becomes `len(cm.Committee) > MaxCommitteeSize`.

types/share.go
```diff
@@ -19,6 +19,56 @@ type ShareMember struct {
...
+// - DomainType must be one of the known SSV domains in this spec
+func (share *Share) Validate() error {
+	if share == nil {
+		return NewError(InvalidShareErrorCode, "nil share")
+	}
+
+	if len(share.SharePubKey) != 48 {
+		return NewError(InvalidShareErrorCode, "invalid share public key length")
+	}
+
```
olegshmuelov: `ValidatorDuty.Validate()` checks for all-zero pubkey (`beacon_types.go:146-154`), but `Share.Validate()` doesn't check `ValidatorPubKey`. Since `ValidatorPK` is `[48]byte` the length is always correct by type, but an all-zero pubkey is still invalid.

For consistency:

```go
if share.ValidatorPubKey == (ValidatorPK{}) {
    return NewError(InvalidShareErrorCode, "zero validator pubkey not allowed")
}
```

types/beacon_types.go
olegshmuelov, line 170: `AggregatorCommitteeDuty` already has `Validate()` (L210) checking slot consistency and role constraints. `CommitteeDuty` has no equivalent — its validation is done inline in `Committee.StartDuty()` (checking for empty duties, nil entries, per-duty validation).

For symmetry, consider adding `CommitteeDuty.Validate()` that checks:
- `ValidatorDuties` is non-empty
- Each duty is non-nil with slot matching `cd.Slot`
- Each duty type is `BNRoleAttester` or `BNRoleSyncCommittee`

This would centralize the container-level invariants instead of spreading them across `StartDuty`.

olegshmuelov (changes requested):
I traced every `Validate()` check against the SSZ generated encoder/decoder. Several checks duplicate what SSZ already guarantees structurally. In a spec repo, `Validate()` should define **semantic invariants** — things SSZ can't enforce. Repeating SSZ's structural guarantees creates noise for implementers in other languages who'll treat these methods as the normative checklist.

**Checks to remove (SSZ already enforces these):**

| Type | Check | SSZ guarantee |
|------|-------|---------------|
| `SSVMessage` | `len(Data) > 726932` | `ssz-max` enforced on decode |
| `Operator` | `len(SSVOperatorPubKey) != 459` | `ssz-size:"459"` — fixed read |
| `CommitteeMember` | `len(SSVOperatorPubKey) != 459` | Same |
| `CommitteeMember` | `len(Committee) > 13` | `ssz-max:"13"` enforced on decode |
| `Share` | `len(SharePubKey) != 48` | `ssz-size:"48"` — fixed read |
| `Share` | `len(Committee) > 13` | `ssz-max:"13"` enforced on decode |
| `Share` | committee member nil | SSZ allocates all members |
| `Share` | `len(member.SharePubKey) != 48` | `ssz-size:"48"` — fixed read |
| `Share` | `len(Graffiti) > 32` | `ssz-size:"32"` — fixed read (**and wrong — see Issue #4**) |
| `ValidatorDuty` | `len(SyncCommitteeIndices) > 13` | `ssz-max:"13"` enforced on decode |
| `BeaconVote` | `Source == nil \|\| Target == nil` | SSZ allocates both on decode |

**Checks to keep (semantic, SSZ doesn't validate):**
- `OperatorID == 0`, `Signer == 0` — zero is a valid encoding but semantically invalid
- FaultyNodes bound, duplicate IDs, self-membership, CommitteeID hash
- `DomainType.IsKnown()`, MsgType range, role `< 0`
- `Source.Epoch >= Target.Epoch`
- Duty type mapping, zero pubkey, committee index bounds

`Validate()` methods should only contain the second list. SSZ tags are the spec for structural constraints — repeating them in Go code is not normative, it's noise.

types/beacon_types.go
```diff
@@ -129,6 +129,48 @@ func (bd *ValidatorDuty) RunnerRole() RunnerRole {
...
+	allZero := true
+	for _, b := range bd.PubKey[:] {
+		if b != 0 {
+			allZero = false
+			break
+		}
+	}
+	if allZero {
+		return NewError(InvalidValidatorDutyErrorCode, "zero validator pubkey not allowed")
+	}
```
nkryuchkov:
```suggestion
	if bd.PubKey.IsZero() {
		return NewError(InvalidValidatorDutyErrorCode, "zero validator pubkey not allowed")
	}
```

ssv/committee.go
julienh-ssv: Let's keep this for consistency, I think another dedicated ticket to migrate the error message can be opened later.

nkryuchkov (dismissed):

types/operator.go
julienh-ssv: Good catch — we now enforce `n == 3f+1` *exactly* in `CommitteeMember.Validate()` (via `(n-1)%3 == 0` and `f == (n-1)/3`) to preserve the `2f+1` quorum intersection guarantees. Thanks for calling this out.

types/messages.go
```diff
@@ -124,6 +124,31 @@ func (msg *SSVMessage) Decode(data []byte) error {
...
 
+// Validate checks the following rules:
+// - MsgType must be in the known range
+// - MsgID must encode a non-negative RunnerRole (unknown positive roles are accepted for forward compatibility)
+func (msg *SSVMessage) Validate() error {
+	if msg == nil {
+		return NewError(NilSSVMessageErrorCode, "nil SSVMessage")
+	}
+
+	switch msg.MsgType {
```
olegshmuelov, line 135: `b.Source.Epoch >= b.Target.Epoch` panics if Source or Target is nil. This was removed as SSZ-structural, but it's actually a semantic check — nil checkpoints are semantically invalid, not just structurally. SSZ decode allocates both, so production callers are safe today. But a Validate() method should be total (never panic), especially in a spec repo where other implementations will mirror this code.

Suggestion: restore the nil guard before the epoch comparison:
```go
if b.Source == nil || b.Target == nil {
    return NewError(BeaconVoteNilCheckpointErrorCode, "nil source or target checkpoint")
}
```

And add test coverage:
```go
t.Run("zero-value beacon vote", func(t *testing.T) {
    err := (&BeaconVote{}).Validate()
    require.Error(t, err)
    require.Equal(t, BeaconVoteNilCheckpointErrorCode, err.(*Error).Code)
})

t.Run("nil source only", func(t *testing.T) {
    bv := &BeaconVote{Target: &phase0.Checkpoint{Epoch: 1}}
    err := bv.Validate()
    require.Error(t, err)
    require.Equal(t, BeaconVoteNilCheckpointErrorCode, err.(*Error).Code)
})
```

Note: I originally classified this as SSZ-structural in the first review, but on re-examination the nil-to-pointer distinction makes it semantic.

types/constants.go
```diff
@@ -0,0 +1,7 @@
+package types
+
+// MaxCommitteeSize is the maximum number of operators in a committee.
+//
+// Note: SSZ struct tags and generated encoders still embed the numeric limit,
+// but validation code should reference this constant to avoid drift.
+const MaxCommitteeSize = 13
```
olegshmuelov: `MaxCommitteeSize` exists but has zero consumers. The validation checks that would use it were removed as SSZ duplicates. Either reference it somewhere (e.g., in a doc comment or at least one validation path) or remove it.

ssv/committee.go
```diff
@@ -39,21 +43,45 @@ func NewCommittee(
...
 
 	// Get objects according to duty type
+	var slot phase0.Slot
 	var runnerMap *map[phase0.Slot]Runner
 	var createFn *CreateRunnerFn
 	var validatorDuties []*types.ValidatorDuty
 	switch d := duty.(type) {
 	case *types.CommitteeDuty:
+		if err := d.Validate(); err != nil {
+			return errors.Wrap(err, "invalid committee duty")
```
olegshmuelov: `d.Validate()` at L73 validates ALL duties in the CommitteeDuty — including duties for validators this committee doesn't own. On `main`, malformed foreign duties were silently skipped by the filtering loop. After this change, a single malformed foreign duty rejects the entire `StartDuty` for all validators.

This also creates redundancy in the filtering loop:
- `bduty == nil { continue }` (L105) is now dead code — `Validate()` already rejected nil
- `bduty.Validate()` (L111) is redundant — `Validate()` already called it on all duties

Two cleaner options:
1. **Validate after filtering** — only validate owned duties, preserves old tolerance for foreign malformed entries
2. **Validate before filtering and remove redundant checks** — clean fail-fast, but document this as an intentional contract change

Currently it's both: validate everything, then re-validate owned ones.

```go
// Option 1: validate only owned duties (after filtering)
case *types.CommitteeDuty:
    slot = phase0.Slot(d.Slot)
    // ... filter first, then validate the filtered duty
```

```go
// Option 2: validate all, skip redundant re-checks
case *types.CommitteeDuty:
    if err := d.Validate(); err != nil {
        return errors.Wrap(err, "invalid committee duty")
    }
    // ... then in the loop, skip bduty.Validate() and nil check
```

types/messages.go
julienh-ssv: Addressed in `c96fc7fb`. `BeaconVote.Validate()` is nil-safe again, and I added regression coverage for zero-value and nil-checkpoint cases in `types/validation_test.go`.

types/constants.go
julienh-ssv: Addressed in `c96fc7fb`. I removed `MaxCommitteeSize` because the SSZ-structural checks that would have used it were intentionally dropped, so keeping an otherwise-unused constant would just add dead spec surface.

ssv/committee.go
julienh-ssv: Addressed in `c96fc7fb`. `Committee.StartDuty()` now filters by owned validators first and validates the resulting `CommitteeDuty`, which preserves the previous tolerance for malformed foreign duties. I also added a regression test in `ssv/committee_test.go`.

olegshmuelov (dismissed):
LGTM!

nkryuchkov (dismissed):

types/consensus_data.go
```diff
@@ -122,6 +122,22 @@ func (b *BeaconVote) Decode(data []byte) error {
...
+func (b *BeaconVote) Validate() error {
+	if b == nil {
+		return NewError(BeaconVoteNilCheckpointErrorCode, "nil beacon vote")
+	}
+	if b.Source == nil || b.Target == nil {
+		return NewError(BeaconVoteNilCheckpointErrorCode, "nil source or target checkpoint")
+	}
+	if b.Source.Epoch >= b.Target.Epoch {
+		return NewError(AttestationSourceNotLessThanTargetErrorCode, "attestation data source >= target")
+	}
```
MatheusFranco99, line 137: I think there may be an edge case in which source == target, but I think this is possible only on genesis, and thus now we would only be caring about testnets.
But still, considering that an SSV validator would need to register itsefl first before this function is ever executed, I think we are fine.

Also, I think we already had such check somewhere else as well.

Anyway, just registering to open for discussion if it brings any thoughts/concerns from anyone

ssv/committee.go
```diff
@@ -74,9 +99,15 @@ func (c *Committee) StartDuty(duty types.Duty) error {
 	filteredValidatorDuties := make([]*types.ValidatorDuty, 0)
 
 	for _, bduty := range validatorDuties {
+		if bduty == nil {
+			continue
+		}
```
MatheusFranco99, line 104: Isn't this an error?
Maybe, it's fine that we make a best effort to advance, but seems a weird possibility
Wdyt?

ssv/spectest/run_test.go
```diff
@@ -56,7 +56,10 @@ func TestJson(t *testing.T) {
 	untypedTests := map[string]interface{}{}
 	byteValue, err := os.ReadFile(path)
 	if err != nil {
-		panic(err.Error())
+		if os.IsNotExist(err) {
+			t.Skipf("missing %s; generate it with `make generate-jsons`", path)
+		}
+		t.Fatalf("failed to read %s: %v", path, err)
```
MatheusFranco99, line 62: IIRC, `Skipf` skips silently while printing the msg.
While locally it's fine, I'm only afraid of this skip in the github action.
Even though we run generator json on it, idk which approach is best/safest.
Wdyt?
Aloo @GalRogozinskiany any opinion?

ssv/committee.go
```diff
@@ -101,12 +132,22 @@ func (c *Committee) StartDuty(duty types.Duty) error {
 		return errors.Errorf("unsupported duty type: %T", duty)
 	}
 
+	if committeeDuty, ok := filteredDuty.(*types.CommitteeDuty); ok {
+		if err := committeeDuty.Validate(); err != nil {
+			return errors.Wrap(err, "invalid committee duty")
+		}
+	}
```
MatheusFranco99, line 139: Would it be better to move this check into the switch clause? And also to add one for the `AggregatorCommitteeDuty` type?

MatheusFranco99 (commented):
Looks good! Good work!
Few comments :)

types/consensus_data.go
julienh-ssv: I kept the strict `source < target` rule. This matches the existing attestation-ordering invariant, and the genesis-only `source == target` edge case should stay out of scope here because validators are registered before this path is exercised. The same ordering check is still enforced in the beacon-vote value checks as well, so I am treating this as discussion-only rather than changing semantics in `

ssv/committee.go
julienh-ssv: I kept the best-effort behavior here intentionally. A `nil` entry has no validator index, so we cannot attribute it to one of this committee's owned validators; turning it into a hard error would bring back whole-duty rejection for malformed foreign entries. I added an inline comment to make that tradeoff explicit.

ssv/spectest/run_test.go
julienh-ssv: Agreed. I changed the spectest loaders to fail in CI when `generate/tests.json` is missing, while still skipping locally for developer convenience. I applied the same behavior consistently in `ssv`, `qbft`, and `types` spectests so GitHub Actions cannot pass silently on a missing generated fixture.

ssv/committee.go
julienh-ssv: Done in `9e38369b`. I moved the filtered-duty validation into the switch and added the missing post-filter validation for `AggregatorCommitteeDuty` as well. There is also a regression test covering invalid owned aggregator duties in `ssv/committee_test.go`.

ssv/committee.go
```diff
@@ -39,21 +43,42 @@ func NewCommittee(
...
+	var slot phase0.Slot
 	var runnerMap *map[phase0.Slot]Runner
 	var createFn *CreateRunnerFn
 	var validatorDuties []*types.ValidatorDuty
 	switch d := duty.(type) {
 	case *types.CommitteeDuty:
+		slot = phase0.Slot(d.Slot)
 		runnerMap = &c.CommitteeRunners
 		createFn = &c.CreateCommitteeRunnerFn
 		validatorDuties = d.ValidatorDuties
```
olegshmuelov, line 79: issue (blocking): Typed-nil `*CommitteeDuty` panics at `d.Slot`. The symmetric `*AggregatorCommitteeDuty` branch two lines below already has an `if d == nil` guard; this branch lost it in `c96fc7fb` when the pre-filter `d.Validate()` was removed.

Mirror the aggregator guard and add a regression test in `ssv/committee_test.go` alongside the existing ones.

  ```go
  case *types.CommitteeDuty:
      if d == nil {
          return types.NewError(UnknownDutyRoleDataErrorCode, "nil committee duty")
      }
  ```

types/beacon_types.go
```diff
@@ -129,6 +129,36 @@ func (bd *ValidatorDuty) RunnerRole() RunnerRole {
...
+
+	switch bd.Type {
+	case BNRoleAttester, BNRoleAggregator, BNRoleSyncCommittee, BNRoleSyncCommitteeContribution:
+		if bd.CommitteeLength == 0 {
+			return NewError(InvalidValidatorDutyErrorCode, "committee length must be non-zero")
+		}
+		if bd.ValidatorCommitteeIndex >= bd.CommitteeLength {
+			return NewError(InvalidValidatorDutyErrorCode, "validator committee index out of bounds")
+		}
+	}
```
olegshmuelov, line 157: suggestion (non-blocking): `CommitteeRunner` (sync-committee role) and `AggregatorCommitteeRunner` (sync-committee-contribution role) don't read `CommitteeLength` / `ValidatorCommitteeIndex`, but `Validate()` rejects sync duties that leave them zero.

  Narrow the switch to the roles that actually consume the fields.

  ```go
  switch bd.Type {
  case BNRoleAttester, BNRoleAggregator:
      if bd.CommitteeLength == 0 {
          return NewError(InvalidValidatorDutyErrorCode, "committee length must be non-zero")
      }
      if bd.ValidatorCommitteeIndex >= bd.CommitteeLength {
          return NewError(InvalidValidatorDutyErrorCode, "validator committee index out of bounds")
      }
  }
  ```

olegshmuelov (changes requested):

ssv/committee.go
julienh-ssv: Addressed in `7222bb4e`. I restored the missing typed-nil guard in the `*CommitteeDuty` switch branch and added a regression test in `ssv/committee_test.go` covering `var duty *types.CommitteeDuty; committee.StartDuty(duty)`.

types/beacon_types.go
julienh-ssv: Addressed in `7222bb4e`. I narrowed the committee-metadata validation to `BNRoleAttester` and `BNRoleAggregator` only, and added regression coverage showing that sync-committee and sync-committee-contribution duties can validate with zero committee metadata.

MatheusFranco99 (approved):
LGTM!

qbft/spectest/run_test.go
```diff
@@ -32,11 +32,17 @@ func TestJson(t *testing.T) {
...
 	byteValue, err := os.ReadFile(path)
 	if err != nil {
-		panic(err.Error())
+		if os.IsNotExist(err) {
+			if os.Getenv("CI") != "" {
+				t.Fatalf("missing %s in CI; generate it with `make generate-jsons`", path)
+			}
+			t.Skipf("missing %s; generate it with `make generate-jsons`", path)
+		}
+		t.Fatalf("failed to read %s: %v", path, err)
```
nkryuchkov, line 41: can we create some kind of `assert...` function to avoid repetition?

ssv/committee.go
```diff
@@ -39,21 +43,45 @@ func NewCommittee(
...
 
 	// Get objects according to duty type
+	var slot phase0.Slot
 	var runnerMap *map[phase0.Slot]Runner
 	var createFn *CreateRunnerFn
 	var validatorDuties []*types.ValidatorDuty
 	switch d := duty.(type) {
 	case *types.CommitteeDuty:
+		if d == nil {
+			return types.NewError(types.UnknownDutyRoleDataErrorCode, "nil committee duty")
```
nkryuchkov, line 74: we have `UnknownDutyRoleDataErrorCode` for `CommitteeDuty` but `InvalidAggregatorCommitteeDutyErrorCode` for `AggregatorCommitteeDuty, they look inconsistent with each other

olegshmuelov (approved):

ssv/share_validation.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+	"github.com/pkg/errors"
+
+	"github.com/ssvlabs/ssv-spec/types"
+)
+
+func validateShareMap(share map[phase0.ValidatorIndex]*types.Share) error {
+	for validatorIndex, validatorShare := range share {
+		if err := validatorShare.Validate(); err != nil {
+			return errors.Wrap(err, fmt.Sprintf("invalid share for validator index %d", validatorIndex))
```
nkryuchkov, line 15: can we use https://pkg.go.dev/github.com/pkg/errors#Wrapf ?

types/validate_methods_test.go
```diff
@@ -0,0 +1,170 @@
...
+		op.OperatorID = 0
+		err := op.Validate()
+		require.Error(t, err)
+		require.Equal(t, InvalidOperatorErrorCode, err.(*Error).Code)
+	})
+
+}
+
+func TestCommitteeMemberValidate(t *testing.T) {
+	valid := func() *CommitteeMember {
```
nkryuchkov, line 39: nit: maybe something like `createValidMember`? `valid()` is a bit unclear to understand what it is

types/validation_test.go
```diff
@@ -0,0 +1,90 @@
...
+		msg := &SSVMessage{
+			MsgType: SSVConsensusMsgType,
+			MsgID:   NewCommitteeMsgID(GenesisMainnet, committeeID, RoleUnknown),
+		}
+		err := msg.Validate()
+		require.Error(t, err)
+		require.ErrorIs(t, err, &Error{})
+		require.Equal(t, SSVMessageInvalidRoleErrorCode, err.(*Error).Code)
+	})
+
```
nkryuchkov, line 34: nit: redundant newline, it may eventually trigger the `whitespace` linter if we add it

types/validation_test.go
```diff
@@ -0,0 +1,90 @@
...
+		Signatures:  [][]byte{{1}},
+		OperatorIDs: []OperatorID{1},
+		SSVMessage: &SSVMessage{
+			MsgType: SSVConsensusMsgType,
+			MsgID:   NewCommitteeMsgID(GenesisMainnet, CommitteeID{}, RoleUnknown),
+		},
+	}
+
+	err := msg.Validate()
+	require.Error(t, err)
```
nkryuchkov, line 48: is it redundant due to the check below?

ssv/committee.go
julienharbulot: Okay I will create a small ticket for it under refactoring/align names and open a PR for it later today

nkryuchkov (approved):
Approving it, though, because all my recent comments seem like nits

julienharbulot:
Fixes https://github.com/ssvlabs/ssv-node-board/issues/865
Fixed

PR #626, opened 2026-05-06
fix: use valid BLS bytes in proposer fixtures

### LINKED ISSUE #622: Capella/Deneb proposer fixtures reuse codec-only BLS placeholder bytes and fail strict decoders
In `types/testingutils/beacon_node_consts.go`, the source comments indicate that the Capella and Deneb proposer fixtures were taken from upstream go-eth2-client tests:

- `capellaBlock`
- `denebBlockContents`

Those upstream fixtures are good codec fixtures, but they contain placeholder-style BLS byte strings such as sequential `0x000102...` pubkeys and `0x606162...` signatures. These values have the correct byte lengths, but they are not guaranteed to be valid compressed BLS points.

That distinction matters because go-eth2-client treats BLS fields permissively in these paths:
- JSON decode mainly enforces prefix / hex / length
- SSZ decode copies raw bytes into fixed-size fields

So these fixtures are acceptable for codec round-trip tests, but they are not suitable as cross-client proposer/interoperability vectors.

Strict `blst`-based decoders reject representative Capella/Deneb placeholder values, while a representative Electra signature from the real-block fixture decodes successfully. This suggests the issue is specifically with the copied Capella/Deneb fixture class, not necessarily all proposer fixtures.

## Why this matters

`ssv-spec` is using these fixtures for proposer/spec coverage, not just codec coverage. That means downstream clients with stricter BLS decoding either:
- fail on otherwise reasonable inputs, or
- need test-only crypto bypasses to consume the vectors

That weakens what the fixtures are proving.

## Proposed fix

Replace the Capella/Deneb placeholder BLS fields in `types/testingutils/beacon_node_consts.go` with valid compressed BLS encodings.

If the goal of these tests is decode / block-handling compatibility rather than signature-verification semantics, the values do not need to be meaningful signatures for the block contents, but they should be valid compressed points so that strict implementations can decode them without special-case bypasses.

diegomrsantos:
## Summary

Fixes #622 by replacing codec-only placeholder bytes in the Capella and Deneb proposer fixtures with byte strings that Herumi accepts as compressed BLS pubkeys and signatures.

The direct fixture change is in `types/testingutils/beacon_node_consts.go`. The generated spectest JSONs were then refreshed so proposer and randao vectors use the updated roots. I also added a regression test that strict-deserializes every BLS pubkey/signature field in the Capella and Deneb proposer fixtures.

## Notes for reviewers

Start with `types/testingutils/beacon_node_consts.go` and `types/testingutils/beacon_node_consts_test.go`; most of the rest is generated fixture output.

While regenerating, `upstream/main` also produced three stale aggregator committee generated fixture updates. Those are included because `go test ./ssv/spectest -run TestAll -count=1` fails on the current code without them. Other unrelated generator drift was excluded.

## Verification

- `env PATH=/Users/diegomrsantos/go/bin:$PATH make generate-jsons`
- `go test ./types/... -count=1`
- `go test ./ssv/spectest -run TestAll -count=1`


diegomrsantos:
Superseded by #628, which keeps the #622 source/test change and generated JSONs split into separate commits and drops unrelated generator drift.

PR #628, opened 2026-05-06
fix: use valid BLS bytes in proposer fixtures

### LINKED ISSUE #622: Capella/Deneb proposer fixtures reuse codec-only BLS placeholder bytes and fail strict decoders

diegomrsantos:
## Summary

Fixes #622 by normalizing the Capella and Deneb proposer fixtures to use BLS pubkey and signature bytes that Herumi accepts as compressed values.

The raw beacon_node_consts.go JSON constants are intentionally unchanged in the final PR diff. They are huge minified literals, so the fix now happens after decoding the fixtures into typed proposer objects. This keeps the reviewable code small while still producing regenerated proposer and randao vectors with valid BLS bytes.

## Notes for reviewers

Start with types/testingutils/beacon_node_proposer.go and types/testingutils/beacon_node_consts_test.go. The JSON files are generated output from the changed proposer fixture roots.

I intentionally excluded unrelated generator drift from current upstream/main:

- three aggregator committee JSON files
- six generated SSZ encoding hash files
- eight signed SSV message / SSV message JSON files

## Verification

Passed: types package tests, fixture generation with goimports on PATH, and proposer-only SSV spectests.

The full SSV spectest still hits the existing unrelated aggregator-committee fixture mismatch when the unrelated generator drift is kept out of this PR.


diegomrsantos:
Closing this as superseded by Shane's #627, which fixes the raw proposer fixture literals directly and includes the regen tooling.

PR #629, opened 2026-05-07
ci: use supported golangci-lint installer

diegomrsantos:
Updates the golangci-lint installer URL from the retired `master` branch script to the supported installer URL recommended by upstream. This keeps the existing `latest` behavior while avoiding the checksum matching issue triggered by newer release assets.

The v2.12.2 tarball checksum itself appears valid; the failure is in the old installer path used by this repo. Upstream closed the issue as a duplicate and recommended this installer URL here: https://github.com/golangci/golangci-lint/issues/6572#issuecomment-4400895705

No sync workflow behavior or generated spec-test behavior is changed.

Verification: installed the current latest linter locally through the supported installer URL and ran the lint target successfully.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,10 @@
 GOPATH?=$(shell go env GOPATH)
 TEST_PKG?=./...
+GOLANGCI_LINT_VERSION?=v2.11.4
 
 .PHONY: lint-prepare
 lint-prepare:
-	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s latest
+	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- $(GOLANGCI_LINT_VERSION)
 
 .PHONY: lint
 lint:
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,9 @@
 GOPATH?=$(shell go env GOPATH)
 TEST_PKG?=./...
-GOLANGCI_LINT_VERSION?=v2.11.4
 
 .PHONY: lint-prepare
 lint-prepare:
-	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- $(GOLANGCI_LINT_VERSION)
+	curl -sfL https://golangci-lint.run/install.sh| sh -s latest
 
 .PHONY: lint
 lint:


Makefile
```diff
@@ -1,9 +1,10 @@
 GOPATH?=$(shell go env GOPATH)
 TEST_PKG?=./...
+GOLANGCI_LINT_VERSION?=v2.11.4
 
 .PHONY: lint-prepare
 lint-prepare:
-	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s latest
+	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- $(GOLANGCI_LINT_VERSION)
```
ldez:
```suggestion
	curl -sfL https://golangci-lint.run/install.sh | sh -s latest
```

Makefile
```diff
@@ -1,9 +1,10 @@
 GOPATH?=$(shell go env GOPATH)
 TEST_PKG?=./...
+GOLANGCI_LINT_VERSION?=v2.11.4
```
ldez:
```suggestion
```

GalRogozinski:
Heads-up: #635 (branch `pr-610-fixes`, into `spec-test-ci`) makes this same change in [`a2505dd9d`](https://github.com/ssvlabs/ssv-spec/commit/a2505dd9d), together with a version pin.

The reason it came up there: `make lint-prepare` pinned to the version `lint.yaml` uses (v2.12.2) installed nothing, because the `install.sh` served from the golangci-lint `master` branch fails sha256 verification for v2 releases — same root cause you fixed here.

```
hash_sha256_verify checksum for golangci-lint-2.12.2-darwin-arm64.tar.gz did not verify
```

[`a2505dd9d`](https://github.com/ssvlabs/ssv-spec/commit/a2505dd9d) is your URL swap plus `GOLANGCI_LINT_VERSION?=v2.12.2`, so local `make lint` and CI run the same version rather than `latest`. Verified it installs 2.12.2.

Leaving this open for now — if #635 lands first, this one should be **closed rather than merged**, since merging both would revert the pin to `latest`. Credit for the installer fix is yours either way; happy to drop [`a2505dd9d`](https://github.com/ssvlabs/ssv-spec/commit/a2505dd9d) and take this PR instead if you'd prefer, with the pin added on top.

GalRogozinski:
Superseded — closing rather than merging, so the version pin survives.

The installer fix landed on `spec-test-ci` (#610) via #635 as [`7e5cb51fd`](https://github.com/ssvlabs/ssv-spec/commit/7e5cb51fd), together with `GOLANGCI_LINT_VERSION?=v2.12.2` so `make lint` and CI run the same version instead of `latest`. Merging this PR as well would set it back to `latest`.

Your diagnosis was right and it went further than the URL: pinning to the version `lint.yaml` uses made the old installer fail sha256 verification outright, so `make lint-prepare` installed nothing at all.

Reopen if you'd rather land it from here instead — happy to revert that hunk and take yours with the pin on top.

PR #94, opened 2022-11-07
Process Consensus Messages w/o Running Duty

amir-blox:
This PR enables sync messages to propagate down the pipeline. Including the following changes:

- allowed consensus messages to be processed even if there is no running duty
- upon decided: set state.round only for relevant instances

PR #632, opened 2026-06-23
types: add ePBS (Gloas) domains, roles, and partial-sig types

iurii-ssv:
Adds the ssv-spec wire constants for the new ePBS (Gloas) duties, per [SIP #94](https://github.com/ssvlabs/SIPs/pull/94).

## What
- **Domains** (`beacon_types.go`): `DomainBeaconBuilder` (0x0B), `DomainPTCAttester` (0x0C), `DomainProposerPreferences` (0x0D) — the consensus-spec Gloas domains.
- **Roles**: beacon roles `BNRolePTCAttester` / `BNRoleProposerPreferences` / `BNRoleEnvelopeBuilder` and matching runner roles `RolePTCAttester` / `RoleProposerPreferences` / `RoleEnvelopeBuilder` (7/8/9), wired through `MapDutyToRunnerRole` and both `String()` methods.
- **Partial-sig types**: `PTCAttesterPartialSig` (7), `ProposerPreferencesPartialSig` (8). (§6 envelope post-consensus reuses `PostConsensusPartialSig`, discriminated by role, so it adds no type of its own.)

## Why
Consumed by the node-side ePBS implementation (PTC attestation runner first). The remote signer (`ekm`) lives in a separate module that can only reference signing domains via ssv-spec, so the PTC signing path needs `DomainPTCAttester` here rather than node-side.

Branched off the commit ssv currently pins, so the downstream go.mod bump is purely additive.


MatheusFranco99:
Just quick small nit. I did a find all for `BNRoleEnvelopeBuilder` according to the PR description but found nothing. In the code it actually is `BNRoleEnvelopeProposer`

iurii-ssv:
@MatheusFranco99 good catch — the role had already been renamed on the branch to `EnvelopeProposer` (`BNRoleEnvelopeProposer` / `RoleEnvelopeProposer`), but the PR description still said `EnvelopeBuilder`. Updated the description to match the code; it now also covers the request-auth pair added since (`DomainBuilderRequestAuth` / `RequestAuthPartialSig`, for ssvlabs/ssv#2962).

Also rebased the PR onto `main` and changed the base branch accordingly (the old base branch's content had landed on `main` via the Boole merge) — that also picks up main's pinned golangci-lint setup, which fixes the failing Lint check here.

GalRogozinski (dismissed):
Verified every constant against upstream rather than against the description.

**Domains** — match `consensus-specs/specs/gloas/beacon-chain.md` exactly:
`DOMAIN_BEACON_BUILDER` `0x0B000000`, `DOMAIN_PTC_ATTESTER` `0x0C000000`, `DOMAIN_PROPOSER_PREFERENCES` `0x0D000000`. `DomainBuilderRequestAuth` `0x0B000001` matches `builder-specs/specs/gloas/builder.md`, and the comment's "genesis-style `compute_domain`" claim matches that spec's `get_builder_request_auth_signature`, which calls `compute_domain(DOMAIN_BUILDER_REQUEST_AUTH)` with defaults. The `0x0B000000` vs `0x0B000001` warning comment is worth having — builder-specs itself flags the same confusion.

`DOMAIN_BUILDER_DEPOSIT` (`0x0E000000`) is the one Gloas domain not added; it's builder-side only, so correctly out of scope here.

**Roles / partial-sig types** — match SIP #94 §3, §5, §6: `BNRolePTCAttester`/`RolePTCAttester` = 7, `BNRoleProposerPreferences`/`RoleProposerPreferences` = 8, partial-sig types 7/8/9, with `RequestAuthPartialSig` carried under `RoleProposerPreferences`. The SIP is explicit that builder-auth adds "no new beacon role, runner role, or `MapDutyToRunnerRole` entry", and that envelope signing rides the proposer post-consensus round — so the absence of any envelope role is right.

The `BeaconRole` iota block lands `BNRolePTCAttester` at 7 and `BNRoleProposerPreferences` at 8 as the test asserts (`BNRoleUnknown` is an untyped assignment and doesn't consume an iota). `RunnerRole` 7/8 and `PartialSigMsgType` 7/8/9 are free; 1/3 stay reserved as the comment says.

Golden test passes locally, and CI is green. (`types/spectest` fails locally only for want of generated fixtures — identical on `main`.)

One nit, description-only, no code change needed: the PR body still lists `BNRoleEnvelopeProposer` / `RoleEnvelopeProposer` (9). Those aren't in the diff, and per SIP §6 they shouldn't be — worth editing the body before merge so the changelog doesn't claim a role that doesn't exist.

LGTM.

types/gloas_constants_test.go
```diff
@@ -0,0 +1,40 @@
...
+	// Beacon domains (consensus-specs Gloas), plus builder-specs' DOMAIN_BUILDER_REQUEST_AUTH
+	// which shares the 0x0b first byte and differs only in the last byte.
+	require.Equal(t, [4]byte{0x0b, 0x00, 0x00, 0x00}, DomainBeaconBuilder)
+	require.Equal(t, [4]byte{0x0c, 0x00, 0x00, 0x00}, DomainPTCAttester)
+	require.Equal(t, [4]byte{0x0d, 0x00, 0x00, 0x00}, DomainProposerPreferences)
+	require.Equal(t, [4]byte{0x0b, 0x00, 0x00, 0x01}, DomainBuilderRequestAuth)
+
+	// Beacon roles are iota-assigned, so inserting or reordering entries above them
+	// would silently renumber them.
+	require.Equal(t, BeaconRole(7), BNRolePTCAttester)
```
momosh-ssv: What do you think about adding `duty.MapPTCAttester()` / `duty.MapProposerPreferences()` spec tests next to the existing `Map*` ones in `types/spectest/tests/duty/`? This golden test only enforces the values inside ssv-spec. The JSON vectors are what ssv and Anchor actually run, so right now nothing checks their duty→runner mapping for the new roles (#633 doesn't add these either).

types/gloas_constants_test.go
```diff
@@ -0,0 +1,40 @@
...
+// renumbering an iota block, editing a domain byte — must fail loudly here.
+func TestGloasWireConstants(t *testing.T) {
+	// Beacon domains (consensus-specs Gloas), plus builder-specs' DOMAIN_BUILDER_REQUEST_AUTH
+	// which shares the 0x0b first byte and differs only in the last byte.
+	require.Equal(t, [4]byte{0x0b, 0x00, 0x00, 0x00}, DomainBeaconBuilder)
+	require.Equal(t, [4]byte{0x0c, 0x00, 0x00, 0x00}, DomainPTCAttester)
+	require.Equal(t, [4]byte{0x0d, 0x00, 0x00, 0x00}, DomainProposerPreferences)
+	require.Equal(t, [4]byte{0x0b, 0x00, 0x00, 0x01}, DomainBuilderRequestAuth)
+
+	// Beacon roles are iota-assigned, so inserting or reordering entries above them
```
momosh-ssv: Should we consider pinning the existing values too (BeaconRole 0–6, RunnerRole 0/2/4/5/6, PartialSigMsgType 0/1/4/5/6, existing domains)? Swapping two existing BeaconRole entries would still pass this test, and those values are just as consensus-critical: `ValidatorDuty.Type` is SSZ-encoded. A name→value table per type, renamed to `TestWireConstants`, would cover it.

types/beacon_types.go
```diff
@@ -24,6 +24,19 @@ var (
...
+	DomainPTCAttester         = [4]byte{0x0c, 0x00, 0x00, 0x00}
+	DomainProposerPreferences = [4]byte{0x0d, 0x00, 0x00, 0x00}
+
+	// DomainBuilderRequestAuth is builder-specs' DOMAIN_BUILDER_REQUEST_AUTH, authenticating direct
+	// validator→builder requests (BuilderRequestAuth) under Gloas (ePBS). Like
+	// DomainApplicationBuilder it is an application-namespace domain: signing uses
+	// genesis-style compute_domain (genesis fork version, zero genesis-validators-root),
+	// never a fork-versioned domain. NOT to be confused with the beacon domain
+	// DomainBeaconBuilder (0x0b000000) — the two differ only in the last byte.
+	DomainBuilderRequestAuth = [4]byte{0x0b, 0x00, 0x00, 0x01}
```
momosh-ssv: Worth considering stating the application-domain rule on `DomainCalls.DomainData` too, or adding a small `IsApplicationDomain` helper. `TestingBeaconNode.DomainData` computes every domain genesis-style, so spec tests can't tell a correct implementation from one that signs `DomainBuilderRequestAuth` with a fork-versioned domain. Right now the rule lives only in this comment.

types/partial_sig_message.go
```diff
@@ -18,6 +18,13 @@ const (
 	VoluntaryExitPartialSig = PartialSigMsgType(5)
 	// AggregatorCommitteePartialSig is a partial signature for combined aggregator and sync committee selection proofs
 	AggregatorCommitteePartialSig = PartialSigMsgType(6)
+	// PTCAttesterPartialSig is a partial signature over a Gloas (ePBS) PayloadAttestationData object (SIP #94 §3)
+	PTCAttesterPartialSig = PartialSigMsgType(7)
+	// ProposerPreferencesPartialSig is a partial signature over a Gloas (ePBS) ProposerPreferences object (SIP #94 §5)
+	ProposerPreferencesPartialSig = PartialSigMsgType(8)
+	// RequestAuthPartialSig is a partial signature over a builder-specs BuilderRequestAuth object,
+	// authenticating direct validator→builder requests for a proposal slot (SIP #94 §5 extension)
+	RequestAuthPartialSig = PartialSigMsgType(9)
```
momosh-ssv: Might be worth naming the role(s) allowed to carry this type in the comment (`RoleProposerPreferences`, per the PR body). As far as I can see, `ValidatePreConsensusMsg` never checks `psigMsgs.Type` against the runner. Only the signing domain keeps a type-9 message apart from other pre-consensus types, so once one runner emits both 8 and 9, #633 will likely need a role/type allow-list at dispatch, similar to `committee.go:228`.

types/partial_sig_message.go
```diff
@@ -18,6 +18,13 @@ const (
 	VoluntaryExitPartialSig = PartialSigMsgType(5)
 	// AggregatorCommitteePartialSig is a partial signature for combined aggregator and sync committee selection proofs
 	AggregatorCommitteePartialSig = PartialSigMsgType(6)
+	// PTCAttesterPartialSig is a partial signature over a Gloas (ePBS) PayloadAttestationData object (SIP #94 §3)
```
momosh-ssv: Should we add a "values 2 and 3 stay reserved" note like the one in `runner_role.go`? They're retired wire values (SelectionProof / ContributionProofs), and without the note someone might reuse the gaps later.

types/gloas_constants_test.go
iurii-ssv: Done — added `duty.MapPTCAttester()` and `duty.MapProposerPreferences()` next to the existing `Map*` tests (registered in `all_tests.go` + `testdoc.go`). They now run as JSON vectors too, so the duty→runner mapping for both new roles is enforced where ssv and Anchor consume it, not only in the golden test.

types/gloas_constants_test.go
iurii-ssv: Done — reworked the golden test into `TestWireConstants` (file renamed to `wire_constants_test.go`) with a name→value table per type. It now pins the existing domains / BeaconRole / RunnerRole / PartialSigMsgType values alongside the Gloas ones, plus the `MapDutyToRunnerRole` routing and the `String()` names — so a swapped or renumbered existing entry fails loudly.

types/beacon_types.go
iurii-ssv: Done — documented the rule on `DomainCalls.DomainData` (in #633, where the interface lives) so it's visible where domains are computed, not only on the `DomainBuilderRequestAuth` constant.

Skipped the `IsApplicationDomain` helper: with `TestingBeaconNode` computing every domain genesis-style there's no in-repo consumer for it, so it'd be dead code — the comment carries the contract instead.

types/partial_sig_message.go
iurii-ssv: Named the carrier role in the comment — `RequestAuthPartialSig` now notes it's emitted by the `RoleProposerPreferences` runner (alongside `ProposerPreferencesPartialSig`), told apart from it only by signing domain.

On the dispatch allow-list: good point. It isn't exploitable today — `ProcessPreConsensus` routes type 9 to `processRequestAuth`, and anything else falls through to the type-8 path where the `DomainProposerPreferences` domain check rejects a foreign type — but an explicit `{ProposerPreferencesPartialSig, RequestAuthPartialSig}` guard at entry (à la `committee.go:228`) would be clearer. Considering it as a small #633 follow-up.

types/partial_sig_message.go
iurii-ssv: Done — added a note that `PartialSigMsgType` values 2 and 3 stay reserved (retired SelectionProof / ContributionProofs), mirroring the one in `runner_role.go`.

y0sher (approved):

